PR #6, opened 2015-09-19
Adds ability to set table names via config file and env() variables

michaeljhopkins:


freekmurze:
In essence your changes are ok, but there a few issues. Could you please change the following things:
- the tests are failing because you renamed the migration-file. It was a stub to avoid composers warnings when installed
- rename the key "tables" to "tableNames" in the config
- above that line in the config file add a comment that describes what these values do.
- add instructions in the readme on how to publish the config file (cfr the other spatie packages)
- the environment variables have a "ACL_" prefix, change the prefix to "PERMISSIONS_TABLE_NAME_"
- do not create properties on the migration class. Just use the config value directly where necessary.

Thanks!

freekmurze:
@sebastiandedeyne made the argument not to use env-variables by to define the table names. I agree. It's beter to have a optinally publishable config file where the names of the tables are hard coded by default.

If someone wants to make the table names vary by environment (a use case that won't occur often), env variables can be introduced manually in the config file.

freekmurze:
The tests are failing, can you look into that?

freekmurze:
I'm closing this for now, reopen if you still want to make the necessary changes.

PR #15, opened 2015-10-11
Fix for running phpunit locally

lowerends:
This PR fixes the initialization of phpunit so that phpunit can be run locally.


freekmurze:
This PR seems fine, but at my end phpunit is running just fine already. Which error are you trying to solve? 

screen shot 2015-10-11 at 15 06 00

lowerends:
The error I got was `ErrorException: mkdir(): File exists` for all tests. This PR first checks if the directory exists before attempting to create it.

Issue #12, opened 2015-10-07
Feature: add blade directives to check if user has certain role


lowerends:
It would be useful to have a blade directive that checks if a user has all or any of a given list (array) of roles. The syntax could be like this:

````
@has(['role_1', 'role_2', 'role_3'])
<some html/php here>
@else
<some other html/php here>
@endhas
````

What are your thoughts on this?

**Comment by freekmurze** [comment]
I think it's a good idea to have that functionality, but I don't like the naming. 

This would be better:
```php
@role('administrator')
<some html/php here>
@endrole
```

`@can` does not have an `else` so I think `@role` shouldn't have `else`.  `@norole` would be better.

Of course the directives should handle both strings and arrays.

If you would make a PR for this, I'd likely accept it.

lowerends:
I'm using `@can` with an `@else` part successfully. I think it will be important that you can define what needs to happen if the user doesn't have any or all of the roles.

As for the naming, it may be even better to make them consistent with the other PR I'm working on, to have `@hasAnyRole` and `@hasAllRoles`.

freekmurze:
In that case our blade directives should have an `@else` has well. Thanks for correcting me on this. 

Good idea on the naming. Blade directives are generally lowercase so I prefer if the ones you add are lowercase as well. On top of the ones you're proposing a `@role` one (that has the behaviour of `@hasanyrole`) would be nice. `@norole` is still valuable so you aren't required to use an `else`.

Thoughts?

lowerends:
Agree, also with the fact that `@norole` can still be valuable.

So to sum up, we would have:

````
@role = @hasanyrole
@hasallroles
@else
@endrole, @endhasanyrole, @endhasallroles
@norole
````

freekmurze:
:+1: go forth and code!

Issue #11, opened 2015-09-30
Feature: hasAnyRole and hasAllRoles


lowerends:
It would be useful to have methods to check whether a user has any of a list of roles or whether a user has all of a list of roles. 

If you agree with this, I can create a PR to add these methods.

freekmurze:
Sounds good. Don't forget to add some tests and update the readme with some examples.

Thanks!

lowerends:
I started working on this and should have something ready later this week.

freekmurze:
:+1:

PR #16, opened 2015-10-11
Add accompanying blade directives

lowerends:
This PR fixes issue #12 as well as #11.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![Total Downloads](https://img.shields.io/packagist/dt/spatie/laravel-permission.svg?style=flat-square)](https://packagist.org/packages/spatie/laravel-permission)
 
 This package allows to save permissions and roles in a database. It is built upon the [Laravel's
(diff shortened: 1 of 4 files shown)


freekmurze:
Thank you very much! Will review it thoroughly, change some small nitpicks and tag it as a new version.

lowerends:
With pleasure! Just ping me if needed.

freekmurze:
The blade directives do not seem to be working. Have you tested them in a project of your own?

lowerends:
I tested them locally and they seemed to work fine. I did notice that I forgot the `use` statement for `Illuminate\View\Compilers\BladeCompiler` while copying into the PR, but I've seen you noticed that already as well. Sorry for that.

I noticed that you've added tests as well and made some modifications to the directives. Is it working out for you? Definitely let me know if I can be of any assistence.

freekmurze:
I made some modifications so the directives won't break when nobody is logged in. All tests are working now!

I've tagged a new versions. All the new features are present in v1.1.0.

Thanks again! :+1:

Issue #18, opened 2015-10-19
Allow extension of Role and Permission


Propaganistas:
Hi

I've ran into the case that I need to be able to assign a description to the Permission model.
Normally, one would extend the Permission class and proceed, however this won't work with this package since both Role and Permission classes are quite a lot hard-coded using class imports.

Some suggestions:
- use the IoC to bind an Eloquent model to Role and Permission to allow developers to override when necessary.
- use a config value like Laravel's built-in `auth.model`

arcanedev-maroc:
@Propaganistas Yeah, there is no config file for this package to customize or extend it.

freekmurze:
Using a config file to specify the model classes and removing the hardcoded imports seems like an excellent idea. I'll try to make this happen in the near future. If you need this quick, feel free to submit a PR.

freekmurze:
Solved in v1.2.0 of the package

Issue #5, opened 2015-09-19
Config file for setting table names


michaeljhopkins:
I think it's likely pretty common that a lot of us already have tables (from some other 3rd party acl) to do most of what this migrates. In interest of keeping db changes to a minimum, it'd be ideal for something like this

```php
'tables' => [
    'permissions' => env('PERMISSIONS_TABLE','permissions'),
    'user_has_permissions' => env('USER_HAS_PERMISSIONS_TABLE','user_has_permissions'),
]
```
and so on with the rest of the tables. 

Short issue, I have every table you're creating already, just named differently

freekmurze:
I'll add a config file to the package to allow naming of the tables in the near future.

michaeljhopkins:
Working on a PR just to get the functionality there. However you'll likely need to touch it up a bit

michaeljhopkins:
see if that change works for you!

cheers
-m

freekmurze:
I've added some comments to your PR.

freekmurze:
Solved in v1.2.0 of the package

PR #20, opened 2015-10-26
Add configuration, Add users relationship

DraperStudio:
This pull requests adds the ability to configure the Models that should be used by Eloquent and the table names that should be used for migrations.

The models now use the `Role` and `Permission` contracts which are bound to the IoC so that they can be resolved to make swapping them out easier.

I've also added a `users` relation so that it is possible to get all associated users for a permission or role like this `Role::first()->users()->count()`.

This PR should solve issue https://github.com/spatie/laravel-permission/issues/18 and https://github.com/spatie/laravel-permission/issues/5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,81 @@ running the migrations:
 php artisan migrate
 ```
 
-Finally add the `Spatie\Permission\Traits\HasRoles`-trait to the User model.
+You can publish the config-file with:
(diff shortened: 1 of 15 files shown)


freekmurze:
On first glance this looks like a quality PR! :+1: I'll review in in detail in the coming days.

PR #21, opened 2015-10-26
Allow configurable Role and Permission models

Propaganistas:
See #18 

I suggest to roll this in a major release since it was necessary to change migration publishing behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 [![Quality Score](https://img.shields.io/scrutinizer/g/spatie/laravel-permission.svg?style=flat-square)](https://scrutinizer-ci.com/g/spatie/laravel-permission)
 [![Total Downloads](https://img.shields.io/packagist/dt/spatie/laravel-permission.svg?style=flat-square)](https://packagist.org/packages/spatie/laravel-permission)
 
-This package allows to save permissions and roles in a database. It is built upon the [Laravel's
+This package allows to save permissions and roles in a database. It is built upon [Laravel's
 authorization functionality](http://laravel.com/docs/5.1/authorization) that
 was [introduced in version 5.1.11](http://christoph-rumpel.com/2015/09/new-acl-features-in-laravel/)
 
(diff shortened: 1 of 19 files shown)


Propaganistas:
I just noticed DraperStudio also submitted a PR (#20) concerning these features while I wrote this PR. Quite cool, 'cause the PR list was empty before I started working on it :-)

Anyhow, an objective summary of both pull requests:

This PR only exposes the Eloquent model classes to use for `Role` and `Permission`. It hereby deducts its table names from the model class as regular Eloquent models would do. If custom models are provided, custom migrations need to be created as well.  Everything related to this package is extracted into a Contract, resulting in 4 Contracts in total.

The other PR explicitly asks for table names and injects it dynamically into the supplied migration table. The provided contracts however do not enforce the additional behavior currently present in traits (e.g. revokeRole etc.). Docblocks aren't updated as well.

DraperStudio:
1. The table names are configurable to solve this issue https://github.com/spatie/laravel-permission/issues/5

2. The DocBlocks didn't needed to be updated because I named the contracts the same as the Models, so it was just a matter of replacing the `use` statements.

3. The Provided contracts are explicitly for the Models that should be bound to the IoC. I did it that way because if a Model is compliant with a Contract which would enforce compliance with the Trait doesn't really matter just for the binding into the IoC.

Propaganistas:
Hi @DraperStudio! Thanks for joining the discussion. Just a quick comment:

Table names: I felt that adding a config for them would be overkill since the described behavior in #5 is quite niche and the desired functionality is simply achieved using a regular Eloquent model specifying its table explicitly.

Migrations: this is a matter of, let's say, preference. If I as a developer implement a custom Role model, I might also add some additional fields to the pivot table (`role_has_permissions`). That's why I put it into the roles migration. It is therefore assumed that if a custom Role model and table is used, the corresponding relation tables are also set up manually.

Anyway, this isn't a competition so I would advise the maintainer to review both PRs and take best of both worlds.

freekmurze:
Fellow artisans,

thank you for your contributions.  I'll soon review both your PR's and pick the changes I like.

freekmurze:
I've decided to go for @DraperStudio 's. But I still greatly appreciate the work you have put into this.

PR #38, opened 2016-01-29
if no permissions exist then return false

tomschlick:
If the object doesn't have any permissions it would error out as ```$this->permissions``` would be null.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -167,6 +167,10 @@ trait HasRoles
      */
     protected function hasDirectPermission(Permission $permission)
     {
+        if( ! $this->permissions) {
+            return false;
+        }
+        
         if (is_string($permission)) {
             $permission = app(Permission::class)->findByName($permission);
 


freekmurze:
I don't think the current code would error out. Using a non-existing permission the function will result in an exception being thrown. This is expected behaviour. 

If you are getting another exception/error than [this one](https://github.com/spatie/laravel-permission/blob/master/src/Models/Permission.php#L71), please provide a failing test.

tomschlick:
It's not that the permission being passed in doesn't exist. Its that the $this->permissions array is null if there are no permissions records for that object. 

So for instance if I'm a new user and have 0 direct permissions (maybe just the role has the permissions) then that method will error out.

freekmurze:
I'm not seeing where this exception would happen. Let's talk code. Could you please modify [this test](https://github.com/spatie/laravel-permission/blob/master/tests/HasRolesTest.php#L105-L110) to illustrate the problem?

tomschlick:
Odd... the test you just provided is returning an empty collection which satisfies the ->contains() call but my app is returning null for the ->permissions which throws the error. Looking into this further to see why.

tomschlick:
My bad.

Turns out an old ```permissions``` column was still active in our DB schema from our previous auth system. It was removed from the ```$visible``` array at the top of our user class so I wasn't able to see it when dumping.

So that was the cause of the ```null``` returns from ```$this->permissions```. 

Sorry for the trouble! :smile:

freekmurze:
Np! Be sure to open up another issue if you have another question/problem.

PR #48, opened 2016-02-21
Add the missing "haspermissionto" blade directive

Maxeee09:
Everything is in the title

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,14 @@ I don't have all of these roles...
 @endhasallroles
 ```
 
+```php
+@haspermissionto('write code')
+I have permission to write code!
+@else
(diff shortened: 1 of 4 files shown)


freekmurze:
Hi, thanks for contributing!

Two reason why I probably won't pull this:
- I'm very particular about the names of blade directives, "haspermissionto" just feel a bit too long
- I think the fuctionality already exists: https://laravel.com/docs/5.2/authorization#within-blade-templates

Maxeee09:
Yes @freekmurze I understand your choice. I didn't noticed that you can use permission with default laravel "can" (or "cannot") blade directive. My bad

freekmurze:
Np!

PR #50, opened 2016-02-23
update arguments Docblock to string|Permission

adaojunior:


adaojunior:
I just noticed that `hasDirectPermission` is type hinting `$permission` as `Permission`, but it stills checks if `$permission` is string.
Should we change it to `hasDirectPermission($permission)`, otherwise it is gonna faild when trying to pass a string.

freekmurze:
Nice catch. Thank you very much for correcting it.

adaojunior:
@freekmurze what about `hasDirectPermission` expecting a `Permission` ?

adaojunior:
https://github.com/spatie/laravel-permission/pull/51

PR #53, opened 2016-03-01
haspermission Blade Compiler adapter

AlexanderZon:
@haspermission('writer')
    ...
@endhaspermission


freekmurze:
You can use Laravel's native `@can` directive to check if a user has a certain permission.

PR #62, opened 2016-04-04
Make names col unique; add update migration

danielsdeboer:
Changed the original migration to make the column unique and added a migration for upgraders. On upgrade if you run the publish command again it will only publish the new migration.


danielsdeboer:
On second thought, this isn't going to work properly with new installs because it publishes both migrations and the second one will fail.

PR #63, opened 2016-04-04
Make names column unique

danielsdeboer:
I've changed the migration to make the names column unique and added a second migration for updating users. Not sure if that's worth the trouble or not but there it is. The second migration only gets published if the first migration has previously been created.


freekmurze:
Could you please only edit the original migration. A second migration is not needed.

Issue #73, opened 2016-05-03
Issue with Migration


LKDevelopment:
Hello,

We have an automated installation manager who installed this package automatically . By dynamically migrating name creation , it can happen , the migration will be created repeatedly .

PR to resolve this error do I enjoy writing .

freekmurze:
We'll handle this in the PR.

PR #74, opened 2016-05-03
It is better to use a static timestamp for migration filename

LKDevelopment:
Fix for
#73

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -22,7 +22,7 @@ class PermissionServiceProvider extends ServiceProvider
 
         if (!class_exists('CreatePermissionTables')) {
             // Publish the migration
-            $timestamp = date('Y_m_d_His', time());
+            $timestamp = '2016_01_01_000000';
             $this->publishes([
                 __DIR__.'/../resources/migrations/create_permission_tables.php.stub' => $this->app->databasePath().'/migrations/'.$timestamp.'_create_permission_tables.php',
             ], 'migrations');


freekmurze:
I'm kinda liking that the timestamp is dynamic. There is a check around the migration to make sure that it isn't published twice.

LKDevelopment:
@freekmurze This Check doesn't work correctly. 
"Cannot redeclare class CreatePermissionTables"

When i vendor:push every times a new migration will be created.

freekmurze:
Could you trying dumping composer's autoloader immediately after publishing the migration? This will make the class known to the rest of the code preventing the migration from being published twice.

LKDevelopment:
That fix the issue but isn't as clean as possible. It's cleaner to have a static file name (and the possible to check the file exisits in the migrations dir).

freekmurze:
I'm liking the fact that, when using the current solution, the migration does have the right timestamp and that you can see in the migrations of a project when it was added.

LKDevelopment:
Year the fact is nice, but it would be cleaner to put the filename static and add a Comment into the Migration with the current timestamp

freekmurze:
I guess we have different opinions on which solution is better/cleaner 😄

I'm going to leave it like this. If you need this changed, adapt your  installation manager by running the dump of the autoloader. Or you can also fork this package and maintain your own copy.

LKDevelopment:
I think it is a bug.. You should install a package once, and a package publish the migrations and so one, once. But so you can publish it unlimited times (without run composer dump-autoload)

image

PR #78, opened 2016-05-07
Adding hasPermissionTo on role model

nWidart:
Adding hasPermissionTo on role model with tests.

Edit:

I see the coverage on the Role model is 82% from phpstorm and 80% from travis gui (funny there's a difference). However the `hasPermissionTo` method is fully tested, according to phpstorm, it's the `users` method that isn't tested.


freekmurze:
Thank you!

PR #84, opened 2016-06-16
Un-nest role_has_permissions creation

rickselby:
Not sure why this was nested within user_has_roles? If that's intentional, then ignore this.


freekmurze:
Thanks for correcting this.

PR #85, opened 2016-06-17
Fix use statements for traits

rickselby:
I'm still a little confused as to how the tests pass, but I was getting "Target [Spatie\Permission\Contracts\Permission] is not instantiable" and traced it back here.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -2,7 +2,7 @@
 
 namespace Spatie\Permission\Traits;
 
-use Spatie\Permission\Contracts\Permission;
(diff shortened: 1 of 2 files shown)


freekmurze:
The Perimssion/Role models implement the Permission/Role contract. So importing the Contract namespaces should be ok.

Did you try this is a fresh install of Laravel?

rickselby:
Ah, I know what the issue was now. I'm adding this to our in-house users module. I added the users module provider to the service provider list, but not the PermissionServiceProvider, so the Contracts won't have been registered, causing the error.

Sorry!

freekmurze:
No problem, have fun using the package.

PR #92, opened 2016-07-09
Update README with sample using a middleware

peterquentin:
Update README with sample using a middleware to check roles and permissions.


freekmurze:
Thanks!

Issue #95, opened 2016-07-15
Give Multiple permissions to a role


peterquentin:
Can someone look at this little function before I make a pull request?

For me it's useful when I populate my database using seeders.

https://github.com/peterquentin/laravel-permission/commit/0f48477fd788dccec6684bebabc182292cd43a61

freekmurze:
Granting multiple permission at once would be a nice features. How about making it work like this

```php
$user->givePermissionTo('permission-one'); // grant a single permission
$user->givePermissionTo(['permission-one', 'persion-two']) // granting multiple permission with an array
$user->givePermissionTo('permission-one', 'persion-two', ...) // granting multiple permission with arguments
```

To make that last example work we could leverage argument unpacking introduced in PHP5.6. The package currently support PHP5.5, but I'm willing to up the requirements to PHP 5.6, as PHP 5.5 is not maintained anymore.

If you create a PR for this (with tests) I'll probably accept it.

autosplitz:
![code](https://user-images.githubusercontent.com/61923177/112390465-a8658100-8cf6-11eb-9f56-e94760fe6b4b.png)

Issue #96, opened 2016-07-15
Ability to sync multiple roles at once


nWidart:
Hello,

Currently I couldn't find a way to sync roles on a user. As it is now I had to work like this:

``` php
// Update method on the UserRepository
if ($user->roles->isEmpty() === false) {
    foreach ($user->roles as $role) {
        $user->removeRole($role);
    }
}
$user->assignRole(data_get($data, 'role', 'customer'));
```

Meaning I first had to remove the roles on the user, to then add the needed ones back.

Using a sync method, those lines can be replaced by:

``` php
$user->syncRoles([data_get($data, 'role', 'customer')]);
```

(Note I wrap in an array, b/c in my current use-case a user can only have one role, ie: not a multi select).

The code I added on the `User` model for this:

``` php
public function syncRoles(array $roles)
{
    $this->roles()->sync($this->getStoredRoleIds($roles));
}

protected function getStoredRoleIds($roles)
{
    $ids = [];
    foreach ($roles as $role) {
        $ids[] = $this->getRoleId($role);
    }
    return $ids;
}

private function getRoleId($role)
{
    if (is_string($role)) {
        return app(Role::class)->findByName($role)->id;
    }

    return $role->id;
}
```

Now, I'm guessing that foreach can be replaced by something like

``` php
return collect($roles)->each(function ($role) {
    return $this->getRoleId($role);
});
```

I haven't tested this though.

What are your thoughts on adding this functionality to the `HasRoles` trait ?

freekmurze:
Hi Nicolas,

I'm ok with adding a `syncRoles` function.

I've I'm not mistaken the native sync method will already take both `id`'s and model.

```php
$this->roles()->sync($ids) // ok
$this->roles()->sync($arrayWIthRoleModels) // also ok
```

So the only thing that's not supported out of the box are an array with the names of permission names.

```php
$this->roles()->sync(['permission-one', 'permission-two']);
```

So the `syncRoles` function could do something like this

```php
public function syncRoles(array $roles)
{
     $roles = collect($roles)->map(
     ... // convert int / strings to roles models
     );

    $this->roles()->sync($roles);
}
```

If would be super nice if there also was a equivalent function `syncPermissions`.

And of course tests are mandatory 😄 

Thoughts?

freekmurze:
I did have some time to implement it myself.

Added `syncRoles` and `syncPermissions` in version `1.5.0`.

nWidart:
That's awesome, thanks! 😄

PR #98, opened 2016-07-20
Extend givePermissionTo, multiple permissions

peterquentin:
Pull request for Feature #95 to allow Multiple permissions!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.idea
 build
 composer.lock
 docs
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-/.idea
 build
 composer.lock
 docs
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -7,15 +7,23 @@ use Spatie\Permission\Contracts\Permission;
 trait HasPermissions
 {
     /**
-     * Grant the given permission to a role.
+     * Grant the given permission(s) to a role.
      *
-     * @param  $permission
+     * @param string|array|Permission|\Illuminate\Support\Collection $permissions
+     * @param array $params
      *
      * @return HasPermissions
      */
-    public function givePermissionTo($permission)
+    public function givePermissionTo($permissions, ...$params)
     {
-        $this->permissions()->save($this->getStoredPermission($permission));
+        $joinings = $this->resolveJoiningAttributes($params);
+
+        if (is_string($permissions) && $this->usingMultipleArguments($params)) {
+            array_unshift($params, $permissions);
+            $permissions = $params;
+        }
+
+        $this->savePermissions($permissions, $joinings);
 
         $this->forgetCachedPermissions();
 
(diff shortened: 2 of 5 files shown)


src/Traits/HasPermissions.php
```diff
@@ -7,20 +7,47 @@
...
+     * @param string|array|Permission|\Illuminate\Support\Collection $permissions
+     * @param array $params
      *
      * @return HasPermissions
      */
-    public function givePermissionTo($permission)
+    public function givePermissionTo($permissions, ...$params)
     {
-        $this->permissions()->save($this->getStoredPermission($permission));
+        $joinings = $this->resolveJoiningAttributes($params);
```
freekmurze: What does this do function do? Could you give an example? 

I'd rather see that `givePermissionTo` has only one parameter.

.gitignore
```diff
@@ -1,3 +1,4 @@
+/.idea
```
freekmurze: Best to leave this out. This should be done in a global .gitignore file.
https://murze.be/2014/12/create-a-global-gitignore/

freekmurze:
Hi, thank you for this PR. I'm not ready to pull it yet. Could you take a look at the comments I added to the code. Also take a look at #96 for some guidelines.

src/Traits/HasPermissions.php
peterquentin: I realize as i'm typing to explain this it's silly, I was trying to give the user the option to use the joinings option that saveMany on Eloquent has. I'll remove this because in this particular case we can work without it.

.gitignore
peterquentin: Ah learned something already, thnx!

peterquentin:
Okay I've removed the joinings option because it wouldn't be used anyway. Also I was not sure what guidelines you ment when you reffered to #96. Anyhow please check again :)

src/Traits/HasPermissions.php
```diff
@@ -7,20 +7,44 @@
...
      *
-     * @param  $permission
+     * @param string|array|Permission|\Illuminate\Support\Collection $permissions
+     * @param array $params
      *
      * @return HasPermissions
      */
-    public function givePermissionTo($permission)
+    public function givePermissionTo($permissions, ...$params)
     {
```
freekmurze: Could you make this work with method signature `givePermissionTo(...$permissions)`? I'm not sure why `params` is needed.

peterquentin: I could ofcourse, here is the ripple effect though.
The method no longer requires you to pass a parameter to givePermissionTo()  even though we do need atleast one. Therefore I would need to write at least one more test to ensure we have one argument if not Throw a proper exception.

If you like it that way I will implement it, any suggestions on the Exception and Exception message?

freekmurze: Create an new exception for this in the Exceptions dir. I'll rename it myself if I can improve on it after pulling.

peterquentin: Okay, it's done :+1:

peterquentin:
@freekmurze can you pull this now?

freekmurze:
I'll pull but do some cleanup to make the code a bit more cleaner. You've provided a good base and tests to work on. Thanks!

freekmurze:
I refactored your changes. The power of the collection is strong with this one.

```php
    public function givePermissionTo(...$permissions)
    {
        collect($permissions)
            ->flatten()
            ->map(function ($permission) {
                return $this->getStoredPermission($permission);
            })
            ->each(function (Permission $permission) {
                return $this->permissions()->save($permission);
            });

        $this->forgetCachedPermissions();

        return $this;
    }
```

peterquentin:
Hi Freek,

Sweet, I was under the impression that saveMany would use 1 INSERT query. 
Am I right about that or totally wrong? thanks for pulling!

freekmurze:
You're absolutely right. 

This is how it works now:

```php
public function givePermissionTo(...$permissions)
    {
        $permissions = collect($permissions)
            ->flatten()
            ->map(function ($permission) {
                return $this->getStoredPermission($permission);
            });

        $this->permissions()->saveMany($permissions);

        $this->forgetCachedPermissions();

        return $this;
    }
```

PR #119, opened 2016-09-02
added display_name to tables

MasterBratac:
This is a small change only to the migrations that ads a display name that is nullable.


freekmurze:
Not everyone needs this column. You can easily add it in your own project.

PR #126, opened 2016-09-17
Mention the name of the published config file

kinow:
Following the (great!) installation instructions, but had to stop and check what was the name of the file published. Maybe adding it to the docs may help others? Thought about adding a comment `// config/laravel-permission.php`, but thought the ()'s would be better for this case maybe?

Thanks for laravel-permission, great job! 
Bruno


freekmurze:
Thanks!

PR #135, opened 2016-11-10
adds getMainRole method in hasRole Trait

ErikDohmen:
- Adds getMainRole method to the hasRole trait in order to get the main (highest) role of a user
- Adds an extra column to the create_permission_table.php.stub to make the level column available
- Updates the README to demonstrate the functionality, both adding a role with a level and retrieving the main role of the user.


freekmurze:
Thank you for this PR.

In my opinion this functionality is not generic enough to be included by default in the package.

Make these changes in your own project.

Issue #109, opened 2016-08-23
Seeing "ALERT: Could not register permissions" using DatabaseMigrations trait in tests


mstralka:
I am trying to write tests that will use sqlite in-memory and the *DatabaseMigrations* trait but I get this alert in the log for every test:
````
testing.ALERT: Could not register permissions
````

Tracking that down, `PermissionServiceProvider->boot()` calls `$permissionLoader->registerPermissions();`, which then calls `getPermissions()`, and this method is trying to load the `Permission` objects from the database - but the table does not exist yet when the application is being bootstrapped.

If I extend the logging statement in `PermissionRegistrar` to print the exception message:
````
SQLSTATE[HY000]: General error: 1 no such table: permissions (SQL: select * from "permissions")
````

Does anyone have any suggestions for how to use the *DatabaseMigrations* trait with this package in tests?

Thanks

**Comment by mstralka** [comment]
I figured out a workaround, but I don't think I'll be able to use in-memory sqlite.  I created a connection  in *database.php* file like this:
````
'sqlite_testing' => [
        'driver' => 'sqlite',
        'database' => database_path('testing.sqlite'),
        'prefix' => '',
    ]
````

My test class now implements `setUpBeforeClass` like this:

````
public static function setUpBeforeClass()
 {
    parent::setUpBeforeClass();
    $p = new Process('php artisan migrate:refresh --seed --database=sqlite_testing');
    $p->run();
}
````

Then instead of using the *DatabaseMigrations* trait, I use the *DatabaseTransactions* trait.  Doing this eliminates the errors in the log.

Thanks

**Comment by drbyte** [comment]
Have you tried:

```php
use \Illuminate\Support\Facades\Artisan;
```
and then in your `setup` do:
```php
        Artisan::call('migrate');
        Artisan::call('db:seed');
```

mstralka:
Thank you for the suggestion.  I tried that too, but this package's `PermissionServiceProvider` is trying to load the *Permission* objects from the database during the application's bootstrap process.

I can't run `Artisan::call('migrate')` before the application is bootstrapped inside of the parent `setUp()` method, and I can't make it run during the bootstrap process, before `PermissionServiceProvider` is booted.  I could write a custom ServiceProvider and inserted it before `PermissionServiceProvider` in `config/app.php`, but that seems strange.  

So I guess I'll settle for testing against a database on the file system and seed it as noted above.

Thank you

nisbeti:
[...]

PR #143, opened 2016-12-12
Move Permissions Registration to booted method

nisbeti:
When using in memory sqlite database for testing the migrations may not be run until after the boot method, so the ```$permissionLoader->registerPermissions();``` call will fail if called from ```boot()```

https://github.com/spatie/laravel-permission/issues/132 and https://github.com/spatie/laravel-permission/issues/109

This is problematic when when using https://gist.github.com/adamwathan/dd46a8501097942a771925c02bac0111


freekmurze:
There current tests use an in memory db. They appear to be run fine.

bryannielsen:
@freekmurze The reason it's working fine in the package tests is because the permissions are explicitly reloaded in the `TestCase@setUp()` after the database migrations run (https://github.com/spatie/laravel-permission/blob/master/tests/TestCase.php#L35).

I think there should either be a note for developers to add a similar refresh in their tests (I did this) or change the registration like this PR suggests.

freekmurze:
Can you check if the current `dev-master` fixes this problem?

colynb:
I don't see any documentation that says to use the booted method... I don't see that in Laravel's lifecycle explanation. When is that method called? When I just add that method in one of the providers, there's no indication that it's being called.

scottgrayson:
@bryannielsen could you give an example of how you added a `refresh` in tests

nisbeti:
@colynb booted is used and documented in Laravel Spark which itself is a laravel package.

bryannielsen:
@scottgrayson you can look at the refresh example here - https://github.com/spatie/laravel-permission/blob/master/tests/TestCase.php#L35

Looking at this closer I think doing the refresh in your tests is the only way to do it, the application is fully booted by the time the DatabaseMigrations trait is called and it doesn't look like that trait fires any events to hook into.  Maybe we can just add a note to the readme for this.

PR #144, opened 2016-12-13
Call to a member function compileSelect() on null

fahmiardi:
Hi,

I'm trying to extend this package for support mongodb/moloquent using EmbedsMany relationships. This is the repo https://github.com/fahmiardi/laravel-mongodb-permission

But, when I want to publish the config, I've got this message
```bash
[Symfony\Component\Debug\Exception\FatalThrowableError]
Call to a member function compileSelect() on null
```

I think, I know the problem. With move the register model bindings before permission registrar, I solve this issue.

Please accept my PR.

Thanks.

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -28,6 +28,12 @@ class PermissionServiceProvider extends ServiceProvider
             ], 'migrations');
         }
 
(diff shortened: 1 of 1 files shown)


freekmurze:
Thanks

PR #149, opened 2016-12-18
Feature/console commands

fahmiardi:


freekmurze:
Hi, thanks for contributing. 

for most users these commands are a bit overkill. 

If you want to do some stuff manually, it's very easy to open up a tinker session and quickly create your roles and permissions there.

Of course you can still use your commands in your own projects are make a separate package out of them.

fahmiardi:
Ah I see.. Make sense. Thanks for point of view.

PR #150, opened 2016-12-19
Removed migration dynamic timestamp

dducro:
When using this package with a deploy service like envoyer.io, it tries to rerun this migration because a new timestamp is set.
A static timestamp resolves this issue.


freekmurze:
Mmm, this check should avoid that from happening: https://github.com/spatie/laravel-permission/pull/150/files#diff-755935efe1800dd1a082f9d41af34f05R23

dducro:
True, but only if you have your migrations in the "base project" in source control.
We ignored the migrations there because we manage them from a package.
This way we ensure that the newest migrations are published after a new deploy.

PR #155, opened 2016-12-27
Add role scope + tests + readme update

Rias500:
Adds a query scope to the HasRoles trait to scope a query based on one or more roles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,13 @@ $permissions = $user->permissions;
 $roles = $user->roles()->pluck('name'); // returns a collection
 ```
 
+The `HasRoles` also adds a scope to your models to scope the query to certain roles.
+
+```php
+$users = User::role('writer')->get(); // Only returns users with the role 'writer'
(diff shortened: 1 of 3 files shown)


freekmurze:
Great work, thanks!

I'll refactor it a little bit and tag a new release.

freekmurze:
I've completed [my refactor](https://github.com/spatie/laravel-permission/blob/a11d6ed/src/Traits/HasRoles.php#L47-L73). 

Though your code surely did work correctly, I could improved. I didn't like that the query was being built up several times in the function. When creating a flexible function that can handle multiple kinds of input, I generally like to normalise that input to a common format first (in this case an array with `Role` objects) and then perform the processing (in this case building up the query) only once.

Your tests were in extremely helpful with this refactor.

I've also add support for a `Collection` to be passed. 

A new version `1.6.0` containing the scope has been tagged. Thanks again for your work on this.

PR #166, opened 2017-01-10
Changed link to Laravel-auth to include Laravel version

mark-vesterskov:
The Laravel-authorize package is not needed on 5.2.28 and later, so i clarified that in the readme.


freekmurze:
Thanks!

PR #167, opened 2017-01-10
Added HasAnyPermission

mark-vesterskov:
I needed this function to figure out whether a submenu should be added to my menu - if the user has any of the given permissions, return true.

I chose a foreach loop to minimize the number of loops done. The function returns on the first hit.

I'm very open to improvements!


freekmurze:
Could you add tests to ensure that the added functionality works correctly?

mark-vesterskov:
Gave it my best shot. I'm not world champion of Pull Requests yet, but i hope i did it right!

drbyte:
Note: I was asked about this, so posting some additional comments: while `hasAnyPermission()` can be called against the user object, the Blade `@canany` directive has been available within Blade since Laravel 5.6.23 - https://github.com/laravel/framework/releases/tag/v5.6.23

It's a way to access `Gate::any()` ... which gets the same outcome, but for different reasons. Your implementation of either may have varying degrees of efficiency due to the way the data is retrieved or already in memory at the point of being called, so use what's best for your app.

PR #186, opened 2017-02-09
Made hasDirectPermission method public

bakhtiyor:
Changed hasDirectPermission to public. Added tests. Added example and use case where it could be used.


freekmurze:
Great thanks!

Issue #191, opened 2017-02-16
scopeRole not working


whfamily2006:
User Model:

```
class User extends Authenticatable implements AuthenticatableContract, CanResetPasswordContract {
	
	use Notifiable, RevisionableTrait, CanResetPassword, HasRoles, SoftDeletes;
	
	protected $softDelete               = true;
	protected $dates                    = [ 'deleted_at' ];
	protected $revisionCreationsEnabled = true;
	protected $dontKeepRevisionOf       = [ 'remember_token' ];
	protected $guarded                  = [];
	protected $hidden                   = [ 'password','remember_token' ];
}
```
Laravel-Permission config file: 

```
return [
    'models' => [
        'permission' => Spatie\Permission\Models\Permission::class,
        'role' => Spatie\Permission\Models\Role::class,
    ],

    'table_names' => [
        'users' => 'users',
        'roles' => 'roles',
        'permissions' => 'permissions',
        'user_has_permissions' => 'user_has_permissions',
        'user_has_roles' => 'assigned_roles',
        'role_has_permissions' => 'permission_role',

    ],

];

```

My Query: 

`return User::role( [ 'User', 'ResellerRemitter'] )->count();`

Error: 

```
QueryException in Connection.php line 770: 

SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous (SQL: select count(*) as aggregate from `users` where exists (select * from `roles` inner join `assigned_roles` on `roles`.`id` = `assigned_roles`.`role_id` where `assigned_roles`.`user_id` = `users`.`id` and (`id` = 16 or `id` = 18)) and `users`.`deleted_at` is null)
```
@Rias500 @freekmurze

AlexVanderbist:
Hi, 

I couldn't replicate the issue even if I used exactly the same query. This might be because we're on different mysql versions (or different database drivers in general).

I did however submit a PR adding a table prefix to the only ambiguous part of the query (around `('id' = ? or 'id' = ?)`). This might fix the issue you're having.

freekmurze:
Fix by @AlexVanderbist is tagged as `1.9.0`

PR #196, opened 2017-02-20
Ambiguous id in query

AlexVanderbist:
Sometimes SQL complains about an ambiguous `id` in some queries selecting models with certain roles. 

For example:
```sql
select count(*) as aggregate from "users" where exists (
select * from `roles` inner join `user_has_roles` on `roles`.`id` = `user_has_roles`.`role_id`
    where `users`.`id` = `user_has_roles`.`user_id` and (`id` = ? or `id` = ?)
) and `users`.`deleted_at` is null
```
In this query the `('id' = ? or 'id' = ?)` part could reference both `roles` and `user_has_roles`

This PR adds the table prefix for the `roles` table as configured in the config file.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -65,7 +65,7 @@ trait HasRoles
         return $query->whereHas('roles', function ($query) use ($roles) {
             $query->where(function ($query) use ($roles) {
                 foreach ($roles as $role) {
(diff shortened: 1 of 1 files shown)


AlexVanderbist:
This should fix #191

freekmurze:
Well spotted, good fix!

PR #206, opened 2017-02-22
Add the ability to list :

mathieutu:
- all the permissions that the user inherits from his roles,
 - all the user’s permissions.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -277,4 +277,27 @@ trait HasRoles
 
         return $role;
     }
+
+    /**
+     * Return all the permissions the user have via his roles.
+     *
+     * @return Collection
+     */
+    protected function getPermissionsViaRolesAttribute()
+    {
+        return $this->load('roles', 'roles.permissions')
+            ->roles->flatMap(function ($role) {
+                return $role->permissions;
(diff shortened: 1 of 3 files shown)


mathieutu:
I think Travis has a problem: tests passed on 7.0 and 7.1 but all other build have failed, but not only on my PR.

freekmurze:
Thanks!

freekmurze:
Tagged as `v1.11.0`

mathieutu:
My pleasure! :D

src/Traits/HasRoles.php
```diff
@@ -277,4 +277,27 @@ protected function getStoredRole($role)
 
         return $role;
     }
+
+    /**
+     * Return all the permissions the user have via his roles.
+     *
+     * @return Collection
+     */
+    protected function getPermissionsViaRolesAttribute()
```
milroyfraser: @freekmurze @mathieutu in master branch still this is not update. It's just a method not an accessors `getPermissionsViaRoles` . + `getAllPermissions`

freekmurze: It felt better to have these as regular function and not as accessors. These functions are mentioned in the readme.

PR #208, opened 2017-02-23
Fix indentation on docblocks in config file

svenluijten:


svenluijten:
Ok, I can kind of see how StyleCI fails (though all the other docblocks contradict that), but Travis failing isn't this PR's fault, right? 😕

freekmurze:
Thanks!

PR #210, opened 2017-02-25
Ignore .editorconfig when exporting the package

svenluijten:
This way, the `.editorconfig` file doesn't get downloaded by the likes of Composer. It'll still be there when people want to contribute to the repository using `git clone` though 🙂


freekmurze:
Thanks!

PR #216, opened 2017-03-03
add blade directive @permission

meta0102:
added blade directive @permission to check if the current user has permissionto....


freekmurze:
You can use '@can' for checking permissions in your views.

JiberBoom:
but ，i did it ,it does't work,so what

PR #221, opened 2017-03-09
hasPermissionTo() string param uses permissions cache

derekmd:
I'm migrating a project from [zizaco/entrust](https://github.com/Zizaco/entrust/) because it:

* is sparingly maintained
* conflicts with gate policy authorization

I'm much happier with this package but I have performance issues on some pages with too many database queries. The use case is 10 different roles accessing pages that each request require dozens of permission checks (out of 180 potential permissions.). Some pages have over 100 duplicate queries because `hasPermissionTo()` for string arguments doesn't maintain a `Permission` model object cache.

Currently the cache is only used to configure authorization abilities on the container. This will also use that `Permission` collection cache to prevent further database queries on each `auth()->user()->can()` or Blade `@can()` use.

## Codes changes

* `PermissionRegistrar@getPermissions()` is now public, accessible in the `Permission` and `Role` models through the `RefreshesPermissionCache` trait.
* `Permission::findByName()` always fetches from the cache first, not directly from the database.

## Testing

I wanted to do:

```php
\Illuminate\Support\Facades\Cache::shouldReceive('rememberForever')
     ->with('spatie.permission.cache')
     ->andReturn(EloquentCollection::make([$permission]));
```

But it looks like Mockery wasn't pulled in as a dev dependency for this project. I still wanted to write it in a way that the database is never referenced so let me know if you want me to re-implement it a different way.

Cheers,
Derek

### DIFF
--- a/src/Contracts/Permission.php
+++ b/src/Contracts/Permission.php
@@ -23,7 +23,9 @@ interface Permission
      *
      * @param string $name
      *
-     * @throws PermissionDoesNotExist
+     * @throws \Spatie\Permission\Exceptions\PermissionDoesNotExist
+     *
+     * @return Permission
      */
     public static function findByName($name);
 }
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -62,10 +62,12 @@ class Permission extends Model implements PermissionContract
      * @param string $name
      *
      * @throws PermissionDoesNotExist
+     *
+     * @return Permission
      */
     public static function findByName($name)
     {
-        $permission = static::where('name', $name)->first();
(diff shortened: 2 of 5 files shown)


freekmurze:
Hi, 

thank you for this. Changes seems good.

The test can be improved. Could you test this by making multiple calls to `hasPermissionTo` and assert that the amount of executing queries does not increase (cfr the other tests in `CacheTest`?

derekmd:
I've widened the scope of that test to instead make assertions on database queries.

freekmurze:
Thank you!

PR #223, opened 2017-03-12
Make user foreign key name configurable

Kyslik:
Needs better wording for configuration description part.


freekmurze:
Thanks

freekmurze:
Tagged as `1.12.0`

PR #233, opened 2017-03-21
Filter out duplicate roles and permissions

developeritsme:
This will prevent 'Integrity constraint violation' QueryException on `user_has_roles`, `user_has_permissions` and `role_has_permissions` tables.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -17,6 +17,7 @@ trait HasPermissions
     {
         $permissions = collect($permissions)
             ->flatten()
+            ->unique()
             ->map(function ($permission) {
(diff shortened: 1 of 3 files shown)


freekmurze:
In which situation a user would have multiple of the same roles or permissions?

developeritsme:
It happens only when you pass same roles/permissions :)
I had a situation where I was creating roles array depending on various (multiple/different) API responses.
This is not actually any bug in package but I thought it should prevent such (developer's) error.

freekmurze:
Thanks for your PR, but I'm not going to merge it in.

Because this package revolves around security, I'm a fan of letting it fail easily instead of silently passing wrong input.

PR #236, opened 2017-03-23
Add support for role's id for hasRole

webdevium:
Useful, if admin can rename roles, but can't remove from db.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -127,6 +127,10 @@ trait HasRoles
      */
     public function hasRole($roles)
     {
+        if (is_numeric($roles)) {
+            return $this->roles->contains('id', $roles);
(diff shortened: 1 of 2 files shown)


freekmurze:
This is breaking in the sense that if you uses something numeric as your permission name, this will now fail.

But since that's such an uncommon use case I'm willing to merge this in.

Could you add some tests to prove that your change works?

webdevium:
Yes, of course.

freekmurze:
After seeing those these and get a bit of a feel for the functionality, I've reconsidered my opinion. I'm not so confortable with the breaking nature of the changes. If just feels a bit unnatural to allow id's.

So I'm not going to merge this is.

If you need this just add it to your own project.

Nevertheless thank you for your work on this.

PR #257, opened 2017-04-07
Blade: Check permission via role

AbbyJanke:
@can allows checking permissions directly assigned but not assigned via roles.


AbbyJanke:
looks like its not working as intended, will send a new one once fixed.

PR #267, opened 2017-04-12
Check for duplicates when adding new Roles and Permissions

AlexVanderbist:
In v1 this was done through `UNIQUE` constraints but this is no longer possible since the combination of the `name` and `guard_name` columns must be unique.


AlexVanderbist:
Or should the duplicates check happen on the `saving` event instead of the `create` method?

freekmurze:
Check in create seems fine.

Thanks!

Issue #269, opened 2017-04-16
How to count users with role?


Nationalcat:
I update this package to v2, and I found a question.

 v1 (can work) :
```
//RoleController
$roles = Role ::withCount('users') -> paginate(20);
return view('Role')->with(compact('roles'));

//Role.blade.php
@foreach($roles as $role)
{{$role->users_count}}
@endforeach
```

I update to v2, and show this "Call to undefined method Illuminate\Database\Query\Builder::users()"

I think it's model_has_roles is not related to users table.Can somebody help?

freekmurze:
This is fixed in `v2.1.3`.

DubStepMad:
Would be a great idea if this was mentioned in the docs, thank you for sharing as this has helped me.

drbyte:
Counting records from an Eloquent query is covered in the Laravel docs:
https://laravel.com/docs/master/eloquent-relationships#counting-related-models

kevin4dhd:
I have searched all over the internet and I have not found the correct way to count, for example, how many administrators I have

Scemist:
Unfortunately, I didn't find out a short way. To avoid fixed tables and columns names, the best option is use the built in 'roles' relation to get Users and then the Collection to count;

```php
$superAdminCount = User::with('roles')->get()->filter(
    fn ($user) => $user->roles->where('name', 'Super Admin')->toArray()
)->count();
```

Is huge databases is not ideal, but it's a start point. Please let me know your solutions.

aliakel:
You can create your custom Role model which extends spatie role, then you can use MorphToMany:

use Illuminate\Database\Eloquent\Relations\MorphToMany;

class Role extends \Spatie\Permission\Models\Role
{

  
    public function users(): MorphToMany
    {
        return $this->morphedByMany(
            User::class,
            'model',
            config('permission.table_names.model_has_roles'),
            'role_id',
            config('permission.column_names.model_morph_key')
        );
    }
}

then you can use Role::withCount('users')->where ....

PR #273, opened 2017-04-20
Fix permission/role relation to users

### LINKED ISSUE #270: model_has_roles model_type Role in stead of User
Hey,

I upgraded to v2.
Made a migration to migrate existing permissions to the new model_has_roles table.
However when calling assignRole on the user the model type is "Spatie\Permission\Models\Role" in stead of User. This prevents the reverse of the relation. You can't find the models associated with a given role.

Any thoughts on this are appreciated!

AlexVanderbist:
- fix a bug where the `role()`/`permission()` relation to user models would be saved incorrectly
- add `users()` relation on `Permission` and `Role`

This fixes #270 and #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to `laravel-permission` will be documented in this file
 
+## 2.1.2 - 2017-04-20
+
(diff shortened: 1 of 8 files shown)


freekmurze:
Thanks!

PR #327, opened 2017-06-19
use cache if not in debug mode

theemfs:
I don't need caching while in debug mode (while often recreating models).


freekmurze:
Thank you for this.

Failing tests + conflicting files prevent me from pulling this in.

Could you also make this a configuration option that is set to `false` by default (for backwards compatibility reasons) and add a test to prove that this setting works?

theemfs:
PR 328

PR #328, opened 2017-06-19
added permissions cache usage flag

theemfs:
I don't need caching sometimes.

### DIFF
--- a/tests/CacheTest.php
+++ b/tests/CacheTest.php
@@ -35,41 +35,77 @@ class CacheTest extends TestCase
     {
         $this->registrar->registerPermissions();
 
-        $this->assertCount(0, DB::getQueryLog());
+        if (config('permission.cache')) {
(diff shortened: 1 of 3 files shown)


freekmurze:
Please add a test to make sure this works.

tests/CacheTest.php
```diff
@@ -35,41 +35,77 @@ public function it_can_cache_the_permissions()
     {
         $this->registrar->registerPermissions();
 
-        $this->assertCount(0, DB::getQueryLog());
+        if (config('permission.cache')) {
```
freekmurze: The current tests should not be modified to test `config('permission.cache')`. Could you revert those changes.

I'm thinking that one test where you do some assertions on the amount of queries should be enough.

freekmurze:
Closing this because of lack of tests.

PR #334, opened 2017-06-23
-fix model "use"

capons:


freekmurze:
Don't know what you're trying to do. This breaks tests. What is `Moloquent?`

PR #340, opened 2017-06-29
Update PermissionServiceProvider.php

coderkk:
Allow blade to check multiple roles

hasanyrole
```
@hasanyrole('Admin|Editor')
I have admin or Editor roles
@endhasanyrole
```

hasallrole
```
@hasallrole('Admin|Editor')
I have admin and Editor roles
@endhasanyrole
```


coderkk:
Sorry, it has bug here. I will commit another with unittest

PR #341, opened 2017-06-29
Check multiple roles in blade with hasanyrole and hasallrole

coderkk:
Allow blade to check multiple roles

hasanyrole
```
@hasanyrole('Admin|Editor')
I have admin or Editor roles
@endhasanyrole
```

hasallrole
```
@hasallrole('Admin|Editor')
I have admin and Editor roles
@endhasanyrole
```

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -89,14 +89,15 @@ class PermissionServiceProvider extends ServiceProvider
         });
     }
 
-    protected function convertPipeToArray(string $pipeString) {
+    protected function convertPipeToArray(string $pipeString)
+    {
         $pipeString = trim($pipeString);
-        if (strlen($pipeString) > 2 ) {
+        if (strlen($pipeString) > 2) {
             $char = substr($pipeString, 0, 1);
-            if (in_array($char, ["'","\""])) {
+            if (in_array($char, ["'", '"'])) {
                 $endChar = substr($pipeString, -1, 1);
                 if ($char == $endChar) {
-                    $pipeString = "[" . implode($char . "," . $char, explode("|", $pipeString)) ."]";
+                    $pipeString = '['.implode($char.','.$char, explode('|', $pipeString)).']';
                 }
             }
         }
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -69,16 +69,7 @@ class PermissionServiceProvider extends ServiceProvider
 
             $bladeCompiler->directive('hasanyrole', function ($arguments) {
                 list($roles, $guard) = explode(',', $arguments.',');
-                $roles = trim($roles);
-                if (strlen($roles) > 2 ) {
-                  $char = substr($roles, 0, 1);
-                  if (in_array($char, ["'","\""])) {
-                    $endChar = substr($roles, -1, 1);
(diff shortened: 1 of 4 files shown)


tests/BladeTest.php
```diff
@@ -142,6 +142,16 @@ public function the_hasanyrole_directive_will_evaluate_true_when_the_logged_in_u
     }
 
     /** @test */
+    public function the_hasanyrole_directive_will_evaluate_true_when_the_logged_in_user_does_have_some_of_the_required_roles_use_pipe_for_the_given_guard()
```
freekmurze: This method name is a bit too long for my taste. What do you think about losing the `use_pipe_for_the_given_guard` part? (same remarks for the other test you added)

src/PermissionServiceProvider.php
```diff
@@ -69,6 +69,16 @@ protected function registerBladeExtensions()
 
             $bladeCompiler->directive('hasanyrole', function ($arguments) {
                 list($roles, $guard) = explode(',', $arguments.',');
+                $roles = trim($roles);
```
freekmurze: I've not followed it in detail, but my gut says that this could / should be solved in a much simpler way. Since the tests prove that this work, it should be relatively save to attempt to refactor this.

Also: this logic is being repeated for the other blade directive. If it can't be made simpler, create a new function so that logic appears only once in the codebase.

tests/BladeTest.php
```diff
@@ -142,6 +142,16 @@ public function the_hasanyrole_directive_will_evaluate_true_when_the_logged_in_u
...
+    {
+        $guard = 'admin';
+
+        auth('admin')->setUser($this->getSuperAdmin());
+
+        $this->assertEquals('does have some of the roles', $this->renderView('guardHasAnyRolePipe', compact('guard')));
```
freekmurze: Could you also add a test that checks if `does not have any of the given roles` is returned when a user has none of the requested roles?

tests/BladeTest.php
```diff
@@ -184,6 +194,22 @@ public function the_hasallroles_directive_will_evaluate_true_when_the_logged_in_
...
+
+        $admin = $this->getSuperAdmin();
+
+        $admin->assignRole('moderator');
+
+        $this->refreshTestAdmin();
+
+        auth('admin')->setUser($admin);
+
+        $this->assertEquals('does have all of the given roles', $this->renderView('guardHasAllRolesPipe', compact('guard')));
```
freekmurze: Could you add a tests the checks if `does not have all of the given roles` when the user does not have all the given roles?

tests/BladeTest.php
coderkk: OK, I will change it

tests/BladeTest.php
coderkk: OK, I will add another test for not match

tests/BladeTest.php
coderkk: OK, I will add another test for not match

src/PermissionServiceProvider.php
coderkk: Sure, no problem

coderkk:
Hi,  I have done the request change.

freekmurze:
Hi, thank you. I'm pulling this in an work a little bit on it myself.

PR #342, opened 2017-06-29
Feature/roles permissions map and command

dicarlosystems:
Hi,

I've added a couple of console commands and a configuration to help scaffold and update the system.  Wondering if you are interested in adding to the base?  Might be useful for other users, perhaps.  I know I found them useful.

Cheers,
Chris

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -86,4 +86,18 @@ return [
      */
 
     'log_registration_exception' => true,
+
+    /*
+     * Default roles and permissions
+     * 
+    */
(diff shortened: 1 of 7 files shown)


freekmurze:
Hi,

thank you for this but I'm going to pass. I'm thinking the bulk of users of this package don't need this. If you want to share this logic, create and maintain your package with this code.

dicarlosystems:
No worries, I'll continue to maintain my fork then!

PR #344, opened 2017-07-01
Update README.md

megawubs:
I needed to dive into the source code to find out how `@hasallroles` or `@hasanyroles` should be used with multiple values as it was not documented in the readme. With this MR it now is.

- Added usage of pipe for blade directives `@hasallroles()` and `@hasanyroles()`.


freekmurze:
Thanks!

PR #353, opened 2017-07-11
Create scope that contains all roles except role[s]

whfamily2006:


README.md
```diff
@@ -212,6 +212,14 @@ $users = User::role('writer')->get(); // Only returns users with the role 'write
 
 The scope can accept a string, a `\Spatie\Permission\Models\Role` object or an `\Illuminate\Support\Collection` object.
 
+The `HasRoles` also adds a scope to your models to scope the query to except roles:
```
drbyte: 'to except roles' might be better stated as "to exclude roles".

README.md
```diff
@@ -212,6 +212,14 @@ $users = User::role('writer')->get(); // Only returns users with the role 'write
...
+
+```php
+$users = User::exceptRole('writer')->get(); // Only returns users which not having the role 'writer'.
+```
+
+The scope can accept a array.
```
drbyte: "a array" should be "an array"

README.md
```diff
@@ -212,6 +212,14 @@ $users = User::role('writer')->get(); // Only returns users with the role 'write
 
 The scope can accept a string, a `\Spatie\Permission\Models\Role` object or an `\Illuminate\Support\Collection` object.
 
+The `HasRoles` also adds a scope to your models to scope the query to except roles:
+
+```php
+$users = User::exceptRole('writer')->get(); // Only returns users which not having the role 'writer'.
```
drbyte: "which not having the role" should be "which do not have the role"

freekmurze:
Thank you for this, but I'm going to pass. 

I feel that the majority of users don't need this scope and those who do can easily add it into their own project.

whfamily2006:
@freekmurze should I change according to @drbyte or not?

freekmurze:
I'm not going to pull in your PR, but sure you can follow @drbyte 's suggestion in your own project.

Issue #335, opened 2017-06-25
Field 'guard_name' doesn't have a default value


kjdion84:
Getting this error simply trying to create a role.

My models:

	namespace App;

	class Role extends \Spatie\Permission\Models\Role
	{
	    protected $fillable = ['name'];
	    protected $hidden = ['guard_name', 'pivot'];
	}

	class Permission extends \Spatie\Permission\Models\Permission
	{
	    protected $hidden = ['guard_name', 'pivot'];
	}

I have updated the permission config to use these models.

My controller methods:

    public function createForm()
    {
        $permissions = Permission::all();

        return view('roles.create', compact('permissions'));
    }

    public function create()
    {
        $this->validate(request(), [
            'name' => 'required|unique:roles',
        ]);

        $role = Role::create(request()->all());
        $role->syncPermissions(request()->input('permissions', []));
        activity()->by(auth()->user())->withProperties($role->fresh('permissions'))->log('Created Role');
        request()->session()->flash('status', 'Role created!');

        return redirect()->route('roles');
    }

And my form:

    <form class="form-horizontal" role="form" method="POST" action="{{ route('roles.create') }}">
        {{ csrf_field() }}

        <div class="form-group{{ $errors->has('name') ? ' has-error' : '' }}">
            <label for="name" class="col-md-3 control-label">Name</label>
            <div class="col-md-6">
                <input type="text" name="name" id="name" class="form-control" value="{{ old('name') }}">
            </div>
        </div>

        <div class="form-group">
            <label class="col-md-3 control-label">Permissions</label>
            <div class="col-md-6">
                @foreach ($permissions as $permission)
                    <div class="checkbox">
                        <label>
                            @php $checked = in_array($permission->name, old('permissions', [])) @endphp
                            <input type="checkbox" name="permissions[]" value="{{ $permission->name }}"{{ $checked ? ' checked' : '' }}> {{ $permission->name }}
                        </label>
                    </div>
                @endforeach
            </div>
        </div>

        <div class="form-group">
            <div class="col-md-6 col-md-offset-3">
                <button type="submit" class="btn btn-primary">Create</button>
            </div>
        </div>
    </form>

This makes no sense. It should work should it not? Yet I get error:

	SQLSTATE[HY000]: General error: 1364 Field 'guard_name' doesn't have a default value (SQL: insert into `roles` (`name`, `updated_at`, `created_at`) values (Test Role, 2017-06-25 19:03:26, 2017-06-25 19:03:26))

[...]

PR #355, opened 2017-07-11
Allow the `guard_name` columns to be nullable.

vaughany:
As per my comment here (https://github.com/spatie/laravel-permission/issues/335#issuecomment-314591736) I came across a problem where I couldn't create a role with the example in the readme (`$role = App\Role::create(['name' => 'writer']);`) as there was no text passed in for `guard_name`. This PR fixes this.

### DIFF
--- a/database/migrations/create_permission_tables.php.stub
+++ b/database/migrations/create_permission_tables.php.stub
@@ -18,14 +18,14 @@ class CreatePermissionTables extends Migration
         Schema::create($tableNames['permissions'], function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
-            $table->string('guard_name');
+            $table->string('guard_name')->nullable();
             $table->timestamps();
         });
 
         Schema::create($tableNames['roles'], function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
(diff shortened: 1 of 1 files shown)


freekmurze:
It seems that under normal circumstances, the `guard_name` should be filled automatically with the active guard name.

https://github.com/spatie/laravel-permission/blob/21079a4/src/Models/Role.php#L33

So I'm a bit hesitant to accept your PR.

vaughany:
Understood. I may well have been doing something incorrectly. I will investigate further.

vaughany:
Sorry @freekmurze, was a mistake on my part which caused the errors I was trying to correct.

jelgervisser:
@vaughany I'm facing the same problem. What was your mistake? Can't find any bugs in my code...

AlexVanderbist:
@jelgervisser Any errors? Are you sure the `auth.defaults.guard` config key is set correctly in `config/auth.php`?

vaughany:
Embarrassingly @jelgervisser, I believe I missed `use Spatie\Permission\Traits\HasRoles;` from the top of my user model and `use HasRoles'` from within the user class.

PR #367, opened 2017-07-21
Update README.md; RoleMiddleware.php

cesargb:
Permission now is not required and Roles can pipe for OR roles

### DIFF
--- a/README.md
+++ b/README.md
@@ -490,7 +490,7 @@ public function handle($request, Closure $next, $role, $permission=null)
         return redirect($urlOfYourLoginPage);
     }
 
-    $role=(!is_array($role)) ? preg_split('/\|/', $role) : $role;
+    $role=(is_array($role) ? $role : explode('|', $role));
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -484,18 +484,19 @@ use Auth;
...
-public function handle($request, Closure $next, $role, $permission)
+public function handle($request, Closure $next, $role, $permission=null)
 {
     if (Auth::guest()) {
         return redirect($urlOfYourLoginPage);
     }
 
-    if (! $request->user()->hasRole($role)) {
-       abort(403);
+    $role=(!is_array($role)) ? preg_split('/\|/', $role) : $role;
```
freekmurze: Could you use `explode` here?

cesargb:
Yes sure, in this moment i think in user a regex with other option, perhaps /[\|,]/ and can split by other character and get other behavior. But  you are right and is more efficient.
I change  this, thanks!

freekmurze:
Thanks!

PR #371, opened 2017-07-24
Updating composer.json to be compatible with future release of laravel 5.5

lk77:
Hello,

i'm updating requirements from ~5.4.0 to ~5.4 to allow the future laravel 5.5 version

thanks

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,10 +20,10 @@
     ],
     "require": {
         "php" : ">=7.0",
-        "illuminate/auth": "~5.4.0",
-        "illuminate/container": "~5.4.0",
-        "illuminate/contracts": "~5.4.0",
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this, but I'm holding out to make any changes around this until L5.5 is publicly available.

lk77:
i need to know if this package will work with laravel 5.5 before it's publicly available,
why not create a 5.5.x-dev branch so we can install laravel 5.5 right now and see if it works ?

freekmurze:
Because I simply don't have the time to create and maintain separate L5.5 dev branches for all our packages. I prefer to do all the necessary work for all our stuff shortly after L5.5 is publicly available.

For now you can fork the repo, do the necessary modifications and use that in your projects.

Issue #374, opened 2017-07-26
commands are not using the models set in the config


ronnievisser:
I want to overwrite the Role and permission model so I can use a different database connection but the commands don't keep this into account.

is there a reason for this?

freekmurze:
You can manually set the `connection` property on a model to specify which connection to use.

I don't know what how you expect the commands to do this.

ronnievisser:
I believe you haven't actually read the issue...

the commands are not using the models specified in config. i have overriden the config with custom models which extends yours. on my custom model I have set the connection property.

freekmurze:
Ok, got it now. We'll fix this soon.

Vimiso:
I believe the `php artisan permission:show` command has been forgotten about.

drbyte:
@Vimiso Fixed in 3.12.0

PR #376, opened 2017-07-26
[WIP] Use configured Permission model instead

deleugpn:
This is a concept on a possible approach for #374


freekmurze:
The problem has already been fixed, but thanks anyway for your work on this.

PR #379, opened 2017-07-27
Add middleware

cesargb:
I hope that this changes will be usefull. I add 2 middelware for greater flexibility. Please, feel free to modify or cancel. Thanks for your time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,7 +477,7 @@ You can use all of the blade directives listed in [using blade directives](#usin
 
 ## Using a middleware
 
-This package comes with RoleMiddleware and PermissionMiddleware middleware. You can add them inside your app/Http/Kernel.php file.
+This package comes with `RoleMiddleware` and `PermissionMiddleware` middleware. You can add them inside your `app/Http/Kernel.php` file.
 
 ```php
 protected $routeMiddleware = [
--- a/README.md
+++ b/README.md
@@ -477,54 +477,28 @@ You can use all of the blade directives listed in [using blade directives](#usin
 
 ## Using a middleware
 
-The package doesn't include a middleware to check permissions but it's very trivial to add this yourself:
-
-``` bash
-$ php artisan make:middleware RoleMiddleware
-```
-
-This will create a `app/Http/Middleware/RoleMiddleware.php` file for you, where you can handle your role and permissions check:
-
-```php
-use Auth;
-
-// ...
-
-public function handle($request, Closure $next, $role, $permission=null)
-{
-    if (Auth::guest()) {
-        return redirect($urlOfYourLoginPage);
-    }
-
-    $role = is_array($role)
-        ? $role
-        : explode('|', $role);
-
-    if (! $request->user()->hasAnyRole($role)) {
-        abort(403);
-    }
-
-    if ($permission && ! $request->user()->can($permission)) {
-        abort(403);
-    }
-
-    return $next($request);
-}
-```
-
-Don't forget to add the route middleware to `app/Http/Kernel.php` file:
+This package comes with RoleMiddleware and PermissionMiddleware middleware. You can add them inside your app/Http/Kernel.php file.
 
 ```php
 protected $routeMiddleware = [
     // ...
-    'role' => \App\Http\Middleware\RoleMiddleware::class,
+    'role' => \Spatie\Permission\Middleware\RoleMiddleware::class,
+    'permission' => \Spatie\Permission\Middleware\PermissionMiddleware::class,
 ];
 ```
 
 Now you can protect your routes using the middleware you just set up:
 
 ```php
-Route::group(['middleware' => ['role:admin,access_backend']], function () {
+Route::group(['middleware' => ['role:admin']], function () {
+    //
+});
+
+Route::group(['middleware' => ['permission:access_backend']], function () {
+    //
+});
+
+Route::group(['middleware' => ['role:admin','permission:access_backend']], function () {
     //
 });
 ```
--- a/src/Middleware/PermissionMiddleware.php
+++ b/src/Middleware/PermissionMiddleware.php
@@ -1,8 +1,9 @@
 <?php
+
 namespace Spatie\Permission\Middleware;
 
-use Closure;
 use Auth;
+use Closure;
 
 class PermissionMiddleware
 {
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -477,54 +477,28 @@ You can use all of the blade directives listed in [using blade directives](#usin
...
-    if ($permission && ! $request->user()->can($permission)) {
-        abort(403);
-    }
-
-    return $next($request);
-}
-```
-
-Don't forget to add the route middleware to `app/Http/Kernel.php` file:
+This package comes with RoleMiddleware and PermissionMiddleware middleware. You can add them inside your app/Http/Kernel.php file.
```
freekmurze: Please wrap all class names and paths in `backticks`

src/Middleware/PermissionMiddleware.php
```diff
@@ -0,0 +1,22 @@
...
+
+class PermissionMiddleware
+{
+    public function handle($request, Closure $next, $permission)
+    {
+        if (Auth::guest()) {
+            return $next($request);
+        }
+
+        if ($permission && ! Auth::user()->can($permission)) {
```
freekmurze: It seems odd that the `RoleMiddleware` can accept multiple roles, but this middleware cannot.

Could you add support for multiple permissions?

src/Middleware/PermissionMiddleware.php
```diff
@@ -0,0 +1,22 @@
...
+
+use Auth;
+use Closure;
+
+class PermissionMiddleware
+{
+    public function handle($request, Closure $next, $permission)
+    {
+        if (Auth::guest()) {
+            return $next($request);
```
freekmurze: This doesn't seem right. It looks like guest can now access protected page. Could you add a test that makes sure that guest cannot access protected pages and fix this?

src/Middleware/RoleMiddleware.php
```diff
@@ -0,0 +1,26 @@
...
+
+use Auth;
+use Closure;
+
+class RoleMiddleware
+{
+    public function handle($request, Closure $next, $role)
+    {
+        if (Auth::guest()) {
+            return $next($request);
```
freekmurze: This doesn't seem right. It looks like guest can now access protected page. Could you add a test that makes sure that guest cannot access protected pages and fix this?

tests/MiddlewareTest.php
```diff
@@ -0,0 +1,110 @@
...
+{
+    public function setUp()
+    {
+        parent::setUp();
+
+        Auth::login($this->testUser);
+    }
+
+    /** @test */
+    public function it_can_access_if_user_has_role()
```
freekmurze: Could you change `if_user` to `if_the_user` in every added test?

tests/MiddlewareTest.php
```diff
@@ -0,0 +1,110 @@
...
+        $request = new Request();
+        $middleware = new RoleMiddleware($this->app);
+
+        $result = $middleware->handle($request, function ($request) {
+            return (new Response())->setContent('<html></html>');
+        }, 'testRole|testrole2');
+    }
+
+    /** @test */
+    public function it_can_access_if_user_has_permiss()
```
freekmurze: `permiss` --> `permission`

src/Middleware/PermissionMiddleware.php
cesargb: @freekmurze I need resolve this  line  ```return redirect($urlOfYourLoginPage);```
My first thought was create a config element to replace $urlOfYourLoginPage. But then why this work not would correspond to middelware Auth?

But when you say this, I understand your position. 

I return my initial idea, add config params to define url to redirect. Perhaps ... ?

```php
if (Auth::guest()) {
    return redirect(config('permission.url_redirect_guest', '/login'));
```

freekmurze: The simplest solution would be to just `abort(403)`. People can do the redirection in the exception handler if they wish.

cesargb:
Hi @freekmurze I was see this commit: https://github.com/spatie/laravel-permission/commit/f590a3f006a85435a1636d5cbee132383c524fd9

I think that, now we can change this code in `PermissionMiddleware.php`:

```php
        if (! Auth::user()->hasAnyPermission(...$permission)) {
            abort(403);
        }
```

for this other:

```php
        if (! Auth::user()->hasAnyPermission($permission)) {
            abort(403);
        }
```

Perhaps, the function `hasAnyRole`, Should I have the same behavior?

emielmolenaar:
Nice changes! :clap:

tests/MiddlewareTest.php
```diff
@@ -0,0 +1,196 @@
...
+use Illuminate\Http\Request;
+use Illuminate\Http\Response;
+use Spatie\Permission\Middleware\RoleMiddleware;
+use Spatie\Permission\Middleware\PermissionMiddleware;
+
+class MiddlewareTest extends TestCase
+{
+    /**
+     * @test
+     * @expectedException Symfony\Component\HttpKernel\Exception\HttpException
```
freekmurze: I'm not a fan of using the `expectedException` annotation. Could you use the equivalent method instead?

tests/MiddlewareTest.php
```diff
@@ -0,0 +1,196 @@
...
+     */
+    public function it_can_not_access_with_middleware_role_if_is_guest()
```
freekmurze: The name of this test feels a bit off. Could you make this more readable eg `a_guest_cannot_access_a_route_protected_by_the_role_middleware` or something better?

freekmurze: Do this for all new test methods

tests/MiddlewareTest.php
freekmurze: Remove the annotation on the other new methods as well.

tests/MiddlewareTest.php
```diff
@@ -0,0 +1,196 @@
...
+    {
+        $request = new Request();
+        $middleware = new RoleMiddleware($this->app);
+
+        $result = $middleware->handle($request, function ($request) {
```
freekmurze: That block

```
$middleware->handle($request, function ($request) {
             return (new Response())->setContent('<html></html>');
         }, 'testRole');
```

is something that is used a lot.

Wrap it into a protected helper method called `testMiddleware` or something better.

```php
function testMiddleware($middleware, $parameter) {
   $middleware->handle($request, function () {
        return (new Response())->setContent('<html></html>');
    }, $parameter)
}
```

freekmurze:
Thanks, I'll fix the remaining nitpicks myself

PR #389, opened 2017-08-02
Demonstrate that the hasanyrole Blade directive does not work with pipes

emielmolenaar:
Just some quick hax to prove there is something wrong with @hasanyrole() :smile_cat: 

The docs say we should use @hasanyrole like this:

`@hasanyrole('admin|writer')`

But the tests pass an array instead of a string to @hasanyrole.

### DIFF
--- a/tests/BladeTest.php
+++ b/tests/BladeTest.php
@@ -117,7 +117,7 @@ class BladeTest extends TestCase
 
         auth()->setUser($this->getMember());
 
-        $this->assertEquals('does not have any of the given roles', $this->renderView('hasAnyRole', compact('roles')));
+        $this->assertEquals('does not have any of the given roles', $this->renderView('hasAnyRole', ['roles' => implode('|', $roles)]));
     }
 
     /** @test */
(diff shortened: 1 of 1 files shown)


tests/BladeTest.php
```diff
@@ -117,7 +117,7 @@ public function the_hasanyrole_directive_will_evaluate_false_when_the_logged_in_
 
         auth()->setUser($this->getMember());
 
-        $this->assertEquals('does not have any of the given roles', $this->renderView('hasAnyRole', compact('roles')));
+        $this->assertEquals('does not have any of the given roles', $this->renderView('hasAnyRole', ['roles' => implode('|', $roles)]));
```
freekmurze: Could you, instead of replace the current assertions, create a PR that adds your news ones. You can put them in the same methods as the current assertions.

freekmurze:
There seems to be something very off with the blade directives and using a string with a pipe character.

The problems is that [in this line](https://github.com/spatie/laravel-permission/blob/master/src/PermissionServiceProvider.php#L79) the `$roles` variable still contains a string "$roles". It will be resolved to the contents of the real `$roles` variable when the blade view is rendered.  The conversions from the pipestring to an array of roles should be done later.

I'd accept a PR that fixes this.

PR #392, opened 2017-08-03
Bugfix HasRoles.php

borger:
Issue happens on 7.1.7 when using variadic like: func(...$permissions) and then using this variable on a foreach. It does not split the array. This is an PHP bug for what I've seen.


borger:
Oops.. I apologize.. I didn't see you had fixed it already.. I was using an older version.. 

That's a great package by the way, thank you!

Issue #359, opened 2017-07-13
PermissionAlreadyExists - A `users_manage` permission already exists for guard `web`.  


connor11528:
I reset my database and deleted all the tables. When I rerun the migration with the seeders I get an error that a permission already exists. `php artisan migrate:reset` runs fine but does not fix the error. How can I reseed the database and start fresh? Thank you

```
php artisan migrate --seed
Migrating: 2014_10_12_000000_create_users_table
Migrated:  2014_10_12_000000_create_users_table
Migrating: 2014_10_12_100000_create_password_resets_table
Migrated:  2014_10_12_100000_create_password_resets_table
Migrating: 2017_07_12_145959_create_permission_tables
Migrated:  2017_07_12_145959_create_permission_tables
Migrating: 2017_07_12_230054_create_products_table
Migrated:  2017_07_12_230054_create_products_table
Seeding: PermissionSeed

                                                               
  [Spatie\Permission\Exceptions\PermissionAlreadyExists]       
  A `users_manage` permission already exists for guard `web`.
```

belzaaron:
Try `php artisan migrate:refresh --seed`.
Also ensure in your `PermissionSeed` that you are ONLY defining one instance of `users_manage`.
Let me know how it goes!

connor11528:
Same error with `php artisan migrate:refresh --seed`

My PermissionSeed file looks like this:

```
<?php

use Illuminate\Database\Seeder;
use Spatie\Permission\Models\Permission;

class PermissionSeed extends Seeder
{
    /**
     * Run the database seeds.
     *
     * @return void
     */
    public function run()
    {
        Permission::create(['name' => 'users_manage']);
    }
}
```

If I comment out the Permission::create seed it moves onto the RoleSeed and I get an error: `There is no permission named `users_manage` for guard `web`. `

The RoleSeed looks like this:

```
<?php

use Illuminate\Database\Seeder;
use Spatie\Permission\Models\Role;

class RoleSeed extends Seeder
{
    /**
     * Run the database seeds.
     *
     * @return void
     */
    public function run()
    {
        $role = Role::create(['name' => 'administrator']);
        $role->givePermissionTo('users_manage');
    }
}
```

connor11528:
I had to clear the permission cache!

Running:

```
php artisan cache:forget spatie.permission.cache
php artisan cache:clear  
```

Got me going again! Thank you

https://github.com/spatie/laravel-permission/issues/231

pablofior:
[...]

PR #396, opened 2017-08-04
Add example of Database Seeding which also clears cache first

drbyte:
Since there are a lot of issues posted about cache errors (see #359 for a recent one) when doing seeding, I thought an example might help avoid confusion.


freekmurze:
Great addition, thanks!

pablofior:
For me sometimes it works and sometimes it doesnt =(((
Very anoying bug

PR #400, opened 2017-08-06
Allow pipe symbols in @hasanyrole and @hasallroles blade directives

### LINKED ISSUE #380: Blade @hasanyrole('writer|admin') wrong result
Hi,
When I am trying the following code with a user having an admin role set
`
@hasanyrole('writer|admin')
  can do it
@else
  can't do it 
@endhasanyrole
`
I get "can't do it" but if I change to
`@hasanyrole('admin')`
I get "can do it"
I expected the same result
The way I solve it for now is passing an array like this
`@hasanyrole(['admin', 'teacher'])`

drbyte:
Fixes #380
Closes #389

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -76,7 +76,6 @@ class PermissionServiceProvider extends ServiceProvider
 
             $bladeCompiler->directive('hasanyrole', function ($arguments) {
                 list($roles, $guard) = explode(',', $arguments.',');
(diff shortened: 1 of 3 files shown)


freekmurze:
Much appreciated, Chris! 👍

emielmolenaar:
Nice, thanks

lucabartoli:
Wow, I was just going to work on this. Great, Chris!

Issue #362, opened 2017-07-19
add FK for `model_id`


ctf0:
atm the delete cascade if we removed the `permissions` or `roles` model but if we removed the other end model ex.`page, user, etc...` it wont be removed, 
which causes issues if we added an item then removed it cuz now the `model_has_permissions` & `model_has_roles` will already have a `model_id` equal to the newly created model.

a way around to fix this is, inside the controller delete method
```php
DB::table('model_has_roles')->where('model_id', $id)->delete();
DB::table('model_has_permissions')->where('model_id', $id)->delete();
```

freekmurze:
I'd accept a fully tested PR that fixes this.

ctf0:
i can make a PR, bu sadly dont know anything about writing tests 😢

freekmurze:
> but sadly dont know anything about writing tests

Research the subject a bit, your future self will be thankful

Take a look around in the code in the `tests` directory of this package. You can run `vendor/bin/phpunit` to run them.

bhill77:
may I make PR for this? or you will do it @ctf0?

ctf0:
@bhill77 its okay, you can make it 👍

bhill77:
sorry @ctf0, I don't really understand what you mean by deleting model. Do you mean delete model class? or what?

ctf0:
@bhill77  what i meant is if u tried to delete ex.`user` that have a permission or role, the user will be removed but the record in the pivot table will remain which may cuz issues if u tried to add a new user with the same permissions and roles to the db as the pivot records will be already in the table.

PR #401, opened 2017-08-07
HasRole enhancments

ctf0:
#### model events
- combined model events (created & updated) into saved
- https://github.com/spatie/laravel-permission/issues/362 [Thanx](https://tighten.co/blog/laravel-tip-bootable-model-traits)
- eager load `roles / perms` relations by default

<s>#### cache
- removed cache expiration as its not needed and creates unnecessary overhead
- update rdme

#### middleware
- fix permission middleware
- make middlewares usable with (comma or pipe)
- use `auth()` instead of the facade</s>

#### tests
...

### DIFF
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -14,10 +14,10 @@ class PermissionMiddleware
 
         if (is_array($permission)) {
             if (strpos($permission, '|')) {
-               $permission = explode('|', $permission);
+                $permission = explode('|', $permission);
             }
             if (strpos($permission, ',')) {
-               $permission = explode(',', $permission);
+                $permission = explode(',', $permission);
             }
         };
 
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -19,7 +19,7 @@ class PermissionMiddleware
             if (strpos($permission, ',')) {
                 $permission = explode(',', $permission);
             }
-        };
+        }
 
         if (! auth()->user()->hasAnyPermission($permission)) {
             abort(403);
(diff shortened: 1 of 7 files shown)


src/PermissionRegistrar.php
```diff
@@ -59,7 +59,7 @@ public function forgetCachedPermissions()
 
     public function getPermissions(): Collection
     {
-        return $this->cache->remember($this->cacheKey, config('permission.cache_expiration_time'), function () {
+        return $this->cache->forever($this->cacheKey, function () {
```
freekmurze: This is a breaking change. Please restore the old behaviour and old versions of readme / config

src/Middlewares/RoleMiddleware.php
```diff
@@ -3,21 +3,25 @@
...
 
 use Closure;
-use Illuminate\Support\Facades\Auth;
 
 class RoleMiddleware
 {
     public function handle($request, Closure $next, $role)
     {
-        if (Auth::guest()) {
+        if (auth()->guest()) {
```
freekmurze: I like the use of the facade better in this situation. Please revert back to the facade everywhere you replace it by the function.

src/Middlewares/PermissionMiddleware.php
```diff
@@ -3,21 +3,25 @@
...
     {
-        if (Auth::guest()) {
+        if (auth()->guest()) {
             abort(403);
         }
 
-        $permission = is_array($permission)
-            ? $permission
-            : explode('|', $permission);
+        if (is_array($permission)) {
```
freekmurze: I don't think these changes are necessary. Could you please revert back to the old behaviour?

src/PermissionRegistrar.php
ctf0: i removed it because cache gets cleared on each create/update/delete event, so why the extra clearing each 24hrs ?

ctf0:
@freekmurze any news on this ?

freekmurze:
Thank you for this.

I'm going to remove that `addGlobalScope` though as this results in much more queries being executed. You can easily add a `with` property in your own project if you want to load relations automatically.

PR #411, opened 2017-08-11
Roles and Permissions DB connection set in config file

wfeller:
The Role and Permission models use the DB connection that's set in the permission.php config file.

This way it uses the right connection and won't prefix the tables.

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -14,6 +14,7 @@ return [
          */
 
         'permission' => Spatie\Permission\Models\Permission::class,
+        'permissions_connection' => env('DB_CONNECTION', 'mysql'),
 
         /*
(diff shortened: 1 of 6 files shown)


config/permission.php
```diff
@@ -14,6 +14,7 @@
          */
 
         'permission' => Spatie\Permission\Models\Permission::class,
+        'permissions_connection' => env('DB_CONNECTION', 'mysql'),
```
freekmurze: Wouldn't it make more sense to use a custom model and use the `connection` instance variable to set the connection? Than we wouldn't need all this code.

wfeller: Ooh... that's right, my bad /cry

Sorry about that and tyvm :)

PR #414, opened 2017-08-13
Added support for Laravel 5.5

edmandiesamonte:
Updated `composer.json` to support Laravel 5.5


freekmurze:
Thanks, but this change has already been done on  the `laravel-55` branch.

Issue #406, opened 2017-08-09
Get role name from user.


devgowa:
Can you tell me how to get roles name from users model?

Thanks before.

sebastiandedeyne:
You can `pluck` them from the roles.

```php
$user->roles->pluck('name');
```

Fatih90:
@sebastiandedeyne but the result is like this
["admin"]

["writer"]

["vip"]

drbyte:
Since v2.5.1 you can use `$user->getRoleNames()`, which returns a collection of role names associated with that user. 
And then you can access that collection using all the normal collection methods.

Fatih90:
i like this project but i will stay with [laratrust](https://github.com/santigarcor/laratrust)

amitprasad2007:
Can you tell me how to get roles name from  Auth User while loing 

Thanks before.

Parterdev:
Hi guys! 
Another way to render "one" role name in a blade view is:
**Remember that users could have multiple assigned roles.**

@forelse ($user->roles->take(1) as $role)
        <td>{{$role->name}}</td>	
@empty
	<td>Not assigned</td>	
@endforelse

PR #415, opened 2017-08-13
Added helpers to get the user's role names (nicely formatted and raw versions).

vaughany:
I've noticed a few users opening issues (e.g. #406 ) to get the user's role's names, and it's something I wanted to do myself, so here's how I did it. If this PR is not accepted then at least the code is here and may help someone else.

`getRoleNames()` will return a string like `'Member'`, or `'Member, Intern, Key-holder'`.

`getRoleNamesRaw()` will return a string like `'["member"]'` or `'["member", "key-holder"]'`.

Simply pass in an instance of `App\User` within a Blade template, e.g.:

```
{{ getRoleNames( Auth::user() ) }}
```

I've included tests but due to my inexperience with developing packages for Laravel, could not get them running.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -25,6 +25,7 @@ function getRoleNames(App\User $user): string
     if ($user) {
         return ucwords($user->roles->pluck('name')->implode(', '));
     }
+
     return '';
 }
 
@@ -40,5 +41,6 @@ function getRoleNamesRaw(App\User $user): string
     if ($user) {
         return $user->roles->pluck('name');
     }
+
     return '';
 }
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -12,3 +12,33 @@ function getModelForGuard(string $guard)
             return config("auth.providers.{$guard['provider']}.model");
         })->get($guard);
 }
+
+/**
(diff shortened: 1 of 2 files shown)


drbyte:
Your tests are failing because you're using global helpers which expect a different kind of User object. 

It would be better in my opinion to NOT use global helpers. Instead, add one getter (ie: `getRoleNames()`) for retrieving raw role names to the HasRoles trait, which plucks names and returns a collection. 

Then in your own app if you want to format the returned collection to a string or to a `ucwords()` string, do that in your own code, not in the package.

This would also make your code testable (albeit with a bit of refactoring away from using the old global helpers).

Implemented cleanly like this might make it more likely to be accepted. Granted, I don't have any influence over those decisions, so can't promise anything.

Be sure to also add an entry to the README which mentions that `$user->getRoleNames()` will return a list of the user's assigned roles. There are already getters for retrieving all the user's permissions. 
(If one wants to get roles for a specific user, outside the model, they can use `User::find($id)->getRoleNames()`, instead of using a global helper.)

drbyte:
See addition in v2.5.1: #442

PR #418, opened 2017-08-15
README updated with Troubleshooting

bahjaat:
When using the database cache driver and a lot of permissions, the cache can't fit in the default laravel cache table. So you have to edit the migration file. Added some help under Troubleshooting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,18 @@ $role->givePermissionTo('edit articles');
 $role->revokePermissionTo('edit articles');
 ```
 
+### DecryptException
+
+Working with a lot of permissions, your cache can grow quickly. When using the database cache driver, you might get a `DecryptException`. In that case, just change your cache migration file from:
+
+```php
+$table->text('value');
+```
+to:
+```php
+`$table->longText('value');
+```
+
 ## Need a UI?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -596,6 +596,18 @@ $role->givePermissionTo('edit articles');
...
+### DecryptException
+
+Working with a lot of permissions, your cache can grow quickly. When using the database cache driver, you might get a `DecryptException`. In that case, just change your cache migration file from:
+
+```php
+$table->text('value');
+```
+to:
+```php
+`$table->longText('value');
```
drbyte: You're talking about the migration suggested by Laravel itself, right? 

ie: https://github.com/laravel/framework/blob/5.4/src/Illuminate/Cache/Console/stubs/cache.stub

... and nothing to do with the Spatie laravel-permission package, right?

bahjaat: That's right indeed. But maybe nice to have it here in the troubleshooting-section, because this package is also using Laravel's cache system.

freekmurze:
Thanks for this, but like @drbyte mentioned that error is not caused by this package, but by Laravel. Send a PR to their docs.

PR #423, opened 2017-08-17
Added test coverage for hasAnyRole in variadic format

Lloople:
I was going to add this as a feature but creating the test first I see that is already working 🙂


freekmurze:
Thanks!

PR #425, opened 2017-08-22
Update readme to remove duplication and standardize examples

drbyte:


freekmurze:
Great work 👍 

Thank you!

PR #426, opened 2017-08-22
Added scopePermission() to HasRoles

abiodunsulaiman694:
Allows us to find users by permission name.
For example,
$users = User::permission($permission_name)->get();

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -82,6 +82,41 @@ trait HasRoles
             });
         });
     }
+    
+    /**
+     * Scope the model query to certain permissions only.
+     *
+     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @param string|array|Role|\Illuminate\Support\Collection $permissions
+     *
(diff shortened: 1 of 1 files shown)


drbyte:
Two things:

1. Do you mind updating the HasPermissionsTest to include tests for this permission scope?  
Similar to https://github.com/spatie/laravel-permission/blob/master/tests/HasRolesTest.php#L142-L209

2. Are you aware that using a permission scope will ONLY fetch users with matching __direct__ permissions?  Such a scope *cannot* find users who have a permission granted via a role. 
I wonder if this potential "gotcha" is the reason why such a scope was not included in this package originally?

freekmurze:
> Such a scope cannot find users who have a permission granted via a role.
I wonder if this potential "gotcha" is the reason why such a scope was not included in this package originally?

Yup, I thought it's better not to include this to avoid tears down the road 😄

drbyte:
> Yup, I thought it's better not to include this to avoid tears down the road 😄

I tend to agree.

PR #429, opened 2017-08-24
Fixed issue with README formatting

SubodhDahal:
The extra spaces before the bash commands was messing up the README formatting in Github, so fixed that.


freekmurze:
👍

PR #431, opened 2017-08-29
Role groups

aTanCS:
More complex web applications need to divide roles into sections (or modules). E.g. group1 has section1 and section2 roles, group2 has section3 and section4 roles, etc. Each section role has its own permissions. I've added roles groups feature to laravel-permission that allows roles grouping in a described way. Permissions of child roles are visible from groups. Permissions can belong to groups. In other words. It is an additional permission level. Some group tests are missing, but could be added later. Most of work was a copypaste of Roles logic. For now I'd like just to know if the idea is correct. It works for me.

laravel-55 branch was used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to `laravel-permission` will be documented in this file
 
+## 2.4.1 -2017-08-05
+- Fix processing of pipe symbols in `@hasanyrole` and `@hasallroles` Blade directives
+
 ## 2.4.0 -2017-08-05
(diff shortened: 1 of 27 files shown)


aTanCS:
It look like it was not a good idea to merge it with laravel-55 branch. All its commits are in this pull request now. Or am I missing something? It is my first pull request.

freekmurze:
Thank you for this. It's clear that you put a lot of work into this.

I'd like to keep this package as simple as it can be.Support for grouping isn't important for the vast majority of users of this package.

If you want to use your changes in your application I suggest you fork our package and do the modifications in your fork.

Issue #356, opened 2017-07-12
About middleware...


abr4xas:
Hi there, 

Currently I have two users: Super Admin, Normal User

* Admin for obvious reasons, has all permissions
* The user only has two (see his own profile, update his own profile profile)

As far as I understand, the second parameter is the permission associated with the role that I send as the first parameter.

That makes me ask, is there any way I can validate that even if the permission is not associated with the role, I can perform the actions related to that permission?

This is my current route:

```php
Route::group(['prefix' => 'admin','middleware' => 'auth'], function () {
    Route::middleware(['role:User,view_profile'])->group(function () {
        Route::get('/home', 'HomeController@index')->name('admin.home');
    }); 
});
```
If I log in with my admin user I can not access that route even if I have the permission.

The same thing happens if I login with my normal user:
```php
Route::group(['prefix' => 'admin','middleware' => 'auth'], function () {
    Route::middleware(['role:Admin,view_profile'])->group(function () {
        Route::get('/home', 'HomeController@index')->name('admin.home');
    }); 
});
```

### What do I need to achieve?

Being able to enter certain routes regardless of whether I have the role, just having the permission is enough.

_This does not apply to all routes._

A practical example of what I want to achieve:

I need my normal user to enter your profile and update it (if applicable) and also I want you to have the possibility to write a post.

Obviously you must have an additional permission that would be `add_post`

My super admin having all the permissions obviously can enter and create their own post additional to the possibility of accepting the post that the users send.

Please @freekmurze can you helpme with this?

freekmurze:
Thanks for the explanation, but I'm not sure I understand your question correctly.

Are you aware that you can associate users with permissions directly? A role is just an optional thing. 

Laravel also allows some logic to be run before checking the permission. This is useful for creating super admin like capabilities. More info here: https://laravel.com/docs/5.4/authorization#gates

abr4xas:
@freekmurze thanks!!

GDanielRG:
so i created a superadmin role and gave it permission to "do everything". what do I need to do so that every permission check that I make, it allows a user with the permission "do everything" to do it eventhough the super admin user does not have specific permission to do that?

say i wanna check $user->hasPermissionTo('edit articles');
it should allow it because the user has the permission "do everything". I believe gates and auth help with this but I don't know how to implement it :(

abr4xas:
@GDanielRG Me neither, but what I did was to create the number of roles needed according to my application with the corresponding permissions and it works without problems.

[...]

PR #444, opened 2017-09-02
Change PermissionMiddleware to use cache

drbyte:
Calling `hasAnyPermission` directly will not trigger this package's cached data.
This PR changes it to call `$user->can` so that the Gate's defined rules are honoured.

Ref: discussion at https://github.com/spatie/laravel-permission/issues/356#issuecomment-326702605

### DIFF
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -13,12 +13,12 @@ class PermissionMiddleware
             abort(403);
         }
 
-        $permission = is_array($permission)
+        $permissions = is_array($permission)
             ? $permission
             : explode('|', $permission);
 
-        foreach ($permission as $p) {
-            if (Auth::user()->can($p)) {
+        foreach ($permissions as $permission) {
+            if (Auth::user()->can($permission)) {
                 return $next($request);
             }
         }
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -17,10 +17,12 @@ class PermissionMiddleware
             ? $permission
             : explode('|', $permission);
 
-        if (! Auth::user()->hasAnyPermission(...$permission)) {
-            abort(403);
+        foreach ($permission as $p) {
+            if (Auth::user()->can($p)) {
+                return $next($request);
+            }
         }
 
-        return $next($request);
+        abort(403);
     }
 }
--- a/tests/MiddlewareTest.php
+++ b/tests/MiddlewareTest.php
@@ -56,6 +56,11 @@ class MiddlewareTest extends TestCase
             $this->runMiddleware(
                 $this->roleMiddleware, 'testRole|testRole2'
             ), 200);
+
+        $this->assertEquals(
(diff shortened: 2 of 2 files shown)


src/Middlewares/PermissionMiddleware.php
```diff
@@ -17,10 +17,12 @@ public function handle($request, Closure $next, $permission)
             ? $permission
             : explode('|', $permission);
 
-        if (! Auth::user()->hasAnyPermission(...$permission)) {
-            abort(403);
+        foreach ($permission as $p) {
```
freekmurze: Let's rename `$permission` to `$permissions (this should be done on line 16 as well).

Generally speaking I don't like abbreviations. Could you rename `$p` to `$permission`?

drbyte: Agreed.

darwinluague: Actually `can` can accept array so no need to loop permissions. We can just replace `hasAnyPermission` to `can`. Like so, `if (! Auth::user()->can(...$permission)) {`

drbyte: True. Thanks.

src/Middlewares/PermissionMiddleware.php
```diff
@@ -13,14 +13,14 @@ public function handle($request, Closure $next, $permission)
...
         }
 
-        $permission = is_array($permission)
+        $permissions = is_array($permission)
             ? $permission
             : explode('|', $permission);
 
-        if (! Auth::user()->hasAnyPermission(...$permission)) {
-            abort(403);
+        if (Auth::user()->can(...$permissions)) {
```
freekmurze: Does this line work?

From the docs:

```
if ($user->can('create', Post::class)) {
    // Executes the "create" method on the relevant policy...
}
```

The second parameter should be a model, no mention about next parameters.

darwinluague: Based on the code the second parameter is optional.

https://github.com/laravel/framework/blob/5.5/src/Illuminate/Foundation/Auth/Access/Authorizable.php

drbyte: The tests are passing. Not sure if there are additional tests needed to prove the middleware is firing correctly. 

I have the code working on a production site, and it appears to be properly functional.

freekmurze: I'm thinking the middleware will now only work by chance if there is only one permission passed. It'll not work correctly when 2 or more permissions are passed and the 2nd, 3th, ... parameters is an ability the user can do. 

```php
if (Auth::user()->can(...$permissions)) {
```

should be changed to 

```php
if (Auth::user()->can($permissions)) {
```

Best to add some tests to prove that my thinking is correct/wrong 😄

drbyte: In playing with the tests, it appears you are correct :)

I've added some more tests: passing an array (even though most middleware calls will be string and not array), and re-ordering options so that the first option isn't the only one under test.

freekmurze:
Thanks!

PR #452, opened 2017-09-07
Register Blade extensions in boot instead of register

### LINKED ISSUE #427: Register Blade Extensions in boot() method, not in register()
Hi,

please change the ServiceProvider to register the blade extensions in the boot() method, not in register().

https://laravel.com/docs/5.4/blade#extending-blade

The current version does not work with packages like HTMLMin\HTMLMin because of that. When i use both packages, your blade extensions are not registered and output as text, e.g. `@hasanyrole...`

Best regards

drbyte:
Fixes #427


freekmurze:
Thanks Chris!

PR #454, opened 2017-09-08
Added permission scope to HasRoles trait

inet123:


drbyte:
This is very similar to what I had come up with. It appears to sufficiently allow the ability to scope users who are authorized for a specified permission whether directly or via a role. 

👍

src/Traits/HasRoles.php
```diff
@@ -84,6 +84,59 @@ public function scopeRole(Builder $query, $roles): Builder
...
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopePermission(Builder $query, $permissions): Builder
+    {
+        if ($permissions instanceof Collection) {
+            $permissions = $permissions->toArray();
+        }
+
+        if (! is_array($permissions)) {
```
freekmurze: We could use that shiny new `array_wrap` function here.

src/Traits/HasRoles.php
```diff
@@ -84,6 +84,59 @@ public function scopeRole(Builder $query, $roles): Builder
...
+            $permissions = [$permissions];
+        }
+
+        $rolesWithPermissions = collect([]);
+        $permissions = array_map(function ($permission) use (&$rolesWithPermissions) {
```
freekmurze: This could be cleaned up greatly. Right now this map operation does two things. Making sure the permissions are models and building up that `rolesWithPermissions`. Do this in two steps (with two map operations I guess) for clarity (so when can use that nasty `&`)

src/Traits/HasRoles.php
```diff
@@ -84,6 +84,59 @@ public function scopeRole(Builder $query, $roles): Builder
...
+            if ($permission instanceof Permission) {
+                $permissionModel = $permission;
+            } else {
```
freekmurze: O how I dislike `else` statements. Refactor this to it's own `protected` function where you can use an early return.

src/Traits/HasRoles.php
```diff
@@ -84,6 +84,59 @@ public function scopeRole(Builder $query, $roles): Builder
...
+                $permissionModel = app(Permission::class)->findByName($permission, $this->getDefaultGuardName());
+            }
+            $rolesWithPermissions = $rolesWithPermissions->merge($permissionModel->roles);
+
+            return $permissionModel;
+        }, $permissions);
+
+        $rolesWithPermissions = $rolesWithPermissions->unique();
+
+        return $query->
```
freekmurze: Looks good!

tests/HasPermissionsTest.php
```diff
@@ -52,4 +52,107 @@ public function it_can_revoke_a_permission_from_a_user()
...
         $this->assertFalse($this->testUser->hasPermissionTo($this->testUserPermission));
     }
+
+    /** @test */
+    public function it_can_scope_users_using_a_string()
+    {
+        $user1 = User::create(['email' => 'user1@test.com']);
+        $user2 = User::create(['email' => 'user2@test.com']);
+        $user1->givePermissionTo(['edit-articles', 'edit-news']);
+        // giving user2 permission throgth the role
```
freekmurze: remove this comment, I think the code is clear enough.

tests/HasPermissionsTest.php
```diff
@@ -52,4 +52,107 @@ public function it_can_revoke_a_permission_from_a_user()
 
         $this->assertFalse($this->testUser->hasPermissionTo($this->testUserPermission));
     }
+
+    /** @test */
+    public function it_can_scope_users_using_a_string()
```
freekmurze: I can always appreciate a PR that adds tests 👍

tests/HasPermissionsTest.php
```diff
@@ -52,4 +52,107 @@ public function it_can_revoke_a_permission_from_a_user()
...
+        $this->assertEquals($scopedUsers2->count(), 1);
+    }
+
+    /** @test */
+    public function it_can_scope_users_using_an_array()
+    {
+        $user1 = User::create(['email' => 'user1@test.com']);
+        $user2 = User::create(['email' => 'user2@test.com']);
+        $user1->givePermissionTo(['edit-articles', 'edit-news']);
+        // giving user2 permission throgth the role
```
freekmurze: Remove this comment

tests/HasPermissionsTest.php
```diff
@@ -52,4 +52,107 @@ public function it_can_revoke_a_permission_from_a_user()
...
+        $this->assertEquals($scopedUsers2->count(), 1);
+    }
+
+    /** @test */
+    public function it_can_scope_users_using_a_collection()
+    {
+        $user1 = User::create(['email' => 'user1@test.com']);
+        $user2 = User::create(['email' => 'user2@test.com']);
+        $user1->givePermissionTo(['edit-articles', 'edit-news']);
+        // giving user2 permission throgth the role
```
freekmurze: You know what do to 😄

src/Traits/HasRoles.php
```diff
@@ -84,6 +84,59 @@ public function scopeRole(Builder $query, $roles): Builder
...
+     * Scope the model query to certain permissions only.
+     *
+     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @param string|array|\Spatie\Permission\Contracts\Permission|\Illuminate\Support\Collection $permissions
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function scopePermission(Builder $query, $permissions): Builder
+    {
+        if ($permissions instanceof Collection) {
```
freekmurze: Let's put everything between line 96 and 108 in it's own dedicated protected function `convertToPermissionModels`

src/Traits/HasRoles.php
```diff
@@ -84,6 +84,59 @@ public function scopeRole(Builder $query, $roles): Builder
...
+            if ($permission instanceof Permission) {
+                return $permission;
+            }
+
+            return app(Permission::class)->findByName($permission, $this->getDefaultGuardName());
+        }, $permissions);
+
+        $rolesWithPermissions = collect([]);
+
+        foreach ($permissions as $permission) {
```
freekmurze: seems to me that this is a reduce operation. Let's use `array_reduce` here.

freekmurze:
Last remark: could you also add a little explanation + example of the new functionality in the readme?

freekmurze:
Thanks!

inet123:
my pleasure, thank you for the notes.

PR #462, opened 2017-09-12
Sync Permissions after create a new permission

Majunko:
Sync Permissions to Admin after create a new permission


AlexVanderbist:
Hi! 

First of all thanks for contributing. However not everyone using the package will have the same roles and set-up. We can't just assume that there's an `Admin` role and assign it all permissions. That's application logic to be added in userland.

PR #472, opened 2017-09-25
Add documentation on multiple roles and permissions

ttomdewit:
Document to use of multiple roles and permissions in a middleware group. This was undocumented as far as I am aware.


drbyte:
You are correct. Thanks!

PR #475, opened 2017-09-26
Documentation updates

drbyte:
These updates are in response to a few recently posted issues.

- fixed a few ` symbols
- clarified blade examples to demonstrate that a collection can be passed, and avoided using facade in example to avoid having to import the class into blade
- added another pipe-separated entry to middleware examples for demo purposes
- added section on re-registering with Gate when unit testing
- tiny wording tweaks to cache tips


freekmurze:
Thank you, I appreciate this very very much! Beers are on me at Laracon 2018

PR #478, opened 2017-09-28
add laravel 5.3 support

mostafamaklad:
Laravel 5.3 supported now in this version


freekmurze:
Thanks!

Pulling it in for now, but as soon as we need features not available in L5.3, we'll drop support it.

drbyte:
Note: Travis here isn't running tests for package combinations lower than Laravel 5.4.0

PR #502, opened 2017-10-12
Don't detach roles and permissions on soft delete

### LINKED ISSUE #501: Soft deleting
I'm using soft deletes to lock out users but attached roles get deleted too. Is there a way to keep the roles attached and still be able to use soft deletes?

AlexVanderbist:
Fixes #501

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to `laravel-permission` will be documented in this file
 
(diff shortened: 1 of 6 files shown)


freekmurze:
Kinda hate `SoftDeletes` but this seems an ok fix to me. Wouldn't mind dropping support for `SoftDeletes` in the next major version.

Issue #384, opened 2017-07-31
Permission for super admin [Question]


azeemhassni:
Hi, I'm trying to use your package in one of my projects. I've a simple question is there something like "giving every permission to a super admin role" exists. 

does the following code makes sense in this context?
```php
  $superAdmin = Spatie\Permission\Models\Role::create(['name' => 'Super Admin']);
  $superAdmin->givePermissionTo('*');
```

freekmurze:
You can use the gate's `before` method to achieve that behaviour:

```
$gate->before(function ($user, $ability) {
    if ($user->isSuperAdmin()) {
        return true;
    }
});
```

More info: https://laravel.com/docs/5.1/authorization

drbyte:
@azeemhassni  ... so, for example, in AuthServiceProvider:

```php
use Illuminate\Contracts\Auth\Access\Gate as GateContract;

...

    public function boot(GateContract $gate)
    {
        $this->registerPolicies();

        // Grant "Super Admin" users all permissions (assuming they are verified using can() and other gate-related functions):
        $gate->before(function ($user, $ability) {
            if ($user->hasRole('Super Admin')) {
                return true;
            }
        });
    }
```

cfreak:
How can i do that in Laravel 5.5?

Does anyone have an hint for me?

drbyte:
Pretty much the same. Just use the Gate facade:

```diff
    public function boot()
    {
        $this->registerPolicies();

+        // Grant "Super Admin" users all permissions (assuming they are verified using can() and other gate-related functions):
-        $gate->before(function ($user, $ability) {
+        Gate::before(function ($user, $ability) {
+            if ($user->hasRole('Super Admin')) {
+                return true;
+            }
+        });
    }
```

azeemhassni:
@drbyte why do you think using the facade would be better?

drbyte:
... because otherwise you'll need to inject a $gate object as I did in my previous example, above.

azeemhassni:
That was exactly my question. Why using the facade is better than DI? is there a performance improvement or it's better just because you save yourself type hinting a class?

drbyte:
Just simpler. 

If you want to use Dependency Injection, that's fine too. Do what suits your app's needs best.

asnawisaharuddin:
After defining Gate::before it will not trigger before method in the policy class if Gate::before return true. 

is there a way to only allow super admin bypass only permissions defined in this library?

[...]

PR #505, opened 2017-10-12
Rewrite PermissionRegistrar

### LINKED ISSUE #480: Service Provider shouldn't access the DB
When PermissionServiceProvider calls PermissionRegistrar::registerPermissions, it creates a delay if a database is not defined or the server times out, etc.  I noticed this because artisan commands were taking forever—even if just calling --help.  I believe that a Service Provider should only register functions and never perform any actual work or connect to external resources.

Perhaps instead of doing it this way, we could use a before hook on the gate like this:
```php
$this->gate->before(function () {
    if (!static::$permissionsRegistered) {
        try {
            $this->getPermissions()->map(function (Permission $permission) {
                $this->gate->define($permission->name, function ($user) use ($permission) {
                    return $user->hasPermissionTo($permission);
                });
            });

            static::$permissionsRegistered = true;
        } catch (Exception $exception) {
            if ($this->shouldLogException()) {
                $this->logger->alert(
                    "Could not register permissions because {$exception->getMessage()}" . PHP_EOL .
                    $exception->getTraceAsString()
                );
            }
        }
    }
});
```
This way, it will only load the permissions once, the first time a Gate check is performed.

AlexVanderbist:
This PR rewrites the `registerPermissions` to use a before callback on the `Gate` instead of registering all permissions at once. 

This way we don't have to load all permissions when the `PermissionServiceProvider` is registered. As an added bonus we can drop the `log_registration_exception` config value and related code.

Fixes #480

I'm not sure if this still counts as a patch or minor version? Apart from removing the config value, rewriting some tests and the actual changes to `registerPermissions` nothing changed. Need some feedback on this.


src/PermissionRegistrar.php
```diff
@@ -17,39 +17,27 @@ class PermissionRegistrar
...
+                if (method_exists($user, 'hasPermissionTo')) {
+                    return $user->hasPermissionTo($ability);
+                }
+            } catch (PermissionDoesNotExist $e) {
             }
+        });
 
-            return false;
-        }
+        return true;
```
drbyte: Is this `return true;` needed?

I know it used to return a boolean that was checked in dozens of tests (never understood why that was done), but it appears you've removed that check from the tests, so this `true` may not be needed.

tests/CacheTest.php
```diff
@@ -20,92 +22,131 @@ public function setUp()
...
+        $this->registrar->getPermissions();
 
-        $this->registrar->registerPermissions();
+        $this->assertQueryCount(self::QUERIES_PER_CACHE_PROVISION);
 
-        $this->assertCount(2, DB::getQueryLog());
+        $this->registrar->getPermissions();
 
-        DB::flushQueryLog();
+        $this->assertQueryCount(self::QUERIES_PER_CACHE_PROVISION);
```
drbyte: If it's been cached, shouldn't the queryCount be 0?

tests/CacheTest.php
```diff
@@ -20,92 +22,131 @@ public function setUp()
...
     /** @test */
-    public function permission_creation_and_updating_should_flush_the_cache()
+    public function it_flushes_the_cache_when_updating_a_permission()
     {
         $permission = app(Permission::class)->create(['name' => 'new']);
-        $this->assertCount(1, DB::getQueryLog());
 
-        $this->registrar->registerPermissions();
-        $this->assertCount(3, DB::getQueryLog());
+        $this->resetQueryCount();
```
drbyte: Isn't this `reset` a duplicate of line 61 below?

src/PermissionRegistrar.php
drbyte: Update: I see, you've declared its return type as boolean, so it's gotta return "something" boolean.

tests/CacheTest.php
```diff
@@ -20,92 +22,131 @@ public function setUp()
...
     public function user_creation_should_not_flush_the_cache()
     {
+        $this->registrar->getPermissions();
+
         User::create(['email' => 'new']);
-        $this->assertCount(1, DB::getQueryLog());
 
-        $this->registrar->registerPermissions();
-        $this->assertCount(1, DB::getQueryLog());
+        $this->resetQueryCount(1);
```
drbyte: `resetQueryCount()` doesn't take any parameters, so the `1` here is superfluous.

tests/CacheTest.php
drbyte: Sigh - I missed that you didn't do a "reset" in between, unlike the other tests. :)
So, this is correct: should return the same as it did 2 statements earlier.
👍

config/permission.php
```diff
@@ -77,13 +77,4 @@
...
     'cache_expiration_time' => 60 * 24,
-
-    /*
-     * By default we'll make an entry in the application log when the permissions
-     * could not be loaded. Normally this only occurs while installing the packages.
-     *
-     * If for some reason you want to disable that logging, set this value to false.
-     */
-
-    'log_registration_exception' => true,
```
drbyte: With this removed, can we also remove `shouldLogException()` from `PermissionRegistrar.php` ?

drbyte:
Apart from my minor questions about the tests, this looks good.

(Also works well on v1 too.)

src/PermissionRegistrar.php
AlexVanderbist: The `registerPermissions` method used to return `true` on success and `false` if an exception was thrown. Useful or not, chances are people are using this in their code and thus changing it would be breaking (which is what I wanted avoid).

tests/CacheTest.php
AlexVanderbist: Yep, you're right!

config/permission.php
AlexVanderbist: Yep, I've removed it now 👍

src/PermissionRegistrar.php
drbyte: I'll accept that. 
Maybe consider a change in v3 if suitable then.

AlexVanderbist: Absolutely :)

src/PermissionRegistrar.php
```diff
@@ -17,39 +17,27 @@ class PermissionRegistrar
...
-            });
-
-            return true;
-        } catch (Exception $exception) {
-            if ($this->shouldLogException()) {
-                $this->logger->alert(
-                    "Could not register permissions because {$exception->getMessage()}".PHP_EOL.
-                    $exception->getTraceAsString()
-                );
+        $this->gate->before(function (Authenticatable $user, string $ability) {
```
freekmurze: Really great idea to leverage the gate's `before` function here. ❤️

AlexVanderbist: All credit goes to @hackel!

maxxscho: I haven‘t it tested yet, but if you use the before method, you can‘t use it for example in the AuthServiceProvider like demonstrated here #384 for a super user. The before method in this example doesn‘t get called.

hackel: @maxxscho The before method isn't a single-use hook, it adds the closure to a list of closures to run before the gate is checked.  If you haven't tested it yet, you should.  I can't think of any reason  other before calls shouldn't also work.

maxxscho: @hackel I've tested it yesterday. I had a `Gate::before()` method in AuthServiceProvider to give all Permissions to a "SuperUser", similar to the example on #384 . The `before` method didn't get invoked.
Setup of Laravel Permission was with Standard-User class of Laravel and anything Auth-related was as described as in the Readme. I was looking for a mistake on my App-Code, but everything looks well und works perfectly, except the `Gate::before()`-method.

stevenyangecho: This seems indeed has issue.via  this laravel gate `before()` callback invoke code
https://github.com/laravel/framework/blob/ac7eee2a13beb53b4806d1fb830f49dbf40df30c/src/Illuminate/Auth/Access/Gate.php#L290
it means although  the  added closure is a list of closures.but onces ths closure return not null value ,just like return true or false,it will immediately return the result and ignore the other closures.

and I looked the order regist of `Gate::before()`.this  spatie/laravel-permission alwasy before my 
`App\Providers\AuthServiceProvider`

maxxscho: @stevenyangecho thanks for the hint. I took a look at the Gate class, but it seems not close enough 😄

PR #506, opened 2017-10-12
Fix __construct middleware block in readme

andreamswick:
The construct block in the middleware section is missing the opening and closing `()`


drbyte:
Thanks!

PR #510, opened 2017-10-16
Remove the unused import

aiiro:
It looks like `Spatie\Permission\Models\Role` is not used in this class.
Instead, RoleContract is used in handle method to create a Role instance.


freekmurze:
Thanks!

PR #514, opened 2017-10-17
Using Docker for testing

aiiro:
Hi

This is a proposal PR about unit testing.
Now TravisCI runs multiple PHP versions tests when the PR is created.

I think it's better to run phpunit with those versions before Creating the new PR.
However, it's hard to install plural PHP versions without polluting the development environment.

I think Docker makes easy to run phpunit with multiple versions.

People can choose to run tests with Docker or not.
```
# without Docker
comoser test

# with Docker
composer test-versions
```

Offcourse if using Docker is not appropriate for this project, I will close this PR.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -678,6 +678,13 @@ The package doesn't come with any screens out of the box, you should build that 
 composer test
 ```
 
+### Testing with Docker
+
+https://www.docker.com/
+``` bash
+composer test-versions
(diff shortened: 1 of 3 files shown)


freekmurze:
Thank you for your proposal, but we're going to keep things the way they are now.

PR #517, opened 2017-10-18
Add dynamic redirect on middleware, and config for it.

lloricode:
So 403 will replace with redirect, dynamically with config.


config/permission.php
```diff
@@ -77,4 +77,25 @@
...
+
+
+    /*
+     * By default we'll make an entry in the application log when the permissions
+     * could not be loaded. Normally this only occurs while installing the packages.
+     *
+     * If for some reason you want to disable that logging, set this value to false.
+     */
+
+    'log_registration_exception' => true,
```
drbyte: I think you just merged in something that was removed in the release published earlier today ... :(

lloricode: hmmm, so we need to remove `log_registration_exception`?

lloricode:
@drbyte, is this ok now?

drbyte:
Looks ready for review. Pinging @AlexVanderbist or @freekmurze to consider whether they want this.

config/permission.php
```diff
@@ -77,4 +77,13 @@
...
     'cache_expiration_time' => 60 * 24,
+
+    /*
+     * By default the RoleMiddleware and PermissionMiddleware will trigger an abort with
+     * http 403 status code if the user doesn't have the specified permission or role.
+     *
+     * To override the default 403 response, provide a named route here.
+     */
+
+    'unauthorized_route_name_redirect' => null,
```
AlexVanderbist: Can we make this into a config value for an url instead of a route name? I'd suggest `redirect_unauthorized_users_to_url`.

src/Middlewares/PermissionMiddleware.php
```diff
@@ -9,7 +9,13 @@ class PermissionMiddleware
...
     public function handle($request, Closure $next, $permission)
     {
+        $routeNameRedirect = config('permission.unauthorized_route_name_redirect');
+
         if (Auth::guest()) {
+            if (! is_null($routeNameRedirect)) {
+                return redirect()
+                    ->route($routeNameRedirect);
+            }
             abort(403);
```
AlexVanderbist: These 6 lines of code are being used in 4 different places. Can we put them somewhere in their own method? `handleUnauthorized()` or something similar. Maybe use a trait that can be used in both `PermissionMiddleware` and `RoleMiddleware`?

src/Middlewares/PermissionMiddleware.php
```diff
@@ -9,7 +9,13 @@ class PermissionMiddleware
 {
     public function handle($request, Closure $next, $permission)
     {
+        $routeNameRedirect = config('permission.unauthorized_route_name_redirect');
```
AlexVanderbist: Don't forget to update this to a relevant variable name for the updated config key `$unauthorizedRedirectUrl` or something.

src/Middlewares/PermissionMiddleware.php
lloricode: @AlexVanderbist , yes much better that, i will push another commit later :)

config/permission.php
lloricode: @AlexVanderbist, we can, but it is easy to retrive a `url` with a `route name` because it is already decleared on `route/web.php`, to make easy manage url's.

what you think?. but it is ok to set as what you want. :+1:

src/Middlewares/PermissionMiddleware.php
lloricode: @AlexVanderbist, yes i will :)

config/permission.php
AlexVanderbist: Not everyone is using named routes. Defining this as an url gives us a bit more freedom as well, e.g. passing route params(not sure why you would need this but still)

src/Traits/HandleUnauthorized.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace Spatie\Permission\Traits;
+
+trait HandleUnauthorized
+{
+    public function responseUnauthorized()
```
lloricode: @AlexVanderbist, is the function name is good?

AlexVanderbist: Good for me, alternative would be `handleUnauthorized()` as it's still part of the middleware `handle` method but up to you!

lloricode: much better your idea @AlexVanderbist :+1:

AlexVanderbist:
Looks good to me!

lloricode:
@AlexVanderbist, great! :)

lloricode:
@AlexVanderbist, maybe i need to add unit test for this.. ive push this because unit test is passed, ive nitice also that calling helper `rediect` inside `trait` has a problem..

sorry i just push.. because i need to sleep from work.. 😑 

i will try to fix this 😊

lloricode:
i try make a custome class exception
it works fine also setting up url on added config index. bit in unit test
it gives error

freekmurze:
Replacing aborting by a configurable redirect is a nice improvement, but I think we can do even better.

If we replace the abort by throwing a custom exception than users can just add their wanted behaviour in Laravel's exception handler.

lloricode:
yes, i think it will resolve all of this.
so maybe i can remove the new index on config.. then remove also a redirec

just replace all with custome exception?
so user can handle it all even adding a redirect inside laravel exception
handler?, if exception is equal of new spatie permission exception.

is this ok? if it is i will push later 😊

freekmurze:
Let's add a config key called `throw_on_unauthorized` (or something with a better name). If it is `null` (and that should be the default value) then the package will just abort like it's doing now. This way we ensure backwards compatibility for the current version of the package. If the config key is set to an exception, then we throw that exception instead of aborting.

In the next major version we could remove that config key and just always throw an exception.

Unless @drbyte or @AlexVanderbist have some remarks/thoughts about this, we'd accept a fully tested PR that adds this. Make sure you update the readme as well with an example on how to use the new functionality.

AlexVanderbist:
What's stopping us from setting `throw_on_unauthorized` to `UnauthorizedHttpException` (which extends from `HttpException`) by default already? All `abort` does is raise a `HttpException` anyways so it would still be backwards compatible.

drbyte:
We wouldn't even need a config switch if we just change the `abort(403)` to throw an exception that is a child of same thing `abort` does, right? 

So, the overall code changes would be:
- create exception which extends appropriately
- change `abort` to throw exception
- add/update tests

(No need for trait or config key)

AlexVanderbist:
Not introducing any new config keys sounds great, definitely the best solution so far, @drbyte! :D

lloricode:
@AlexVanderbist  great, so can we close this now?
this is resolve with https://github.com/spatie/laravel-permission/pull/526#pullrequestreview-70871106

AlexVanderbist:
Sounds good, thanks for the initial PR! :)

lloricode:
:) welcome

PR #529, opened 2017-10-21
correct to correct config value for auth.defaults.guard

ziming:
correct to correct config value for auth.defaults.guard

As can be seen here

https://github.com/laravel/laravel/blob/master/config/auth.php


drbyte:
Thanks!

PR #530, opened 2017-10-21
Adding a clear cache command to clear permissions cache from command line .

rizerzero:
This is useful when adding the or removing permissions from the database and not using the package methods.

### DIFF
--- a/src/Commands/ClearCache.php
+++ b/src/Commands/ClearCache.php
@@ -11,7 +11,6 @@ class ClearCache extends Command
 
     protected $signature = 'permission:clear-cache';
 
-
     protected $description = 'Clear permissions cache';
 
     public function handle()
--- a/src/Commands/ClearCache.php
+++ b/src/Commands/ClearCache.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Spatie\Permission\Commands;
+
+use Illuminate\Console\Command;
+use Spatie\Permission\Contracts\Permission as PermissionContract;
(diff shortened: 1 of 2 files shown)


src/Commands/ClearCache.php
```diff
@@ -0,0 +1,23 @@
...
+
+    protected $signature = 'permission:clear-cache';
+
+
+    protected $description = 'Clear permissions cache';
+
+    public function handle()
+    {
+        $this->forgetCachedPermissions();
+        $this->info("Permissions chache was cleard");
```
drbyte: > Permissions `chache` was `cleard`

A couple typos here ;)

AlexVanderbist:
Thanks for the PR! Is there any reason you can't `cache:clear` or even more specifically `cache:forget spatie.permission.cache`?

src/Commands/ClearCache.php
rizerzero: thanks for the corrections

rizerzero:
cache:clear clears the global cache of the entire app, but the main reason for adding this command is that it allows us to know that permissions are cached by the package without digging in the code. 
i was creating some rules for testing and making some changes directly on the database but i did not see the effect so i had to search and find out about the caching by myself this  command would have saved me some time ;)

AlexVanderbist:
This is a good commit but gonna close this for now. We've already got the `php artisan cache:forget spatie.permission.cache` command documented in the readme and I don't think changing that to a custom command (which internally does the same thing) would make much difference.

PR #538, opened 2017-10-25
Gate::before should return null if fail.

ziming:
If you return false at Gate::before, you reject all other user-defined gates. Better to return null so the user-defined gates can take over. If no gates take over then is a fail too so more flexible.

As seen in the comments on this link where a user (Joseph Sibler of bouncer package) pointed out Jeffery mistake to return false in Gate::before

https://laracasts.com/series/whats-new-in-laravel-5-1/episodes/15

If I'm wrong about this let me know too. Cheers!

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -31,7 +31,7 @@ class PermissionRegistrar
         $this->gate->before(function (Authenticatable $user, string $ability) {
             try {
                 if (method_exists($user, 'hasPermissionTo')) {
-                    return $user->hasPermissionTo($ability);
+                    return $user->hasPermissionTo($ability) ?: null;
                 }
             } catch (PermissionDoesNotExist $e) {
             }
--- a/tests/GateTest.php
+++ b/tests/GateTest.php
@@ -2,12 +2,26 @@
 
 namespace Spatie\Permission\Test;
 
+use Illuminate\Contracts\Auth\Access\Gate;
+
 class GateTest extends TestCase
 {
     /** @test */
     public function it_can_determine_if_a_user_does_not_have_a_permission()
     {
         $this->assertFalse($this->testUser->can('edit-articles'));
+    }
+
+    /** @test */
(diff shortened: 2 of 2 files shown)


drbyte:
Agreed. 

Do you mind writing a couple unit tests for this?

ziming:
I wanted to, but got an emergency company site issue I have to handle now. =/. It would be great if someone can take over.

AlexVanderbist:
On it :)

AlexVanderbist:
@drbyte any other tests you'd like to see for this?

drbyte:
That looks good. For v3 we might consider adding a test for a gate on a specific model; this would serve as an example too.

drbyte:
Please go ahead with merge and tag.

ziming:
Hello!

Although I added the fix part of this commit, I still lack understandings on Gate::before

From what I understand, you can have multiple Gate::before() calls in AuthServiceProvider, they just get called 1 after another. At the end, after all the Gate::before() calls assuming none of them return true or false. It goes to the Policy and other user defined normal gates.

In this case if I define my own Gate::before() in AuthServiceProvider, which get called first? this package Gate::before comes first or my Gate::before() calls?

Secondly, I wanted to do something like

```
class UserPolicy {

    public function view(User $user, User $model) {

        // say compare if the domain of their email are the same
        if ($user->can('view users') && $user->email_domain() === $model->email_domain()) {
            return true;
        }

        return false;
    }
}
```

Then in my controller

```
$this->authorize('view', $user);
```

My question is, will it get authorized for a user that has 'view users' permission but not having a matching email_domain as the user he is about to edit? I'm not sure how to describe it but I suddenly got confused thinking `$user->can('view users')` will call it's own Gate::before() which call $user->hasPermissionTo('view users') which get authorized and ... the user get authorized to view the user despite being having emails of different domains (jack@yahoo.com and john@hotmail.com)

Would be great if anyone can enlighten me on this.

PR #545, opened 2017-10-31
Ignore errors when assigning role/permission previously given to user

arthursorriso:
If you are trying to assign a role or a permission previously added to a user, the system will show an query error.
Doing this, before add, I do a check and add if the user have not the role or permission.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -18,14 +18,14 @@ trait HasPermissions
      */
     public function givePermissionTo(...$permissions)
     {
-        $permissions = collect($permissions)
+        collect($permissions)
             ->flatten()
             ->map(function ($permission) {
                 return $this->getStoredPermission($permission);
             })
             ->each(function ($permission) {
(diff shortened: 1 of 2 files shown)


src/Traits/HasPermissions.php
```diff
@@ -18,18 +18,19 @@
...
     {
-        $permissions = collect($permissions)
+        collect($permissions)
             ->flatten()
             ->map(function ($permission) {
                 return $this->getStoredPermission($permission);
             })
             ->each(function ($permission) {
                 $this->ensureModelSharesGuard($permission);
+                if (! $this->hasPermissionTo($permission->name)) {
```
AlexVanderbist: Maybe we can use a `filter` method on the collection here to filter out permissions that are already given to the user? Same for the `assignRole` method.

AlexVanderbist:
Gonna close this one for now. I agree that this might be confusing and it's worth revisiting soon when we introduce some more breaking changes.

PR #546, opened 2017-10-31
Command to check for role middleware

xtrasmal:
A command that checks for the existence of 'role' middleware on routes. Handy for checking forgotten routes.


xtrasmal:
@freekmurze Hoi Freek, how have you been? 
I needed to check for routes that did not have 'role' or 'permission' middleware yet. All our routes need the proper permissions or roles and so I've created these commands to give me a list of those that do not have role/permission middleware.
I thought I'd share, cause the might be folks out there that have the same use-case.

freekmurze:
Thank you for this, but in my mind this functionality is a little bit too esoteric for us to maintain.

Issue #548, opened 2017-11-01
syncRoles should accept id's too.


arakash92:
There's a package (kristijanhusak/laravel-form-builder) I'm using which integrates with laravels models to provide dropdown/select fields for relationships.

The problem is that it uses the 'id' as the keys. In other words, I'm trying to do `$user->syncRoles([]3, 4, 7])` and it doesn't work that way obviously.

I think it ought to since it's common for data to be in that way.

AlexVanderbist:
I totally agree, let's continue this issue in your PR.

PR #549, opened 2017-11-01
assignRole and syncRoles using ids as well as names.

arakash92:
I implemented #548. Namely the ability to for `HasRole::syncRoles` and `HasRole::assignRole` to accept role ids as well as names.

I needed this for compatibility with a form framework (kristijanhusak/laravel-form-builder).

This is my first time contributing to a project so I hope I've done it correctly.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -383,7 +383,7 @@ trait HasRoles
     {
         if (is_numeric($role)) {
             return app(Role::class)->findById($role, $this->getDefaultGuardName());
-        }else if (is_string($role)) {
+        } elseif (is_string($role)) {
             return app(Role::class)->findByName($role, $this->getDefaultGuardName());
         }
 
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -381,7 +381,9 @@ trait HasRoles
 
     protected function getStoredRole($role): Role
     {
-        if (is_string($role)) {
+        if (is_numeric($role)) {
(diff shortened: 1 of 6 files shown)


AlexVanderbist:
This looks really good! Thanks for fixing the random spelling errors as well ;)

src/Traits/HasRoles.php
```diff
@@ -381,7 +381,9 @@ public function getRoleNames(): Collection
 
     protected function getStoredRole($role): Role
     {
-        if (is_string($role)) {
+        if (is_numeric($role)) {
+            return app(Role::class)->findById($role, $this->getDefaultGuardName());
+        } elseif (is_string($role)) {
```
AlexVanderbist: This can be changed to a single `if` statement

AlexVanderbist: This can be changed to a single `if` statement as the previous statement will return if true.

arakash92: Like this, perhaps?

    return is_numeric($role)
            ? app(Role::class)->findById($role, $this->getDefaultGuardName())
            : app(Role::class)->findByName($role, $this->getDefaultGuardName());

drbyte: You could do that, but I think @AlexVanderbist meant to just remove the `elseif` clause, like this:

```php
if (is_numeric($role)) {
   return foo;
}
return bar;
```

AlexVanderbist:
Merging this in.

Thanks again for the PR!

tomlankhorst:
Why only add numeric syncs for roles, not permissions?

PR #556, opened 2017-11-09
Add support for Multi-Tenant

ChadLRoss:
In this PR I have added support for an optional multi-tenant.  Whenever possible I overrode existing functionality.  There is not support in this version for guards,  nor is there support for adding permissions directly to a user as it relates to a tenant.  A user is granted permissions to a role / tenant and then that role inherits the permissions.  Permissions are checked through the use of a inheritance.  

Ideally you would accept this as a new branch and we could continue to develop out the missing functionality.


ChadLRoss:
I see some of the issues are failing with styleci most of these are not my issues.  I am happy to fix off all of them in a different PR if you'd like.

AlexVanderbist:
First of all, thanks for all your work! This looks really interesting. I'd love to see something like this in the actual package but also keeping it simple enough if you don't know about multi-tenancy.

Is there any way we can implement this on the "main" branch whilst keeping it compatible with people who don't have multi-tenancy? For v3 we're planning a lot of breaking changes so that wouldn't be a problem.

ChadLRoss:
Hey Alex you could merge this into the main branch everything is backwards compatible and you don't need to know about the tenant stuff at all.  A purposely create a new HasTenant class, a user could easily  continue to use HasRoles and never see or use any of the Tenant information.  I also create the migration separately to abstract it one step further.

Furthermore a user could run the migration, use HasTenants and still use the package as it is today.  I did not override any of the previous functionally, in fact in the application I am using it for I uses both the tenant and the non-tenant calls throughout.

ChadLRoss:
I am going to decline this PR and submit another, there were some PSR-2 failures and a test was failing after I squashed.  Let me fix and re-submit.

PR #557, opened 2017-11-10
Add support for Multi Tenant

ChadLRoss:
A fully backwards compatible addition to for tenant functionality.

Added a new migration stub for tenant db.  It was more of a boiler plate than anything as I assume most people wanting to have multi-tenant functionality already have a tenant table.  The package includes support for both incrementing primary ids and well as support for char primary keys on the tenant table (tests for both were written)

A multi-tenant functionality is completely optional and should the user not need that part they do not need to use it (much like how the guard stuff is optional)

There are some ideas areas that are not yet supported from the mutli-tenant perspective that are in the standard implementation such as the use of guards is not supported in the multi-tenant relationship.  The use of assigning direct permissions to the user with a tenant is also not support when using multi-tenant.  You must assign a role permissions and then you must assign the role to the user with the scope of the tenant.  Let me know if you have any questions, happy to jump on Skype and discuss as well.


ChadLRoss:
I am not sure why builds are failing with --prefer-lowest it I shows that you're requiring illuminate/database as 5.4 and it appear that setKeyType was available in that version.

https://github.com/illuminate/database/blob/5.4/Eloquent/Model.php#L1168

Thoughts?

AlexVanderbist:
This looks really promising! Will read through the code and try it out soon.


>Added a new migration stub for tenant db. It was more of a boiler plate than anything as I assume most people wanting to have multi-tenant functionality already have a tenant table.

If at all possible I'd like to keep boilerplate like that out of the package. Is there any way we can put the necessary methods for the `Tenant` interface in a trait? This way we can move the `Tenant` model and migration to the `tests` directory and use them solely for testing.


>A multi-tenant functionality is completely optional and should the user not need that part they do not need to use it 

👌👌👌


>the use of guards is not supported in the multi-tenant relationship

Using the guard names explicitly will be dropped in v3 anyways so don't worry about that too much.

Thanks a lot for the PR!

AlexVanderbist:
For the failing tests: looks like `setKeyType` was only added in 5.4.17: https://github.com/illuminate/database/commit/3e2215e21c06bb8eff7d5eb5871ab8178a787239. Feel free to set a minimum requirement for that version in composer.json.

ChadLRoss:
Great feedback.  moving it to a trait makes perfect sense.  Thanks for checking into the failing tests for me, I'll get that updated as well.  Should have time next week to knock those out.  Thanks for the response

drbyte:
> A multi-tenant functionality is completely optional and should the user not need that part they do not need to use it 

Like @AlexVanderbist I'm happy that it's not forced on people.

However, your PR as it stands now changes all the documentation from instructing to use `HasRole` trait to `HasTenant` ... which basically makes `HasTenant` looking like it's the primary use case for the package. While that might be convenient for a couple people, it may create confusion for the masses. 


Additionally, it's important to note that this implementation of multi-tenant is only one of many possible ways it can be done. Wondering how that can be described for those needing support in that direction.

sagearslan:
This is perfect, I've been sweating creating a custom implementation. Any ETA for merging into master?

freekmurze:
We'd like to keep this package as simple as it can be. For our projects we don't have any need for multitenancy and don't want to maintain the code in this PR.

You're of course free put your trait and extra coe in a repo of your own, we'd be happy to put a link to it in our readme.

PR #559, opened 2017-11-12
Update command signature to "make:role and make:permission"

yamenarahman:
I find it more compatible to Laravel to use these command signatures rather than the hyphen between old signatures like `create-permission` and `create-role`.

Hope this make sense.
<3 @freekmurze


freekmurze:
Not going to pull this in now because it's a breaking change. 

Seems like a good change for the next major version.

yamenarahman:
Okay, Glad to be a part of the next release.

PR #573, opened 2017-11-27
Give and revoke multiple permissions

### LINKED ISSUE #572: Return type does not allow Collection, and 'guard_name' condition uses "whereId" instead of "whereIn"
https://github.com/spatie/laravel-permission/blob/03eb748141432b5d94eb1613b99f818aac1f5070/src/Traits/HasPermissions.php#L73

HasPermissions::getStoredPermission supports multiple permissions as an input, but is tied to a single permission for the return type.

Also, if $permissions is an array, there's a call to "whereId" instead of "whereIn", which I'm pretty sure is generating an invalid query:

Original query:
```sql
select * from `permissions` where `name` in ("test-permission") and `id` = "guard_name";`
```

Updated query:
```sql
select * from `permissions` where `name` in ("test-permission") and `guard_name` in ("web");
```

I've prepared [a fix](https://github.com/artissant/laravel-permission/tree/feature/give-and-revoke-multi-permissions) but thought I'd create an issue before making a pull request. This _shouldn't_ introduce any breaking changes because passing an array/Collection to revokePermissionTo would currently trigger an error.

artissant:
Fixes #572

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -52,7 +52,7 @@ trait HasPermissions
     /**
      * Revoke the given permission.
      *
-     * @param \Spatie\Permission\Contracts\Permission|string $permission
+     * @param Permission|Permission[]|string|string[] $permission
      *
      * @return $this
      */
@@ -68,9 +68,9 @@ trait HasPermissions
     /**
      * @param string|array|\Spatie\Permission\Contracts\Permission|\Illuminate\Support\Collection $permissions
      *
-     * @return \Spatie\Permission\Contracts\Permission
+     * @return Permission|Permission[]|\Illuminate\Support\Collection
      */
-    protected function getStoredPermission($permissions): Permission
+    protected function getStoredPermission($permissions)
     {
         if (is_string($permissions)) {
             return app(Permission::class)->findByName($permissions, $this->getDefaultGuardName());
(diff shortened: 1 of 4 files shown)


src/Traits/HasPermissions.php
```diff
@@ -52,7 +52,7 @@ public function syncPermissions(...$permissions)
     /**
      * Revoke the given permission.
      *
-     * @param \Spatie\Permission\Contracts\Permission|string $permission
+     * @param Permission|Permission[]|string|string[] $permission
```
AlexVanderbist: Can we use the FQCN here?

tests/PermissionTest.php
```diff
@@ -41,4 +41,16 @@ public function it_has_user_models_of_the_right_class()
         $this->assertTrue($this->testUserPermission->users->first()->is($this->testUser));
         $this->assertInstanceOf(User::class, $this->testUserPermission->users->first());
     }
+
+    /** @test */
+    public function it_can_give_and_revoke_multiple_permissions()
```
AlexVanderbist: Can we move this test to the `HasPermissionsTest`? I think that'd be a more appropriate context

src/Traits/HasPermissions.php
```diff
@@ -68,9 +68,9 @@ public function revokePermissionTo($permission)
     /**
      * @param string|array|\Spatie\Permission\Contracts\Permission|\Illuminate\Support\Collection $permissions
      *
-     * @return \Spatie\Permission\Contracts\Permission
+     * @return Permission|Permission[]|\Illuminate\Support\Collection
```
AlexVanderbist: FQCN as well :)

artissant:
Hi, nice one for looking over it, I've pushed those tweaks up now.

drbyte:
Thanks!

PR #577, opened 2017-12-01
Dynamic permission_id and role_id columns according to tables name

mohamednagy:
permission_id and role_id names are hard coded, its a good idea to dynamically generate the names according to tables names in the config file,


mohamednagy:
ok, done :)

drbyte:
Thanks!

PR #583, opened 2017-12-06
Add findOrCreate function to Permission model

skoTner:
When trying to find a permission that does not exist, it could be added to the database so that it easily can be used in the application going forward. It could be further built by adding a boolean variable in the config-file wheter the developer wants it to trigger automatically everytime the @can/user->can() function is used.

Eg. for extending the User model:
```php
class User extends Authenticatable {

  public function hasAccess($permission, $guard = null) {
    if(!$this->can($permission)) {
      Permission:findOrCreate($permission, $guard);
      abort(403);
    }

    return true;
  }

}
```

### DIFF
--- a/tests/RoleTest.php
+++ b/tests/RoleTest.php
@@ -181,6 +181,14 @@ class RoleTest extends TestCase
     }
 
     /** @test */
+    public function it_creates_permission_object_if_it_does_not_have_a_permission_object()
+    {
+        $permission = app(Permission::class)->findOrCreate('other-permission');
+
+        $this->assertFalse($this->testUserRole->hasPermissionTo($permission));
+    }
+
+    /** @test */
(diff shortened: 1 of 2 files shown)


tests/RoleTest.php
```diff
@@ -180,6 +180,14 @@ public function it_returns_false_if_it_does_not_have_a_permission_object()
         $this->assertFalse($this->testUserRole->hasPermissionTo($permission));
     }
 
+    /** @test */
+    public function it_creates_permission_object_if_it_does_not_have_a_permission_object()
+    {
+        $permission = app(Permission::class)->findOrCreate('other-permission');
+
+        $this->assertFalse($this->testUserRole->hasPermissionTo($permission));
```
drbyte: I'm not sure this is the right assertion here. Wouldn't it be better to check that the `other-permission` now exists, or that assigning `other-permission` to the user is successful?

drbyte:
Thanks!

PR #587, opened 2017-12-11
Fixed getting default guard name

d-lamers:
It did not return the default guard name but the first available option

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -117,8 +117,9 @@ trait HasPermissions
     protected function getDefaultGuardName(): string
     {
         $default = config('auth.defaults.guard');
+        $guardNames = $this->getGuardNames();
 
-        return $default ?: $this->getGuardNames()->first();
+        return $guardNames->contains($default) ? $default : $this->getGuardNames()->first();
     }
 
     /**
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -119,7 +119,7 @@ trait HasPermissions
         $default = config('auth.defaults.guard');
         $guardNames = $this->getGuardNames();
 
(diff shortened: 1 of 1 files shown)


src/Traits/HasPermissions.php
```diff
@@ -117,8 +117,9 @@ protected function getGuardNames(): Collection
     protected function getDefaultGuardName(): string
     {
         $default = config('auth.defaults.guard');
+        $guardNames = $this->getGuardNames();
 
-        return $default ?: $this->getGuardNames()->first();
+        return $guardNames->contains($default) ? $default : $this->getGuardNames()->first();
```
drbyte: Would this be more performant to call `$guardNames->first()` instead of rerunning `$this->getGuardNames()` a second time?

(I'm still not sure that I agree with the overall proposed logic change, but if it is to go ahead, it feels like it should be optimized.)

d-lamers: You are right, it's an oversight. I'll change it in this PR

drbyte:
The more I look at this, the more I agree with it. 

But is it a breaking change for existing sites? 

OLD BEHAVIOR:
- if model has a $guard_name property, return it
- else if model has any guards in config, use the first one
- else return default guard

NEW BEHAVIOR:
- if the default guard is one of:
 -- $guard_name property on the model
 -- one of the guards in config
then return it
- else return the $guard_name property, if set, or the first configured guard in config

drbyte:
Note: #622 has been merged, which changes the signature of where the changes from this PR are located. So merge conflicts will need to be fixed.
I'm still considering the BC impact of this proposed change ... not because the logic is wrong, but because projects may be coding for the accidental behavior of this, and changing/correcting it could be a breaking change.

drbyte:
It seems the changes in #622 mitigate the circumstances that gave rise to the proposal in this PR. 
Closing this one. 

Thanks for submitting!

PR #598, opened 2017-12-23
add flag model field

### LINKED ISSUE #594: It is suggested that the flag field be added, and name is description
```
Support multilingual languages for example：
Schema::create($tableNames['permissions'], function (Blueprint $table) {
            $table->increments('id');
            $table->string('name')->comment('编辑');
            $table->string('flag')->comment('edit articles');
            $table->string('guard_name');
});
Schema::create($tableNames['roles'], function (Blueprint $table) {
            $table->increments('id');
            $table->string('name');
            $table->string('flag');
            $table->string('guard_name');
            $table->timestamps();
        });
```

hinet:
Closes #594


AlexVanderbist:
Thanks for the PR! Added a reply on #594 why we're not pulling this one in.

PR #620, opened 2018-01-05
Use double quotes in Artisan commands

Epistol:
Because it'll think that single quote + two words = two params;

Btw, there is command to create role and permissions, why is there not to associate them ?


drbyte:
Thanks!

PR #622, opened 2018-01-11
Support  getting guard_name from extended model when using static methods

Te7a-Houdini:
this **PR** , doesn't alter the current behavior , it will allow static methods to access the **$guard_name**  for example:- 

```php
class ApiRole extends Spatie\Permission\Models\Role
{
    protected $guard_name = 'api';
}

/*
* so if we make this now we will get the role of guard  api , 
* while previous was getting the default auth.php guard and we had to make this ,
* ApiRole::findByName('admin','api');
*/

 ApiRole::findByName('admin'); 
```

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -1,0 +1,47 @@
+<?php
+
+namespace Spatie\Permission;
+
+use Illuminate\Support\Collection;
+
+class Guard
+{
+    /**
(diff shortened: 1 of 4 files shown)


drbyte:
@freekmurze While #587 attempts to address the same issue, I think this is more thorough.
Are you okay with this architectural change?

drbyte:
Thanks!

Issue #593, opened 2017-12-15
Why send through permission array and discard it


Jigsaw5279:
When an UnauthorizedException is raised, the array of requested permissions or roles is sent trough.

However, in the Exception class itself that parameter is discarded completely.
Is there any specific reason for that?

    public static function forPermissions(array $permissions): self
    {
        return new static(403, 'User does not have the right permissions.', null, []);
    }

drbyte:
I think that's a question for @AlexVanderbist 

Alex, was this intentional?

drbyte:
Ref #526 and #517

flashadvocate:
Seems like it'd be useful for debugging if it was included in the output

drbyte:
A PR is welcome.

AlexVanderbist:
>Seems like it'd be useful for debugging if it was included in the output

That was the case. The exception message used to be something along the lines of "User does not have the right permissions. Necessary permissions are ..." but I left that out because it felt like it was exposing too much information.

Wouldn't mind having that in there again tho :)

benyanke:
That seems like a feature that would be useful `if(config('app.debug'))`. Or perhaps as it's own config.

benyanke:
PR written above.

Jigsaw5279:
Resolved in #629

PR #628, opened 2018-01-22
Optionally display required permissions in exception message

benyanke:
Resolves issue discussed in #593

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -77,4 +77,13 @@ return [
      */
 
     'cache_expiration_time' => 60 * 24,
+
+    /*
+     * When set to true, the name of the required permission/role is added to the
+     * exception message. This could be considered an information leak in some contexts,
+     * so enable if you know what you're doing.
+     */
+
+     'display_permission_in_exception' => false,
+
 ];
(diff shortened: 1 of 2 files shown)


benyanke:
@AlexVanderbist

drbyte:
> Thanks for the patience - this is my first non-docs PR iirc.

Woot!  Good on you for stretching yourself!

Jigsaw5279:
Wouldn't it make sense to pass down the array instead of converting it to a string? Or at least store them within the exception, so you can decide what to do with them yourself (instead of having to parse them from the string again..)

Jigsaw5279:
Added PR #629 which does exactly what I described

benyanke:
Ok, I fixed the issues you mentioned, but styleCI still failed. Where can I see what styleCI is not liking? I just see PR info, no errors there.

Jigsaw5279:
You need to have a space between the if keyword and the opening bracket

benyanke:
Figured it out - found the diff option which displays the issue. I think this PR is ready to go @drbyte. 
StyleCI is pretty awesome, I might have to implement that at work for my dev...

drbyte:
Thanks @benyanke !

PR #629, opened 2018-01-22
Add reason for failed role/permission as exception property

Jigsaw5279:
Resolves issue #593 by also allowing to access the permssions/roles required (as I originally intended to do for logging purposes)

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -77,4 +77,11 @@ return [
      */
 
     'cache_expiration_time' => 60 * 24,
+
+    /*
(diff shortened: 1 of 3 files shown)


drbyte:
Can you please add tests to verify that the passed array can be accessed when the exception is thrown.

After the PR #628 is merged then we'll review your copy of their contribution along with your changes for merging.

drbyte:
Thanks @Jigsaw5279 !

Jigsaw5279:
Sure, solving my own problems here :P

PR #637, opened 2018-02-02
Allow a collection containing a model to be passed to HasRoles scopes

mattdfloyd:
When passing a `Role` model as part of a collection to `scopeRole` or a `Permission` to `scopePermission`, the model would be cast to an array. This is because `toArray` converts all of the collection's nested objects to an array. `all` retains the models within the underlying array instead.


mattdfloyd:
Use case:
```
$roles = Role::whereIn('name', ['super', 'admin'])->get();
User::role($roles)->get();
```

This is an improvement over the following, which make a database call for each role that is passed.
```
User::role(['super', 'admin'])->get();
```

tests/HasRolesTest.php
```diff
@@ -429,6 +445,23 @@ public function it_can_list_all_the_coupled_permissions_both_directly_and_via_ro
         );
     }
 
+    /** @test */
+    public function it_can_scope_users_with_permissions()
```
drbyte: Do you mind moving this test to `HasPermissionsTest`?
Perhaps near https://github.com/spatie/laravel-permission/blob/master/tests/HasPermissionsTest.php#L105

PR #644, opened 2018-02-05
Move permissions methods out of HasRoles trait

mattdfloyd:
Diffs are a bit funky here so let me know if I should break this up into many commits or PRs to make it clean. Only added an assertion to the `it_can_determine_that_user_has_direct_permission`. Otherwise just copy and paste.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -54,24 +54,24 @@ trait HasPermissions
         }, []));
 
         return $query->
-            where(function ($query) use ($permissions, $rolesWithPermissions) {
-                $query->whereHas('permissions', function ($query) use ($permissions) {
-                    $query->where(function ($query) use ($permissions) {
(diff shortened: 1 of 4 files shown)


AlexVanderbist:
I'd really love to see this merged into v2. It'd be great to have this cleaned up before forking v3 as well.

Does this introduce any breaking changes? Feels like it shouldn't (even if a user would `use HasPermissions` directly) but I'm not 100% sure.

drbyte:
Thanks @mattdfloyd

drbyte:
@AlexVanderbist note: users still can't use `HasPermissions` trait directly (they couldn't previously either), especially since it has a dependency on the `HasRoles::hasRole()` method. I guess we could add an abstract method for `hasRole()`, but technically it depends on other things in the `HasRoles` trait, and around the circle we go ... 
So, really, the "usage signature" hasn't changed: one still must add just the `HasRoles` trait.

PR #645, opened 2018-02-05
Allow assign/sync/remove Roles from Permission model

fitztrev:
This will allow the Roles to be synced from a Permission.

My use case is on the CRUD pages for Permissions. When creating/editing a Permission, I'd like to be able to specify which role(s) it applies to.

Currently I am doing this:

```php
$permission->roles()->sync($request->roles);
app(PermissionRegistrar::class)->forgetCachedPermissions();
```

After this PR:

```php
$permission->syncRoles($request->roles);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,27 @@ $role = Role::create(['name' => 'writer']);
 $permission = Permission::create(['name' => 'edit articles']);
 ```
 
+A permission can be assigned to a role using 1 of these methods:
+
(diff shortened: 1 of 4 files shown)


drbyte:
Do you mind adding a test to cover this use-case?

fitztrev:
No problem. Tests added.

drbyte:
Thanks!

PR #646, opened 2018-02-06
Improve documentation on extending models

CasperBE:
Extending a model doesn't work with implementing the contract, but works with extending the model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,18 +717,23 @@ Two notes about Database Seeding:
 
 ## Extending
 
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

drbyte:
(Technically it does work by implementing the contract ... but extending often feels like the "only" approach because in many cases you will also want most of what's already in the model. Both approaches are correct, depending on the intentions or needs.)
Thanks for pointing out that it could be clearer.

PR #664, opened 2018-02-19
Add ability to use hasPermissionTo() / hasDirectPermissionTo() with Permission's ID (int)

pkeogan:
I created a fork to have the ability to use hasPermissionTo() with and int or a string.

If a string is given, it will do what it normally does,

If and Integer is given, it will compare that against all the permission's IDs. 

This also allows hasAnyPermission to accept ints in the array. Also did the same to hasDirectPermissionTo(). I have tested within my application, and haven't found any errors. I added the exception message Permissiondoesnotexist:findId -  so a message will display correctly when and ID is not found. 

please check my code below and tell me what you think!

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,3 @@
-# This fork adds permission hierchy (parenting and children)
-
 # Associate users with permissions and roles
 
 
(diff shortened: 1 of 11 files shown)


pkeogan:
Never honestly used StyleCI, so let me know what you want for that to pass.

drbyte:
Looks good @pkeogan - Thanks for submitting this!

drbyte:
How StyleCi works in a nutshell: The StyleCI review suggests code changes, either for formatting or optimal syntax. You can take their suggestions and apply them to your code and commit them to the PR, or you could go a step further and link StyleCI to your github account and let it do a PR against your repo so you can just merge their suggestions directly.

PR #666, opened 2018-02-20
Ignore default-guard-config in favor of the first in list

bussnet:
Hi Guys.

We have a project with some subdomains, every subdomain has there own guard.

We also use the same views in more than one subdomain, so we only check the permission/role without the guard_name. In a middleware we change the auth.defaults.guard depending on the subdomain which is requested.

Im sure this is a bug, cause why would someone prefer the first guard in an list instead of the explicit defined default guard?

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -46,6 +46,6 @@ class Guard
     {
         $default = config('auth.defaults.guard');
 
-        return static::getNames($class)->first() ?: $default;
+        return $default ?: static::getNames($class)->first();
     }
 }


drbyte:
The tests are failing because your code change causes the model-specific defined guards, and all other possible overrides, to be completely ignored if a config default is set (which usually is not blank, so will always be set).

bussnet:
> However, an equally fair question is: why would someone want a Default guard when a model-specific guard has been specified? Defaults are usually used for fallbacks when something more specific isn't offered.

Because this is the only option to globally set/change the guard. ;)
We have one user table (also one model) with 4 guards and change the guard in a middleware which detects which guard is required for the requested url. (The guard checks if the user is allowed to login in this url)

drbyte:
This change cannot be merged because it breaks the primary uses of guard rules.

Do you mind sharing the middleware you've written which requires the change you proposed?

drbyte:
I wonder if adding something like this to the `Guard` class might accommodate the need? 

```diff
    public static function getDefaultName($class): string
    {
        $default = config('auth.defaults.guard');
+
 +       $override = config('auth.defaults.guard_override', null);

-        return static::getNames($class)->first() ?: $default;
+        return $override ?? static::getNames($class)->first() ?: $default;
    }
```

... then you could leverage the new `auth.defaults.guard_override` config setting.

Still, I think it'd be helpful to look at your middleware code in order to see some of the broader context of what you're doing.

bussnet:
@drbyte thanks for your help. I solved our specific problem with overwriting the getDefaultGuardName() in the model and returning the **auth.defaults.guard**.

I just want to help to solve a generic problem, cause it doesn't seem right to return the first item of a list instead of a selected default. But in case the different guards are for different models, it does make more sense.

drbyte:
Your point is understood. That's why I requested to see your code in order to study the bigger picture, in hopes of helping find a more generic solution.

bussnet:
OK, got you. The code is to complex to post it, but the structure is this:
We have 5 RouteGroups, every group has his own middleware and is mapped to a subdomain.
In the middleware we do a specific config of the system, also define the guard for this subdomain with `\Config::set('auth.defaults.guard', $this->guard);`.

I think the best solution (changing the default.guard config param is not good) were to change the guard globally: `Permission::setActiveGuard($guard)`

drbyte:
@bussnet wrote:
> I think the best solution were to change the guard globally: Permission::setActiveGuard($guard)

Yes, that feels sensible.
If its purpose is to always override ALL other default behavior, then I guess `setActiveGuard` is an okay name.
Perhaps add it to the `Guard` model: `Guard::setActiveGuard($guard)`.

drbyte:
I see where some of the confusion has come from: Laravel's [AuthManager specifically uses that config value for passing guard](https://github.com/laravel/framework/blob/5.6/src/Illuminate/Auth/AuthManager.php#L186-L212) state. Unfortunately while it's opinionated about that, it has no opinion or practice at the per-model level.

PR #693, opened 2018-03-21
Additional Language

lloricode:
in Exceptions 

Thanks :)


AlexVanderbist:
Hi, thanks for the PR! Gonna hold off merging this one for now. The exception messages in this package are mainly to be used during development. I wouldn't recommend returning them to your front users. Instead you can use validation rules to avoid most of these exceptions to be thrown.

If you really want to translate these exception messages in your application you can always surround your code with a `try-catch` and manually throw new exceptions or flash error messages using your translated strings.

Thanks again and good luck 👍

lloricode:
I see @AlexVanderbist , :+1:  Thanks

Issue #703, opened 2018-04-02
Permission throught role issue


MickaelViaud:
Hello,

I've seeded permissions through various roles and when I try to check the current user permissions, it doestn't work.

I've also try to clear the app cache without success. The issue seems to be similar to #702 

I already did this operation with older version of laravel-permission on L5.5 without any problem.

- user()->permissions : don't work
- user()->getAllPermissions() : work
- user()->can('PERMISSION_NAME') : don't work

Here is a screenshot of a result : 
![Code](https://i.imgur.com/Oo8TjZi.png)

MickaelViaud:
#702

drbyte:
> I already did this operation with older version of laravel-permission on L5.5 without any problem.

Which older version of laravel-permission? 
Is this new scenario also on L5.5 ?

MickaelViaud:
I've made the same things (I think !) with L5.5 and Spatie 2.8.1 and it's working.

Here's all the process : 
1 - Creating manager guard on config/auth.php
```php
return [

    'defaults' => [
        'guard' => 'web',
        'passwords' => 'users',
    ],

    'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],

        'manager' => [
            'driver' => 'session',
            'provider' => 'manager',
        ],

        'api' => [
            'driver' => 'token',
            'provider' => 'users',
        ],
    ],

    'providers' => [
        'users' => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],

        'manager' => [
            'driver' => 'eloquent',
            'model' => App\User::class,
        ],

        // 'users' => [
        //     'driver' => 'database',
        //     'table' => 'users',
        // ],
    ],

    'passwords' => [
        'users' => [
            'provider' => 'users',
            'table' => 'password_resets',
            'expire' => 60,
        ],
    ],

];
```

2- Seed user, role and permissions
```php
app()['cache']->forget('spatie.permission.cache');

// Create roles
$admin = \Spatie\Permission\Models\Role::create(['name'=>User::ROLE_ADMIN, 'guard_name'=>'manager']);
$writer = \Spatie\Permission\Models\Role::create(['name'=>User::ROLE_WRITER, 'guard_name'=>'manager']);
$moderator = \Spatie\Permission\Models\Role::create(['name'=>User::ROLE_MODERATOR, 'guard_name'=>'manager']);

// Creating permissions
[...]

Issue #702, opened 2018-03-29
@can('permission') see model_has_permission only?


sadalsuud:
HI, I have a role with a permissions, but I do @can('add_user') not work, and if I add add_user permission into model_has_permission, @can('add_user') work. 

My question is if @can('permision') see role->persmission or permision into model_has_permission. 

I am use Laravel 5.5

drbyte:
1. If you are manually inserting records into the tables, then you need to also manually clear the cache.
You should NOT be manually editing the database. Use the methods provided in the README.

2. The `$user->can()` and blade `@can()` function similar to the `$user->hasPermissionTo()`, but they do it using the Laravel Guard infrastructure so that permissions are registered very early in the bootstrap process. If you have other guards or middleware that are interfering then you may get unexpected results.

PR #704, opened 2018-04-03
Add ability to pass guard name to gate methods like can()

drbyte:
Before this, guards passed to gate methods were ignored.
ie:
`can($ability, $guard)`
was treated as just:
`can($ability)`


drbyte:
This may help close #702 and close #703

freekmurze:
I'm thinking it's pretty save to merge this. I don't think it's a breaking change.

Issue #550, opened 2017-11-02
slow hasPermissionTo


andrubeldie:
Hi,

I have the following context:
142 permissions attached to 27 roles, 2000 users - vast majority with one or more roles, only few with direct permissions, and a blade template that use @can for 101 times in order to generate a menu according to current user's permissions.

rendering this menu is very time consuming (~4 seconds) and I discovered that problem is with hasPermissionTo.

in order to not touch laravel-permissions files the solution was to overwrite the method in the my model and to do some cache:
```php
public function hasPermissionTo($permission, $guardName = null): bool
    {
        if (is_string($permission)) {
            $permission = \Cache::remember('spatie.permissions.'.$permission.'.'.$guardName, 3600, function()use($permission, $guardName){
                return app(Permission::class)->findByName(
                    $permission,
                    $guardName ?? $this->getDefaultGuardName()
                );
            });
        }
        return $this->hasDirectPermission($permission) || $this->hasPermissionViaRole($permission);
    }
```

what I achieved with this cache is to render that menu in 0.03 seconds.

I write you in hope you'll find a way to optimize this method inside this awesome laravel module :)

andrubeldie:
in order to test what is slow in the process, I used a console command like this:
```php
    public function handle()
    {
        Auth::loginUsingId(SOME_VALID_USER_ID);
        for ($i=0; $i<100; $i++){
            $time_start = microtime(true);
            auth()->user()->can('some_existent_permission');
            $time_end = microtime(true);
            $t = $time_end - $time_start;
            $time[] = $t;
        }
        dd($time, array_sum($time));
    }
```

AlexVanderbist:
Hi, 

Thanks for all your work looking into this. Ideally these cached permissions would come from `app(PermissionRegistrar::class)->getPermissions()` as they're already being cached there. 

This is definitely something we'll address in the next version of the package! 

Thanks again!

andrubeldie:
lovely!

I would also strongly recommend use ->filter() instead of ->where() on collections (i.e. findByName method of Permission::class) as is 10-15 times faster.

```php
    public function handle()
    {
        $collection_size = 1000;
        $faker = Faker::create();
        for ($i=0; $i < $collection_size; $i++) {
            $a[] = ['firstname' => $faker->firstNameMale, 'lastname' => $faker->lastName];
        }
        $a[] = ['firstname' => 'John', 'lastname' => 'Doe'];
        $c = collect($a);

        $iterations = 1000;
[...]

PR #710, opened 2018-04-09
improve speed of findByName and findById on permissions

barryvanveen:
This PR addresses issue #550. The `findByName` and `findById` methods were filtering the collection of permissions using the `where` method. Using `filter` on a collection is faster.

All tests still pass. I've also tested the integration with the (private) project I'm currently working on and everything looks good. Tests on my own project indicate a 10-15% performance increase when running these tests:
```php
Auth::loginUsingId(1); // change to active user

for ($i = 0; $i < 1000; ++$i) {
    $time_start = microtime(true);
    auth()->user()->can('my_permission_name'); // change to real permission name
    $time_end = microtime(true);
    $t = $time_end - $time_start;
    $time_by_name[] = $t;
}

for ($i = 0; $i < 1000; ++$i) {
    $time_start = microtime(true);
    auth()->user()->can(42); // change to real permission id
    $time_end = microtime(true);
    $t = $time_end - $time_start;
    $time_by_id[] = $t;
}

dd(array_sum($time_by_name), array_sum($time_by_id));
```

Please let me know if I can help to get this PR merged :)

Barry

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -84,7 +84,9 @@ class Permission extends Model implements PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
 
-        $permission = static::getPermissions()->where('name', $name)->where('guard_name', $guardName)->first();
+        $permission = static::getPermissions()->filter(function($permission) use ($name, $guardName) {
+            return $permission->name === $name && $permission->guard_name === $guardName;
+        })->first();
 
         if (! $permission) {
             throw PermissionDoesNotExist::create($name, $guardName);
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for submitting this!

Thoughts doing `findOrCreate` as well? 

What are your thoughts on doing similarly with the `Role` model? I can see there being some degree of benefit there, but given that your head is already on the subject, I'd be interested in your thoughts.

barryvanveen:
I think other methods will benefit from this optimization too.  However, in most cases it will have less of an impact because checking permissions is done more often. At least it will improve consistency throughout the code.

Will try to update this PR today and change other occurrences of the `where` methods.

barryvanveen:
All instances of the `Collection::where` method have been changed to `Collection::filter`. What remains are calls to the `where` method on the query builder, which is unrelated to the performance-issues mentioned in #550.

@drbyte, could you have another look at this PR?

drbyte:
Thanks @barryvanveen !

PR #715, opened 2018-04-15
Update PermissionRegistrar to use Authorizable instead of Authenticatable

### LINKED ISSUE #694: Does the package prevent the use of simple Authorizable models?
Hi,

The package is great, I have added roles and permissions to my site in less than two hours! The API is easy and very versatile. Many thanks for such value!

However, it seems that the package has broken the ability to use can/cant on a model that is only Authorizable.

I have added the roles and permissions to my User model and this part works fine. But I also have a Subscriber model which only implements Authorizable and uses the Authorizable trait in order to benefit from the can/cant methods. This part is now broken: when the can method is called on the Subscriber model, I get this Error:

```
Type error: Argument 1 passed to Spatie\Permission\PermissionRegistrar::Spatie\Permission\{closure}()
must implement interface Illuminate\Contracts\Auth\Authenticatable,
instance of App\Models\Subscriber given,
called in /home/vagrant/Code/laravel/vendor/laravel/framework/src/Illuminate/Auth/Access/Gate.php on line 314      
```
The before callback implemented in PermissionRegistrar makes the assumption that the passed $user is an Authenticatable, this is what breaks my call to the can method on my Authorizable model. 

I am not a specialist of Gates, but as far as I understand, any call to the Authorizable methods will pass through any callback that have been registered on the gate. So enforcing Authenticatable here will break any call to can on a model that is not an Authenticatable but is legit to use can.

Also, I see no use of any of the Authenticatable methods in this callback, so why is it enforced here? This restriction seems to be recent, it was added in 1f98264.

If I remove this Authenticatable type hint in the callback it works again, but I wanted to be sure, am I missing something here?

BTW, I am using Laravel 5.4 and laravel-permission 2.9.2.

elghobaty:
The package assumes that the model which is being assigned roles/permissions (e.g. User) implements the Authenticatable contract. This is not always the case. Laravel itself adds its authorization method, i.e. can to the Authorizable trait, and not Authenticatable.
This commit changes the type-hinted Authenticatable to Authorizable instead. Closes #694.

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -6,7 +6,7 @@ use Illuminate\Support\Collection;
 use Illuminate\Contracts\Auth\Access\Gate;
 use Illuminate\Contracts\Cache\Repository;
 use Spatie\Permission\Contracts\Permission;
-use Illuminate\Contracts\Auth\Authenticatable;
+use Illuminate\Contracts\Auth\Access\Authorizable;
 use Spatie\Permission\Exceptions\PermissionDoesNotExist;
 
 class PermissionRegistrar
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

PR #719, opened 2018-04-20
Support multiple db connections

### LINKED ISSUE #697: Specify the connection in the models
I'm getting the following error

`SQLSTATE[42P01]: Undefined table: 7 ERROR: relation "permissions" does not exist LINE 1: select * from "permissions" ^ (SQL: select * from "permissions")`

But the permissions table does exist in the database. in my project I have multiple database connections two of Postgres and one of Mysql, so in the models I always specify the connection.

`   protected $connection = 'pgsql';
    protected $table = 'users';
`

In the past I had a problem similar to that reported with the user model, because I did not recognize which connection to use to bring the information from the database, but in this case I can not solve it in the same way (well, I can but I should not) because the models are in vendor folder and therefore out of version control.

there is some correct way of doing what I tell you, some way of modifying the models, or being able to specify the connection as the migrations with the artisan command

`php artisan vendor:publish --provider="Spatie\Permission\PermissionServiceProvider" --tag="migrations"`

in the migrations I specify it in the following way

`Schema::connection('pgsql')->create($tableNames['permissions'], function (Blueprint $table) {
            $table->increments('id');
            $table->string('name');
            $table->string('guard_name');
            $table->timestamps();
        });`

What can I do? I'm waiting for your help.
Greetings.

drbyte:
To support storing the role/permission data and its associated models on a specific database connection, set the config `db_connection` (default is null) to the connection-name of a specific connection defined in your `config/database.php` connections array. ALL associated models using roles/permissions MUST have their `$connection` property set to the same name.

Closes #697

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,14 @@ return [
     ],
 
     /*
+     * Occasionally an app may use separate db connections for storing permission data
+     * for its models. You may specify which connection to use for role / permission
+     * tables as long as your models use the same name in their $connection property.
+     */
+
+    'db_connection' => null,
+
+    /*
      * By default all permissions will be cached for 24 hours unless a permission or
      * role is updated. Then the cache will be flushed immediately.
      */
--- a/config/permission.php
+++ b/config/permission.php
@@ -72,6 +72,14 @@ return [
     ],
 
     /*
+     * Occasionally an app may use separate db connections for storing permission data
+     * for its models. You may specify which connection to use for role / permission
+     * tables as long as your models use the same name in their $connection property.
+     */
+
+    'db_connection' => null,
+
+    /*
      * By default all permissions will be cached for 24 hours unless a permission or
      * role is updated. Then the cache will be flushed immediately.
      */
--- a/database/migrations/create_permission_tables.php.stub
+++ b/database/migrations/create_permission_tables.php.stub
@@ -11,25 +11,27 @@ class CreatePermissionTables extends Migration
      *
      * @return void
      */
-    public function up()
+    public function up($connection = null)
     {
+        $connection = $connection ?? config('permission.db_connection');
+
         $tableNames = config('permission.table_names');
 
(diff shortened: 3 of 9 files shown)


drbyte:
FEEDBACK WELCOME.  At the present time I'm not using this in an app that requires these multiple connections.
Is there any negative side-effect to this? Backward compatibility breaks I've missed?

freekmurze:
Thanks for your work on this @drbyte 

We could also instruct users to use a custom permission / role / ... model and set the `$connection` property on that custom model. That way no code changes to our code base are required to support a feature that isn't needed for most users.

patogalarzar:
Thank you for your help, sorry I have been busy for work and I have not been able to answer, I will apply your recommendation.

flashadvocate:
It's not immediately clear what the recommendation is here. I'm in a similar situation where my account / user table is on another database. Extending the models and providing the connection attributes did not solve the `Base table or view not found: 1146 Table 'auth.model_has_roles'` issue.

GeorgeGedox:
I cannot find this in V3. Was this removed ?

drbyte:
> I cannot find this in V3. Was this removed ?

It was never merged.

To support custom connections for your role/permission models, you can extend the model and set the Eloquent `$connection` property on that model.

syedhassan25:
hello hi guys  laravel spatie  package role base  permission  is very owsum but i am using in my project to different laravel projects
using database in both project single database but  its not working problem is i am assign permission to user in first project  its working fine ... but its user using in second project permissions not working... i see problem is ...  please tell me what we do i am stuck in my both project have you any idea pleaase suggest me.
problem conclusion

assign permission to user first project is woking in same project...
but this user is not  wrokign in second prject... but if you can assign  user  to permisssion  in second he is working fine in second project but dosent work in first project please help me  and sorry for poor englihs

PR #724, opened 2018-04-30
Added Events firing when assigning roles or permissions

leopado:
I was using this package to manage roles and permissions on a project.
However, I needed to keep a log of the permissions and roles that were assigned or revoked to users.
I saw that it was not present so I added support for these events. I have not added the events on the creation, modification and deletion because for these it is sufficient to create an observer on Role and Permission models.
In addition, I also managed the activation of this feature via config.
Probably the implementation is far from perfect but I hope you can find it useful too to increase the potential of this already very interesting package.

Best Regards,
Leonardo Padovani.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,8 +26,9 @@
         "illuminate/database": "~5.4.0|~5.5.0|~5.6.0"
     },
     "require-dev": {
+        "mockery/mockery": "^1.0",
         "orchestra/testbench": "~3.3.0|~3.4.2|^3.5.0",
-        "phpunit/phpunit" : "^5.7|6.2|^7.0"
+        "phpunit/phpunit": "^5.7|6.2|^7.0"
     },
     "autoload": {
         "psr-4": {
--- a/config/permission.php
+++ b/config/permission.php
@@ -85,4 +85,8 @@ return [
      */
 
     'display_permission_in_exception' => false,
+    /*
+     * If set to false Custom Events will not be fired when roles/permission is assigned
+     */
+    'events_enabled' => true,
 ];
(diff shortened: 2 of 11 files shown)


drbyte:
Thanks for sharing.
I hope to review it further later this week.

drbyte:
Hi @leopado 
Thanks for submitting this PR.
Clearly you've put a lot of work into writing all these additions.

I'm going to confer with @freekmurze for further input, as I think these changes may be beyond the scope of what this package is aiming to provide.

freekmurze:
It's a lot of code here for us to maintain.

Couldn't you just rely on the events that Eloquent already fires?

leopado:
Hi, thank for your review.

Initially I sure wanted to rely on events fired by Eloquent and for creating/updating and deleting roles or permissions.  

But when a role or a permission is assigned to someone I wasn't able to listen on Eloquent events and
my recent projects need to log these actions also to be compliant with the recent GDPR privacy law.

If you know a best way to do it using the native events of Eloquent I'll be happy to use them instead of forking your package :)

Thanks a lot.

ChrisSFR:
Just a short question. The events for syncing models is something I am currently facing. Precisely because nova came out and I want to call some actions, when a role gets assigned.

Maybe a possible solution: creating a MorphPivot Model. Then package users could observe the Pivot create/delete actions and fire events using the built in Eloquent events like @freekmurze suggested.

But that requires to add withPivot methods to the HasRoles trait and the Role Model. Furthermore creating the pivot models.

I don't know if its possible, but I'll try it out. Any doubts about this?

ChrisSFR:
>  I don't know if its possible, but I'll try it out. Any doubts about this?
Ok it seems laravel doesn't trigger sync and attach actions on pivot model events. So my suggestion will not work.

robjbrain:
I gave an example of how to add your own events in the simplest way in this issue: https://github.com/spatie/laravel-permission/issues/754#issuecomment-468229002

PR #732, opened 2018-05-07
Changed the caching strategy, to improve lookup speed. 

klincheg:
changed getPermissions method

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -34,7 +34,7 @@ class Permission extends Model implements PermissionContract
     {
         $attributes['guard_name'] = $attributes['guard_name'] ?? Guard::getDefaultName(static::class);
 
-        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
+        $permission = static::getPermissions(['name' => $attributes['name'], 'guard_name' => $attributes['guard_name']])->first();
 
         if ($permission) {
             throw PermissionAlreadyExists::create($attributes['name'], $attributes['guard_name']);
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -106,7 +106,6 @@ class Permission extends Model implements PermissionContract
     public static function findById(int $id, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-
         $permission = static::getPermissions(['id' => $id, 'guard_name' => $guardName])->first();
 
         if (! $permission) {
@@ -127,10 +126,7 @@ class Permission extends Model implements PermissionContract
     public static function findOrCreate(string $name, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-
-        $permission = static::getPermissions()->filter(function ($permission) use ($name, $guardName) {
-            return $permission->name === $name && $permission->guard_name === $guardName;
-        })->first();
+        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
 
         if (! $permission) {
             return static::create(['name' => $name, 'guard_name' => $guardName]);
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -34,9 +34,7 @@ class Permission extends Model implements PermissionContract
     {
         $attributes['guard_name'] = $attributes['guard_name'] ?? Guard::getDefaultName(static::class);
 
-        $permission = static::getPermissions()->filter(function ($permission) use ($attributes) {
-            return $permission->name === $attributes['name'] && $permission->guard_name === $attributes['guard_name'];
-        })->first();
+        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
 
         if ($permission) {
             throw PermissionAlreadyExists::create($attributes['name'], $attributes['guard_name']);
@@ -87,11 +85,7 @@ class Permission extends Model implements PermissionContract
     public static function findByName(string $name, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-
-        $permission = static::getPermissions()->filter(function ($permission) use ($name, $guardName) {
-            return $permission->name === $name && $permission->guard_name === $guardName;
-        })->first();
-
+        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
         if (! $permission) {
             throw PermissionDoesNotExist::create($name, $guardName);
         }
@@ -113,9 +107,7 @@ class Permission extends Model implements PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
 
-        $permission = static::getPermissions()->filter(function ($permission) use ($id, $guardName) {
-            return $permission->id === $id && $permission->guard_name === $guardName;
-        })->first();
+        $permission = static::getPermissions(['id' => $id, 'guard_name' => $guardName])->first();
 
         if (! $permission) {
             throw PermissionDoesNotExist::withId($id, $guardName);
@@ -150,8 +142,8 @@ class Permission extends Model implements PermissionContract
     /**
      * Get the current cached permissions.
      */
-    protected static function getPermissions(): Collection
+    protected static function getPermissions($params = null): Collection
     {
-        return app(PermissionRegistrar::class)->getPermissions();
+        return app(PermissionRegistrar::class)->getPermissions($params);
     }
 }
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -42,12 +42,12 @@ class PermissionRegistrar
 
     public function forgetCachedPermissions()
     {
-        $this->cache->tags('spatie')->flush();
+        $this->cache->tags($this->cacheKey)->flush();
     }
 
     public function getPermissions($params = null): Collection
     {
-        return $this->cache->tags('spatie')->remember($this->cacheKey.($params ? '.'.implode('.', array_values($params)) : ''), config('permission.cache_expiration_time'), function () use ($params) {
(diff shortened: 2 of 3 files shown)


klincheg:
thanks. Will fix it on monday - it will flush only spatie cache.

drbyte:
Hi @klincheg, Thanks for submitting this.

1. First, it appears that your code is reversing the performance improvements added in #710.  Please revert your changes to those parts of the Permissions model. 

2. It appears that this PR is intended to improve the caching-lookup speed by adding more granular cache key levels.
Perhaps we can re-title this PR to indicate that it's changing the caching strategy, to improve lookup speed?

klincheg:
@drbyte thanks for your comment.  Will try to explain my changes. 
1. I can't revert it because in my PR code selects only needed data. In previous version it was "select all and after that filter" But in old version i have a time problem if there are more than 150 permissions in the table (code workedin some cases more than 2-3 secs, and after my changes time was reduced to 300-500ms) 
2. PR title changed for more relative.

drbyte:
@drbyte wrote:
> it appears that your code is reversing the performance improvements added in #710. Please revert your changes to those parts of the Permissions model.

@klincheg replied:
> I can't revert it because in my PR code selects only needed data. In previous version it was "select all and after that filter" But in old version i have a time problem if there are more than 150 permissions in the table (old code in some cases worked more than 2-3 secs, and after my changes time was reduced to 300-500ms)

There were benchmark measurements done in #710 and #550 both for exposing performance problems and demonstrating the improvements realized by the changes.
I'm not interested in undoing the improvements made by #710 without substantial proof that they are problematic. In my experience they were positive improvements.

Are you saying that the changes in #710 actually made performance worse?
If so, then let's create a test application to demonstrate the problem, showing before-and-after speed statistics.

Nowi5:
Cab we may define per config if the system should load all entries from DB or only the needed? I think the developer should decide base on the use case. As I wrote at #795 we need to get more flexible on that, which also depends on environment (data/file cache vs mem etc)

fullstackfool:
Tested this locally on a large SaaS product and the speed increase was dramatic.

Would it be possible to get a very rough estimate of when this will be merged in?

Is there further work needed? If so I would be willing to help if needed.

drbyte:
@fullstackfool I've been meaning to set up a few scenarios to run performance benchmarks.
I figure creating a few seeders which can pre-populate the simulated scenarios, and then running some hits against it, should suffice.
You could help by either offering a checklist of components to consider in building the seeders, or offering such a seeder. 
Since you've posted that this has helped you, I'd like to know more about your environment, such as number of roles, permissions, users-with-roles, users-with-specific-permissions, variations of user models (or any other models associated with roles/permissions), whether you focus primarily on permissions vs roles, and the sort of app execution path that uses the roles/permissions (web-based hits put load on roles/permissions via routes, controllers, model-policies, and blade templates; api hits are more likely to hit a middleware or controller, but not templates). Sorry, that list got big very quickly!

drbyte:
I've done a bunch of simulating and testing with just creating hundreds of roles/permissions/users and linking them up (not necessarily reflective of any "specific" app type, just very generic), and this PR gives about a 5x performance boost. So that's good. 

Unfortunately this PR relies on a cache system that supports tagging (cannot use `flush()` if tagging is not supported).

Further, when using a fallback if `TaggableStore` is not detected, tests using `CACHE_DRIVER` of `array` and `file` gave different cached-response results, so there's something inconsistent there. Maybe the length of the cache key?

Maybe you wanna test the alterations I've been working with:


```php
class PermissionRegistrar
{
    //
    /** @var boolean */
    protected $cacheIsTaggable = false;

    public function __construct(Gate $gate, Repository $cache)
    {
        //
        $this->cacheIsTaggable = $cache->getStore() instanceOf \Illuminate\Cache\TaggableStore;
    }

    //
    public function forgetCachedPermissions()
    {
        if ($this->cacheIsTaggable) {
            $this->cache->tags($this->cacheKey)->flush();
            return;
        }

        $this->cache->forget($this->cacheKey);
    }

    public function getPermissions($params = null): Collection
    {
        $permissionClass = $this->getPermissionClass();

        if ($this->cacheIsTaggable) {
            $key = $this->cacheKey . ($params ? '.' . implode('.', array_values($params)) : '');
            return $this->cache
                ->tags($this->cacheKey)
                ->remember($key, config('permission.cache_expiration_time'), function () use ($params, $permissionClass) {
                    if ($params) {
                        return $permissionClass->where($params)->with('roles')->get();
                    }
                    return $permissionClass->with('roles')->get();
                });
        }

        return $this->cache
            ->remember($this->cacheKey, config('permission.cache_expiration_time'), function () use ($params, $permissionClass) {
                return $permissionClass->where($params)->with('roles')->get();
            });
    }
```

fullstackfool:
So far we've tried this with a subset of our users, ~400, with ~500 permissions. Everything is role-based for the moment (one of the main reasons for us experimenting with this package is to allow direct user->permission association). This gives a user/role combination of slightly over 1200. Roles/permissions is over 2100. Quick query shows we have a user/permission count of ~43000.

There's another specific set of users, ~500, who are mostly associated with another subset of permissions, which we are considering new ways of handling. This set of 'permissions' is more along the lines of `a` can do `b` when `c`.`d`.`e`. This leads to a possible permission list of >400,000. Of course, we're not sure whether this should be treated as a set of permissions yet - we may be trying to fit a square peg in a round hole, but your proposed alterations certainly give more felixibility to the system.

I'll have a test with the above code see how far I can push it before flames erupt from my rig.

drbyte:
@fullstackfool wrote: 
> This set of 'permissions' is more along the lines of a can do b when c.d.e. This leads to a possible permission list of >400,000. Of course, we're not sure whether this should be treated as a set of permissions yet - we may be trying to fit a square peg in a round hole

Ya, sometimes it's tempting to use "dynamic" permissions where the rules are really app-encoded and don't need the added overhead of being stored in the db. In such cases I find that implementing a strategy geared around Policy use is not only more performant, but also easier to debug and easier for the dev team to grok.

fullstackfool:
Sorry for the late response, hectic here.

With the section below, are we not overwriting each result with the next permission call?

```
return $this->cache
            ->remember($this->cacheKey, config('permission.cache_expiration_time'), function () use ($params, $permissionClass) {
                return $permissionClass->where($params)->with('roles')->get();
            });
```

Imagine you meant: `$key = $this->cacheKey . ($params ? '.' . implode('.', array_values($params)) : '');`

The tests so far look good, 100 roles, 250 users, assigned 1-4 roles, assigned 30-500 permissions each. 

I'm also playing with caching the user-permission result further up the chain. This improves the speed even further, for example one set of tests went from 76 seconds to 4!

fullstackfool:
Hopefully the below isn't too hideous. As stated above, this results in speed increases over an order of magnitude faster.

Not shown below but `PermissionRegistrar` has also been made a singleton.

Caching the result of a permission check directly against the model instance also would allow a user to clear cache for a single user, as opposed to the entire permissions cache.

```php
    /**
     * Determine if the model may perform the given permission.
     *
     * @param string|int|\Spatie\Permission\Contracts\Permission $permission
     * @param string|null                                        $guardName
     *
     * @return bool
     */
    public function hasPermissionTo($permission, $guardName = null): bool
    {
        if (PermissionRegistrar::$cacheIsTaggable) {
            $config = config('permission');

            return cache()
                ->tags($this->getCacheTags($config, $permission))
                ->remember($this->getCacheKey($config, $permission), $config['cache_expiration_time'],
                    function () use ($permission, $guardName) {
                        return $this->checkHasPermissionTo($permission, $guardName);
                    });
        }

        return $this->checkHasPermissionTo($permission, $guardName);
    }

    /**
     * Check the uncached permissions for the user.
     *
     * @param      $permission
     * @param null $guardName
     *
     * @return bool
     */
    public function checkHasPermissionTo($permission, $guardName = null)
    {
        $permissionClass = $this->getPermissionClass();

        if (is_string($permission)) {
            $permission = $permissionClass->findByName(
                $permission,
                $guardName ?? $this->getDefaultGuardName()
            );
        }

        if (is_int($permission)) {
            $permission = $permissionClass->findById(
                $permission,
                $guardName ?? $this->getDefaultGuardName()
            );
        }

        if (!$permission instanceof Permission) {
            throw new PermissionDoesNotExist;
        }

        return $this->hasDirectPermission($permission) || $this->hasPermissionViaRole($permission);
    }

    /**
     * Construct the key for the cache entry.
     *
     * @param array $config
     * @param       $permission
     *
     * @return string
     */
    private function getCacheKey(array $config, $permission)
    {
        $permission = $this->getPermissionKey($config, $permission);

        return implode('.', [
            $config['cache_key'],
            get_class($this),
            $this->getKey(),
            $config['models']['permission'],
            $permission,
        ]);
    }

    /**
     * Construct the tags for the cache entry.
     *
     * @param array $config
     * @param       $permission
     *
     * @return string
     */
    private function getCacheTags(array $config, $permission)
    {
        $permission = $this->getPermissionKey($config, $permission);

        return [
            $config['cache_key'],
            $config['models']['permission'] . '.' . $permission,
            get_class($this) . '.' . $this->getKey(),
        ];
    }

    /**
     * Get the configured key to cache the model by.
     *
     * @param array $config
     * @param       $permission
     *
     * @return mixed
     */
    private function getPermissionKey(array $config, $permission)
    {
        if ($permission instanceof $config['models']['permission']) {
            $permission = $permission[$config['cache_attribute']];
        }

        return $permission;
    }
```

fullstackfool:
I'd like to open this as a separate PR, if that's ok? It will still contain the great work from @klincheg and the alterations from @drbyte, but with the additional model-permission caching.

drbyte:
Sure. It'll be easier to discuss as a separate PR.

klincheg:
> I'd like to open this as a separate PR, if that's ok? It will still contain the great work from @klincheg and the alterations from @drbyte, but with the additional model-permission caching.

it will be cool ) thanks 
PS sorry. Was offline during long time. If I can help you just let me know )

drbyte:
Closed in favor of #926

Issue #479, opened 2017-09-28
Next major release -- Features/Requests for v3


AlexVanderbist:
It feels like the last couple of months a lot of issues were related to the multi-guard functionality  introduced in v2. Apart from that there have been some vague caching issues as well. Time to do something about that in v3 of the package.

## The multi-guard issue

Supporting multiple guards has added a lot of complexity to the package and made it feel quite "heavy". The multi-guard functionality is definitely a selling point for this package but it also seems to confuse some users that are new to the framework and haven't learned about guards yet. Not to mention the countless times a misconfiguration in `auth.php` causes the wrong guard to be used.

Some possible solutions:
- completely dropping support for guards - as @drbyte and I discussed at Laracon removing all support for multiple guards would make the package a lot easier to maintain, simpler and lighter.
- my personal favourite: drop guard support but keep `Permission`s and `Role`s polymorphic relations. This means any model can have roles and permissions but there's no extra functionality related to guards.
- maintaining two branches: one with- and one without support for guards.

## Caching issue

There's no interface or CLI tool that comes with the package to create and manage roles/permissions. This means a lot of users revert to editing the DB to manage permissions causing the cache to be rendered invalid. Even though it's well documented this seems to be an issue that keeps coming back. Maybe caching should be limited to a single request lifecycle by default and configurable for anyone that needs more anything else?

------

Any thoughts on these two issues or suggestions for other features for v3 of the package are very welcome!

freekmurze:
👍👍👍 for simplifying this package.

> maintaining two branches: one with- and one without support for guards.
It'll be too much work, so let's don't do that.

> completely dropping support for guards
The nuclear option, but given that not so much users need multi-guard support, this is certainly an option.

>  This means any model can have roles and permissions
How would that look like in the db? I'm guessing not many people are going to need this, so if we go this route it has to be really really simple.

> Maybe caching should be limited to a single request lifecycle by default and configurable for anyone that needs more anything else
Good idea, let's do that!

Let's also drop support for anything below Laravel 5.5

bloggervista:
@freekmurze  Thanks. Its very hard to develop the app for mobile device using this package and that is due to name spacing issue. I guess we can drop guards  so that it will work on both. But we cannot ignore the fact that using guards have some use case. I guess we can decide using guard or not according to null value on table etc..

And I do agree dropping support below laravel 5.5 is good as Laravel 5.5 is LTS

barryvanveen:
[...]

PR #737, opened 2018-05-13
add $table->string('display_name');

vahidalvandi:
i added field display name in permissions and roles .
when create a role must select same permission with name in english , we need save title permission in other languages like persian but have name in english .

ویرایش پست
edit-post

### DIFF
--- a/database/migrations/create_permission_tables.php.stub
+++ b/database/migrations/create_permission_tables.php.stub
@@ -18,6 +18,7 @@ class CreatePermissionTables extends Migration
         Schema::create($tableNames['permissions'], function (Blueprint $table) {
             $table->increments('id');
             $table->string('name');
+            $table->string('display_name');
             $table->string('guard_name');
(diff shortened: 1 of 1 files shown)


ankurk91:
@vahidalvandi
How are you supposed to store `display_name` by just updating the schema?
You should be discussing the topic here
https://github.com/spatie/laravel-permission/issues/479

drbyte:
@vahidalvandi 
Thanks for your submission. 

The tests are failing because you didn't mark those fields as `nullable()`, and didn't update the models to insert a non-null name if no display-name was set.

As @ankurk91 pointed out, there's more that needs to be done to this before it could be merged: there would need to be method calls that allow setting the display-name programmatically, as well as tests included to confirm it functions correctly.

We hope to consider additional fields as part of v3, but haven't fully determined what features will actually be in v3 yet. 

Closing this PR for now.

PR #742, opened 2018-05-19
Readme: changed order of steps in lumen installation

napestershine:
right now if run migration it gives error with empty table name. So first user developer need to define settings in app.php then migration run without trouble.


drbyte:
Thanks!

Issue #440, opened 2017-09-02
I wrote an improved version of PermissionMiddleware with hierarchical capabilities


eduardoarandah:
I wrote an improved version of PermissionMiddleware with hierarchical capabilities, might be of your interest

TLDR; checks permissions in a hierarchical structure, this permission **admin.auth.user.modify.create** will check:

- admin
- admin.auth
- admin.auth.user
- admin.auth.user.modify
- admin.auth.user.modify.create

[https://github.com/eduardoarandah/spatie-laravel-permission-permission-middleware](https://github.com/eduardoarandah/spatie-laravel-permission-permission-middleware)

freekmurze:
Kinda interesting, but holding off on this for now.

Issue #382, opened 2017-07-30
(Improvement) Wildcard Permission.


twf-nikhila:
What do you think about having a wild card permission, something like:

- If a user/role is assigned star (*) Permission, he will have all permission with him? 
It will be useful for assigning to admin.

Your thoughts? If you find useful I can send a PR.

freekmurze:
You can already give admins all permission by leveraging the `before` method of the `Gate`. You could use that to return `true` for all admins.

More info: https://laravel.com/docs/5.4/authorization

benyanke:
I would also find wildcard permission, but with string checks.

For example, if I have the following permissions:

- sales.emailClient
- sales.emailInternal
- sales.viewClients
- marketing.viewJobs
- marketing.makeAd
- marketing.viewJobs

And the following roles:

- MarketingUser
- SalesUser

I could apply all marketing permissions to a marketing user by running 

```php
$role->givePermissionTo('marketing.*');
```

This would be extremely valuable for my workflow. Any chance this might get added?

drbyte:
To accomplish similarly, in one of my projects I have separate collections for the grouped permissions, and then I just assign (or sync) those collections to the role.
(Using this approach means the "department" prefixes "may" be redundant, depending on the implementation. Of course, the examples given are simple, so prefixes may serve a larger purpose than just visual cues.)

ie: using your example above, would go something like:
```php
$marketingRole->givePermissionTo($marketingPermissions);

$salesRole->givePermissionTo($salesPermissions);

$managerRole->givePermissionTo($marketingPermissions);
$managerRole->givePermissionTo($salesPermissions);

...
```

benyanke:
Not quite as flexible as I'd hoped, but great idea @drbyte - that'll help me move along with this project much quicker! I'll probably be having hundreds of permissions in this app, so this will be a major time saver.

drbyte:
Somewhat related: Awhile back someone posted an alternate middleware to handle some cascading levels, which might also be of interest: https://github.com/spatie/laravel-permission/issues/440

drbyte:
@benyanke Would the permissions wildcard code referenced in #440 do what you were looking for ?
Or were you really more interested in doing Role assignment using wildcard patterns?

benyanke:
Yes, that would cover every use case I am using wildcards for.

Currently I've hacked things into place with a preprocessor which loops though all available permissions and assigns them one at a time, selecting via the wildcard and LDAP group membership. 

[...]

PR #746, opened 2018-05-23
Replacing static Permission::class and Role::class with dynamic value

lk77:
Hello,

i'm proposing this change, 
which remove all Permission::class and Role::class and using the classes defined in the config instead,
to prevent the overriding of user preferences and the fallback to the default class that appen systematically  when using custom models.

now, with this change, if you override a method, the default one from the default model will not be called anymore, your custom method will be called every time.

tell me what you think.

the main purpose is to be able to modify the logic with heritage (adding wildcard implementation in my case), without beeing forced to modify the package core code.

PS: if you think wildcard capabilities is something useful for this package, i can try to implement it in another PR if you want.

thanks.

```
PHPUnit 6.2.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.0alpha3
Configuration: /var/www/laravel-permission/phpunit.xml.dist

...............................................................  63 / 127 ( 49%)
............................................................... 126 / 127 ( 99%)
.                                                               127 / 127 (100%)

Time: 1.79 seconds, Memory: 26.00MB

OK (127 tests, 223 assertions)

```

### DIFF
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -138,12 +138,14 @@ class Role extends Model implements RoleContract
      */
     public function hasPermissionTo($permission): bool
     {
+        $permissionClass = $this->getPermissionClass();
+
         if (is_string($permission)) {
-            $permission = app(PermissionRegistrar::class)->getPermissionClass()->findByName($permission, $this->getDefaultGuardName());
+            $permission = $permissionClass->findByName($permission, $this->getDefaultGuardName());
         }
 
         if (is_int($permission)) {
-            $permission = app(PermissionRegistrar::class)->getPermissionClass()->findById($permission, $this->getDefaultGuardName());
+            $permission = $permissionClass->findById($permission, $this->getDefaultGuardName());
         }
 
         if (! $this->getGuardNames()->contains($permission->guard_name)) {
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -4,6 +4,7 @@ namespace Spatie\Permission\Models;
 
 use Spatie\Permission\Guard;
 use Illuminate\Database\Eloquent\Model;
+use Spatie\Permission\PermissionRegistrar;
 use Spatie\Permission\Traits\HasPermissions;
 use Spatie\Permission\Exceptions\RoleDoesNotExist;
 use Spatie\Permission\Exceptions\GuardDoesNotMatch;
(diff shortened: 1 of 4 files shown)


drbyte:
Question: does your own implementation currently extend the model itself, or the Contracts?
Ref: https://github.com/spatie/laravel-permission/blob/master/src/PermissionServiceProvider.php#L52-L58

drbyte:
@lk77 wrote:
> PS: if you think wildcard capabilities is something useful for this package, i can try to implement it in another PR if you want.

If you're talking about something different than #382 or #440 then please open an Issue to discuss this. I'd like to understand more about what you envision. Or add to the discussion in #382 if it's the same thing as being discussed there.

lk77:
My implementation override the model itself by extending it, and the default model is still beeing called,
And yes it's the same kind of things that #382

lk77:
I've updated my PR, using config helper is too time consuming.
```
PHPUnit 6.2.3 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.0alpha3
Configuration: /var/www/laravel-permission/phpunit.xml.dist

...............................................................  63 / 127 ( 49%)
............................................................... 126 / 127 ( 99%)
.                                                               127 / 127 (100%)

Time: 1.81 seconds, Memory: 26.00MB

OK (127 tests, 223 assertions)
```

lk77:
And also, i think that we should replace getPermissions function in Permission model by something like that : 

```
public static $permissions;
[...]
    /**
     * Get the current cached permissions.
     */
    protected static function getPermissions(): Collection
    {
        if(static::$permissions === null)
        {
            static::$permissions = app(PermissionRegistrar::class)->getPermissions();
        }

        return static::$permissions;
    }
```

Calling the cache for each record is too slow

lk77:
```
Question: does your own implementation currently extend the model itself, or the Contracts?

```

I'm updating the config at runtime in my service provider, so the registered class do not match anymore.

drbyte:
@lk77 I see you went through several iterations here, presumably because you found more efficient ways of doing it.
Did you discover any potential breaking changes with this?
Can you think of a Feature Test that we could add to make sure it is correctly fixed?

lk77:
For me there is no breaking change, unless someone changed the default class by a custom one dynamically and there is a bug in the custom one, but it was never called, and now it will be.

We could add a test to ensure that the new behaviour works alongside the old one.

drbyte:
👍

PR #760, opened 2018-06-07
Shorten generated primary-key name

vcastro-overall:
This error show when execute "php artisan migrate" in mysql 5.5


vcastro-overall:
spatie-error_mysql5 5

drbyte:
Thanks!

I've updated it to keep the `_primary` suffix, since that's still the same overall length as the other generated index names, and retains some of the semantics.

Issue #765, opened 2018-06-11
Determine if a model has all permissions


SamuelNitsche:
First of all thanks for this package, it helps me a lot!
I am wondering if there is a way to determine if a model has all permissions like the method `hasAllRoles`. I could not find something like `hasAllPermissions`.
Thanks!

drbyte:
Not at the present time.

Can you describe the scenario you're working with and the problem you're trying to solve?

drbyte:
I see your point.

When I run into this sort of scenario it makes me strongly lean towards abstracting that combination of permissions into a policy. This way I can also name it something semantic (which also encourages me to consider additional opportunities elsewhere in my code).

One thing you could do in the short term is change `hasPermissionTo()` to just `can()`, since all permissions are pushed into the Gate so calls to `can()` and `@can()` can be used in your code and templates respectively.

drbyte:
We're happy to review all pull requests! Especially those that are explained well and include supporting tests.

While you're coding ...
One thing to keep in mind about permissions (that's different from roles) is that since roles currently support only one tier, the "hasAnyRole" is a very simple lookup. But permissions might exist in other roles held by the user/model. 
While simply making `hasAnyPermission($permissions)` do a looped lookup is "one" way to tackle it, there may be a more performant solution that could be explored.

PR #768, opened 2018-06-12
Added method hasAllPermissions

SamuelNitsche:
Hi :)

As already announced (#765), I added a method to this pull request to check if a model has all permissions passed. This functionality is very similar to `hasAllRoles` so I think it it fits very good into this package. I think the way I implemented this funtionality is the most performant and also the best way to do.
I have added tests that are all passing.

```
$ vendor/bin/phpunit
PHPUnit 7.2.4 by Sebastian Bergmann and contributors.

Runtime:       PHP 7.2.6
Configuration: C:\Users\samuelnitsche\Documents\Development\laravel\laravel-permission\phpunit.xml.dist

...............................................................  63 / 130 ( 48%)
............................................................... 126 / 130 ( 96%)
....                                                            130 / 130 (100%)

Time: 2.21 seconds, Memory: 26.00MB

OK (130 tests, 229 assertions)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,13 +374,13 @@ $user->hasPermissionTo(Permission::find(1)->id);
 $user->hasPermissionTo($somePermission->id);
 ```
 
-...or if a user has any permissions:
+You can test if a user has Any of an array of permissions:
 
 ```php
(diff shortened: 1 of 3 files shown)


drbyte:
Thanks!

PR #785, opened 2018-07-08
sync active routes with permissions

khaninejad:
I always waste too much time to create new permissions and add them to my cms so I decided to create a commend that help me to solve this issue.
this commend will look all of your active routes and add them to permissions then you can handle this permission with below command:
```
public function __construct(Request $request)
    {
        if (!App::runningInConsole()) {
            $action=$request->route()->getAction();
// second option for excluded routes
            if (!in_array($action['as'], ["cars.list",'cars.getmodels'])) {
                $this->middleware(['permission:'.$action['as']]);
            }
        }
    }
```

### DIFF
--- a/src/Commands/SyncPermissions.php
+++ b/src/Commands/SyncPermissions.php
@@ -1,0 +1,74 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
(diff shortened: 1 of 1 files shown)


freekmurze:
Thank you for this, but this functionality is out of scope for this package.

I suggest you create a repo of your own to maintain the code.

PR #809, opened 2018-07-31
 Fixed HasRoles::hasRole() function.  is_array does not work to Illuminate\Support\Collection

zgldh:
is_array does not work to Illuminate\Support\Collection
It needs a proper type detection.

On Laravel 5.5

```
   $collection = new Illuminate\Support\Collection([11, 33, 55]);
   var_dump(is_array($collection);
```

It outputs false. So is_array should not apply to Illuminate\Support\Collection object.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -142,7 +142,7 @@ trait HasRoles
             return $this->roles->contains('id', $roles->id);
         }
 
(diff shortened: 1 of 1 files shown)


zgldh:
I should update my Permission model with:
```
    protected $casts = [
        'roles'      => 'array'
    ];
```

PR #814, opened 2018-08-03
Modify hasPermissionTo & hasDirectPermission to return boolean

### LINKED ISSUE #802: hasPermissionTo method should return bool
When there is no permission in the database, this method currently returing an exception. It should return false

```[2018-07-21 17:24:04] local.ERROR: There is no permission named `list-products` for guard `web`. {"userId":1,"email":"admin@admin.com","exception":"[object] (Spatie\\Permission\\Exceptions\\PermissionDoesNotExist(code: 0): There is no permission named `fetch-products` for guard `web`. at /app/vendor/spatie/laravel-permission/src/Exceptions/PermissionDoesNotExist.php:11)```

### PR DESCRIPTION
**Pull request by Te7a-Houdini**
this **PR** , resolves #802 .

i extracted the filtering logic in **Permission model** , to a method called **filterPermissionsBy** , and in **HasPermissions Trait** , i extracted the logic of checking wither the permission is **boolean** or **string** to a public method in **Permission model** called **findByNameOrId** .

with these extractions i removed some of duplicated logic like this :

```php
static::getPermissions()->filter(function ($permission) use ($name, $guardName) {
            return $permission->name === $name && $permission->guard_name === $guardName;	
})->first();
```
one last thing , i think this PR **will cause a breaking changes** , for example if someone was doing this , then this code will **never throws an exception** :
```php

try
{
   $user->hasPermissionTo('some permission')
}
catch(Exception $e)
{
   Permission::create(['name' => 'some permission']);
}
```

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -33,7 +33,7 @@ class Permission extends Model implements PermissionContract
     public static function create(array $attributes = [])
     {
         $attributes['guard_name'] = $attributes['guard_name'] ?? Guard::getDefaultName(static::class);
-        
+
         $permission = static::findByNameOrId($attributes['name'], $attributes['guard_name']);
 
         if ($permission) {
@@ -157,7 +157,7 @@ class Permission extends Model implements PermissionContract
     }
     
     /**
-     * Filters permissions by a given (key , value) and guardName
+     * Filters permissions by a given (key , value) and guardName .
      *
      * @param string $key
      * @param string|int $value
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -162,7 +162,7 @@ class Permission extends Model implements PermissionContract
      * @param string $key
      * @param string|int $value
      * @param string $guardName
-     * 
+     *
      * @return \Illuminate\Support\Collection
      */
     protected static function filterPermissionsBy(string $key, $value, string $guardName)
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -155,7 +155,7 @@ class Permission extends Model implements PermissionContract
 
         return static::filterPermissionsBy('id', $permission, $guardName)->first();
     }
-    
+
     /**
      * Filters permissions by a given (key , value) and guardName .
      *
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -33,10 +33,8 @@ class Permission extends Model implements PermissionContract
     public static function create(array $attributes = [])
     {
         $attributes['guard_name'] = $attributes['guard_name'] ?? Guard::getDefaultName(static::class);
-
-        $permission = static::getPermissions()->filter(function ($permission) use ($attributes) {
-            return $permission->name === $attributes['name'] && $permission->guard_name === $attributes['guard_name'];
-        })->first();
+        
+        $permission = static::findByNameOrId($attributes['name'], $attributes['guard_name']);
 
         if ($permission) {
             throw PermissionAlreadyExists::create($attributes['name'], $attributes['guard_name']);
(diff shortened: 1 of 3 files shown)


src/Models/Permission.php
```diff
@@ -147,6 +139,39 @@ public static function findOrCreate(string $name, $guardName = null): Permission
...
 
+    /**
+     *  Find a permission by its name or id (and optionally guardName).
+     *
+     * @param string|int $permission
+     * @param string|null $guardName
+     *
+     * @return null|\Spatie\Permission\Contracts\Permission
+     */
+    public static function findByNameOrId($permission, $guardName = null)
```
drbyte: What are your thoughts about whether this public method should also be part of the interface? 

I'm okay with leaving it `public` since it can probably be adapted to support uuid/guid (which are strings). But that makes me wonder whether another name might be better for this method.  (If it was protected/private then renaming is more trivial.)
What were your thoughts when coming up with naming here?

Te7a-Houdini: @drbyte actually i thought about a method that encapsulates the logic of (is_string & is_int ) in HasPermission trait .

and while i was thinking, i found that the logic is, wither to find a permission by  name in case the passed $permission parameter is string , or find a permission by id in case the  passed $permission parameter is integer.

so i named the method with findByNameOrId.

[example of is_string then calling findByName and is_int then calling findById](https://github.com/spatie/laravel-permission/blob/master/src/Traits/HasPermissions.php#L108)

if you have a better suggestion or a better approach i would appreciate it .

drbyte: Okay -- your thinking process was the same as mine. :)

I want to think a bit further about the best naming for the public method, since I'd like to expand to uuid as well.
And then must decide whether the Interface/Contract should require the method also.

tests/HasPermissionsTest.php
```diff
@@ -188,19 +188,15 @@ public function it_can_determine_that_the_user_does_not_have_a_permission()
     }
 
     /** @test */
-    public function it_throws_an_exception_when_the_permission_does_not_exist()
+    public function it_returns_false_when_the_permission_does_not_exist()
```
drbyte: @AlexVanderbist are you okay with this no longer throwing an exception if permission does not exist? 
Is this sort of change too significant to not be considered a breaking change in Spatie terms?

src/Models/Permission.php
Te7a-Houdini: @drbyte , for  the contract i think it should be required , and for naming am not quite sure , but i think we can make something like this : 

```php
public static function findByColumn($permission, $guardName = null, $column = null)
 {
    if (!$column) {
        $column = is_string($permission) ? 'name' : 'id';
     }

     return static::filterPermissionsBy($column, $permission, $guardName)->first();
 }
```

and also change
```php
protected static function filterPermissionsBy(string $key, $value, string $guardName)

//instead of $key to be $column

protected static function filterPermissionsBy(string $column, $value, string $guardName)
```

tests/HasPermissionsTest.php
AlexVanderbist: Hi! As discussed over email; I definitely agree on not throwing the exception. It's been bugging me for a while but I wasn't sure about removing it as that would probably be a breaking change. 

I'm not sure if this would even be considered a feature. I can't imagine anyone checking if a permission exists based on that exception being thrown... If it's not a feature we can tag it as a patch.

drbyte:
@AlexVanderbist and I are a bit concerned that this may introduce breaking changes for a few sites.
Thus will likely tag it for a major release. 

Leaving this open and will direct it to another branch before merging.

Also still contemplating naming for a few things.

drbyte:
@Te7a-Houdini as of v2.24.0 the `hasAnyPermission()` calls the new `checkPermissionTo()` which returns a boolean.

Given that a boolean can be returned from `checkPermissionTo()`, do we still need this PR?

Te7a-Houdini:
@drbyte i think no we don't need this PR anymore , unless we need to remove the behavior of throwing exceptions from ```hasPermissionTo``` from the package , other wise the ```alias solution``` seems good and supports backward compatibility without breaking changes.

drbyte:
Great! 
Thanks for all your work on this PR. It took us down the right road of considering more carefully how we handle these requests. In our next major version we'll revisit where we throw exceptions more thoroughly.
Thanks again!

PR #820, opened 2018-08-08
add type checking in hasPermissionTo and hasDirectPermission

### LINKED ISSUE #764: hasPermissionTo() and hasDirectPermission() with inexpected argument types
Hello,
First of all: thank you for this so very useful package!

I think there is a potential problem in some corner use cases with the hasPermissionTo() and hasDirectPermission() methods:

https://github.com/spatie/laravel-permission/blob/f4763cc8957c2ce04efe4a84dbe7ad9d45c8cb39/src/Traits/HasPermissions.php#L106-L120

https://github.com/spatie/laravel-permission/blob/f4763cc8957c2ce04efe4a84dbe7ad9d45c8cb39/src/Traits/HasPermissions.php#L163-L180

specifically the  hasPermissionTo() performs no checks to ensure the $permission argument is no other type than an instance of \Spatie\Permission\Contracts\Permission, string or int (btw, the int option has been omitted in the PhpDoc block).  If it is anything other than an int or a string, it is passed over to the hasDirectPermission() method where again it is checked only if being a string or an int and only presuming that it is an instance of \Spatie\Permission\Contracts\Permission otherwise, but it could be anything else. If it is null, it would not have an id attribute to be called in the return statement and if it is an instance of another class type having an id attribute, it would yield an irrelevent id and difficult to debug weird behaviour. Also, it seems the intention is to return false in case a name or an id has not been found in the permission records, however the findByName() and findById() methods throw exceptions in this case (and the hasPermissionTo() method counts on that as it does not perform the checks in case name or id are not found).

Thanks for your consideration and again - thanks for the great work you've been doing with the package!

Jasper9041:
this pr fixes #764

I corrected the missing int parameter in the docblocks and added type checking in hasPermissionTo and hasDirectPermission

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -2,7 +2,6 @@
 
 namespace Spatie\Permission\Traits;
 
-use Spatie\Permission\Exceptions\PermissionDoesNotExist;
 use Spatie\Permission\Guard;
 use Illuminate\Support\Collection;
 use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks!

PR #823, opened 2018-08-10
Command-line added for role assignment and giving permission

ikidnapmyself:


drbyte:
@ikidnapmyself ,
Thanks for your work on this.
Would you mind adding some tests as well?

ananelle:
remove

drbyte:
Hmmmm ... while working on a way to add tests that could reasonably allow for app-specific User contracts, it became clear that this PR is brittle in that it makes assumptions about the app that we probably don't want to do in this package. (ie: assumes \App\User is the one and only model type and namespace that is ever used)

As such I'm going to close this PR. 

Nevertheless, thanks for your contribution - I know it takes time to submit something. It is still appreciated even though it isn't a good fit for this package's objectives. 
Thanks!

ikidnapmyself:
@drbyte thanks for your attention. You are totally right to do it. It was my mistake to request. I am seeing that User model may does not exist in a project. So I had to request model's name in command-line which will be ugly and pretty long to use for cli.

If I ever have a better thing to add / contribute, I will be more careful.

Cheers!

PR #827, opened 2018-08-13
Using another key name than 'model_id' - Fixes #777

### LINKED ISSUE #777: Using another key name than 'model_id'
In our app, we use UUIDs as ids.

For consistency within the application, we suffix the foreign key names with _uuid instead of _id as our primary key column is 'uuid'.

We can easily update the migration files to change the `$table->morphsTo('model')` to use a different name and type for the schema.

However, it it more difficult to let the `HasRoles` and `HasPermissions` traits know about it.

My first attempt was to redefine just the relationship method inside a custom trait :  

```
namespace App\Models\Traits;

use Illuminate\Database\Eloquent\Relations\MorphToMany;

trait HasRoles
{
    use \Spatie\Permission\Traits\HasRoles;

    /**
     * A model may have multiple roles.
     */
    public function roles(): MorphToMany
    {
        return $this->morphToMany(
            config('permission.models.role'),
            'model',
            config('permission.table_names.model_has_roles'),
            'model_uuid',
            'role_id'
        );
    }
}
```

And same kind of thing for `HasPermissions`

However, the `HasRoles` trait includes your own `HasPermissions` trait and thus it does not take into account my customized version. So, currently I am stuck with either : 

- Redefining the whole trait (not so great as I would copy-paste all methods including the ones which do not show any issues)
- Sticking with `model_id` as column name. But does not feel right/consistent when looking at the whole application DB schema.

Possible ways to solve it (unless you have a better idea) : 

- Allow a custom column name within the traits (could be a property which could be set in the class, like `private $permissionModelForeignColumn = 'model_uuid';` and check if that propery is set in the relationship method (defaulting to current name if not)
- Do not include `HasPermissions in your `HasRoles` trait and let the user include both traits in their models
- Have a configuration variable for that column name

Option 1 seems the nicest to me. Would you accept a PR for that?

vpratfr:
Using another key name than 'model_id' - Fixes #777

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,6 @@ return [
          * For example, this would be nice if your primary keys are all UUIDs. In 
          * that case, name this `model_uuid`.
          */
-        'model_morph_type' => 'model_type',
         'model_morph_key' => 'model_uuid',
     ],
 
--- a/README.md
+++ b/README.md
@@ -88,9 +88,8 @@ If you're using UUIDs or GUIDs for your `User` models you can update the `create
 
 ```php
 $table->uuid(config('permission.column_names.model_morph_key'));
-$table->string(config('permission.column_names.model_morph_type'));
-$table->index([config('permission.column_names.model_morph_key'), 
-               config('permission.column_names.model_morph_type'), ]);
+$table->string('model_type');
+$table->index([config('permission.column_names.model_morph_key'), 'model_type', ]);
 ```
 
 For consistency, you can also update the package configuration file to use the `model_uuid` column name instead of the default `model_id` column.
--- a/README.md
+++ b/README.md
@@ -87,9 +87,13 @@ php artisan vendor:publish --provider="Spatie\Permission\PermissionServiceProvid
 If you're using UUIDs or GUIDs for your `User` models you can update the `create_permission_tables.php` migration and replace `$table->morphs('model')` with:
 
 ```php
-$table->uuid('model_id');
-$table->string('model_type');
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -87,10 +87,14 @@ php artisan vendor:publish --provider="Spatie\Permission\PermissionServiceProvid
 If you're using UUIDs or GUIDs for your `User` models you can update the `create_permission_tables.php` migration and replace `$table->morphs('model')` with:
 
 ```php
-$table->uuid('model_id');
-$table->string('model_type');
+$table->uuid(config('permission.column_names.model_morph_key'));
+$table->string(config('permission.column_names.model_morph_type'));
```
drbyte: Did you mean to also include `model_morph_type` in the `config` file and in the migration? 
The README example here will get nulls if a fallback isn't provided and if it's not defined in the config file.

drbyte: Great timing :)  I see you addressed this in da20346
https://github.com/spatie/laravel-permission/pull/827/commits/da2034676fe954c7407440e0cebb83090a034ebd

vpratfr: No, sorry, that was an attempt to make the other column configurable too, but Laravel will not allow a custom type column name on MorphToMany relationships.

Hence I reverted the change in the code. But forgot to do it in the docs.

drbyte: Er, well, partially. The unused config key is still mentioned here in the readme.

Evidently you're actively looking at it ... so I'll review it a bit later ;)

vpratfr:
README should be fine now

Issue #808, opened 2018-07-29
assignRole - Duplicate entry


Nowi5:
Hi,

when I use assignRole([..]) with multiple roles I get an MySQL error because the user has already one of the roles. Can we please surround that with a nice check? Splitting this up in a loop to run hasRole() in front is not the best, as I would need to replicate in a lot of Controller.

Thank you.

Nowi5:
```
$roles = $request->input('roles') ? $request->input('roles') : [];

 foreach($roles as $role){
          if(!$user->hasRole($role){
            $user->assignRole($role);
          }
        }
```

warlord0:
Why not use syncRoles if you're passing roles and not just a role? No need for duplicate checking then.

Nowi5:
Because syncRoles would delete all other roles. To use syncRoles I would need to read all roles, put it into an array, add my new role (if not already exists) and than use syncRoles. But more important, a php function should never let a mysql dump unhandled.

PR #830, opened 2018-08-15
Make duplicate role assignment not throw an exception

elghobaty:
Closes https://github.com/spatie/laravel-permission/issues/808

Assigning roles now uses `sync` instead of `saveMany` with `detaching=false`. This way when assigning roles, non-existing roles will be added (hence, no SQL Exception) without affecting already assigned roles.


drbyte:
Thanks!

Nowi5:
Thank you very much for solving #808

elghobaty:
You are more than welcome :)

PR #867, opened 2018-09-10
Adds ability to pass in id's or mixed values to role scope.

surgiie:
Currently you can use the scopeRole to query by role names:

```php
 // Returns only users with the roles 'writer,editor,...
$users = User::role(['writer', 'editor', ...])->get();'

```

However it be useful in some scenarios to be able to use the scope to query by id's 

```php
// Returns only users with the roles with the given id's
$users = User::role([1,4,6, ...])->get(); 

```

This pull request adds that ability by checking whether the role being mapped over in the scopeRole method is numeric or not then calls either `findById` or `findByName` :

```php

 $method = is_numeric($role) ? 'findById' : 'findByName';

```

Might be a trivial use case, but this would also allow you to call by mixed values

```php
User::role([1, $role, 'writer'])->get();
```

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -71,7 +71,9 @@ trait HasRoles
                 return $role;
             }
 
-            return $this->getRoleClass()->findByName($role, $this->getDefaultGuardName());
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks!

Issue #870, opened 2018-09-11
Integrity constraint violation: 1048 Column 'model_id' cannot be null


musman92:
my package details are 
"laravel/framework": "5.6.*",
"spatie/laravel-permission": "^2.19"

when i assign a role 
$user->syncRoles(['role']);
$user->assignRole('role');

 Illuminate\Database\QueryException  : SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'model_id' cannot be null (SQL: insert into `model_has_roles` (`model_id`, `model_type`, `role_id`) values (, App\User, 2))

thanks in advance

drbyte:
In your Eloquent model for `App\User`, what is the column-type for your `id` field? 

This package expects it to match the `model_id` type of the `model_has_roles` table.

musman92:
`id` INT(10) UNSIGNED NOT NULL AUTO_INCREMENT,
how I match model_id?

drbyte:
What is the value for `model_morph_key` in your `config/permission.php`?

https://github.com/spatie/laravel-permission/blob/master/config/permission.php#L83

musman92:
the value is 
'model_morph_key' => 'model_id',

drbyte:
No, it should be model_id.

This is puzzling. I think you've changed something in your `App\User` model that's causing it.
It's hard to troubleshoot this without seeing your code.

musman92:
below is the my user model`<?php

namespace App;

use Illuminate\Notifications\Notifiable;
use Illuminate\Foundation\Auth\User as Authenticatable;
use Spatie\Permission\Traits\HasRoles;

class User extends Authenticatable
{
  use Notifiable, HasRoles;

  /**
   * The attributes that are mass assignable.
   *
   * @var array
   */
  protected $fillable = [
    'name', 'email', 'password',
  ];

  /**
   * The attributes that should be hidden for arrays.
   *
   * @var array
   */
  protected $hidden = [
    'password', 'remember_token',
  ];

}
`

drbyte:
Can you share the code where you're trying to assign and sync roles to users?

drbyte:
You can't sync something to a model that hasn't been persisted yet.

You must call `$user->save()` before you sync or assign.

musman92:
@drbyte  thank you it works. but it should send a proper message. thank you anyway

seasoftnaymyohtet:
> _hhhdfhdndjcjkdsaj****_

[...]

PR #874, opened 2018-09-14
Prevent QueryException when syncing to models that are not persisted

drbyte:
Fixes Integrity Constraint Violation error as reported in #870

EDIT: Registers a `saved` callback to sync after save.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to `laravel-permission` will be documented in this file
 
 
 ## 2.20.0 - 2018-09-19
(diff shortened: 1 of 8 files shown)


drbyte:
@freekmurze @AlexVanderbist 
Are you okay with this just ignoring the sync? Or prefer to create and throw another exception which explains that the sync hasn't been done?

freekmurze:
I won't just ignore it. People might think they've just attached a permission/role, while in fact it has not been attached.

What you could do is, if the model doesn't exist yet, register an `saved` binding that will sync the permission/roles as soon as the model is saved. Here's some code from another package where we do this: https://github.com/spatie/nova-tags-field/blob/df74a82748acd6b905b59cca404b2b3e4ea61324/src/Tags.php#L73-L75

PR #899, opened 2018-10-10
Add role & permission middleware helper functions to Route class

eemrekaya:
Added `role` and `permission` middleware helper functions to `Route` class from Laravel 5.5 and higher

Example:
```php
Route::get('users', 'UserController@index')->role('superadmin');
```

```php
Route::get('users', 'UserController@index')->permission('show users list');
```

```php
Route::get('admin/users', 'UserController@index')->role(['superadmin', 'admin']->permission(['show users list', 'another permission']);
```


drbyte:
Thanks!

PR #902, opened 2018-10-11
Add role_or_permission middleware

josepostiga:
Hi guys! Nice package you've got here.

This PR adds a new middleware to support the verification that a user has either a role OR a given permission. It's most useful for including a verification of a high access level user and lower access level users, without having to add any specific permissions to the high access level user.

Take, for example, the following:

```php
public function __construct()
{
    $this->middleware('role_or_permission:super-admin|edit-news');
}
``` 

This expression would pass if the user has either the role `super-admin` or the permission `edit-news` (whatever the role has that permission).

Wish you all a good day, 
José Postiga

### DIFF
--- a/README.md
+++ b/README.md
@@ -645,7 +645,7 @@ You can use all of the blade directives listed in [using blade directives](#usin
 
 ## Using a middleware
 
-This package comes with `RoleMiddleware`, `PermissionMiddleware` nad `RoleOrPermissionMiddleware` middleware. You can add them inside your `app/Http/Kernel.php` file.
+This package comes with `RoleMiddleware`, `PermissionMiddleware` and `RoleOrPermissionMiddleware` middleware. You can add them inside your `app/Http/Kernel.php` file.
 
 ```php
 protected $routeMiddleware = [
--- a/README.md
+++ b/README.md
@@ -671,11 +671,11 @@ Route::group(['middleware' => ['role:super-admin','permission:publish articles']
     //
 });
 
-Route::group(['middleware' => ['role_or_permission:super-admin]], function () {
+Route::group(['middleware' => ['role_or_permission:super-admin']], function () {
     //
 });
 
-Route::group(['middleware' => ['role_or_permission:publish articles]], function () {
+Route::group(['middleware' => ['role_or_permission:publish articles']], function () {
     //
 });
 ```
--- a/README.md
+++ b/README.md
@@ -645,13 +645,14 @@ You can use all of the blade directives listed in [using blade directives](#usin
 
 ## Using a middleware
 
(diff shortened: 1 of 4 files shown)


drbyte:
Thanks!

ujackson:
Thanks @josepostiga for the update. Can I check multiple roles or permissions? Doesn't seem to work. 
`    public function __construct()
{
    $this->middleware(['role_or_permission:admin|doctor|create patients|delete patients|update patients|read patients']);
}`

bart-heideman:
Hi ujackson,

I think you should use a comma to separate the roles and or permissions if they are multiple.

josepostiga:
@ujackson can you confirm that all those permissions (not the roles) exist on your database and that you're including the `RoleOrPermissionMiddleware` on your `http/kernel.php` file?

I tested with many roles and permissions (although the tests only have one role and/or one permission, but I can update that if needed) and I concluded that the only way that won't work is if any of those permissions don't exist on the database, and that's because of the behaviour described on #802.

drbyte:
@josepostiga When you worked on this, did you intentionally use `||` in the middleware test, instead of `&&` ?  
I'm thinking that if the meaning of "role or permission" means "any of the passed roles or permissions is required", then the `does not have any of the roles OR does not have any of the permissions` logic should actually be `AND` in order to trigger the `UnauthorizedException`. 
Or am I missing something here?

josepostiga:
@drbyte I was actually looking at this when you pinged me. I did confuse myself on the operator used and your understanding is correct: it should, indeed, be a `&&`.

However, when I tried to update the middleware to use the correct operator, I encounter the problem described on #802, via the `hasAnyPermission`. Probably gonna need to work around that while #814 is not merged.

drbyte:
Ya, there are mixed scenarios: if ALL roles AND ALL permissions are NOT matched, then the OR operator properly gives a 403.  But an AND operator lets it pass with a 200. :(

josepostiga:
Yeah and that's because of the (dirty) try-catch block I added to the middleware, to avoid getting the `PermissionDoesNotExist` exception.

I'm actually working on a solution, right now.

drbyte:
FWIW I'm not bothered by the use of the try-catch of `PermissionDoesNotExist` in the interim until #814 goes in.

josepostiga:
I thought it was pretty clever at the time but then I tried to use this middleware on a project I'm working on and it breaks the logic if a permission is not yet created when being checked. It's a very edgy case, but it can happen.

I think that a good way of working around the issue, while #814 is not merged, is to create an alias of `hasPermissionTo()` where the try-catch will be, ensuring that the return is only true/false.

I'm about to create a PR to handle this and we can discuss this further, if needed, there. It's really a small change to the `HasPermissions` trait that I think keeps things under control, for now...

Issue #886, opened 2018-10-01
Unnecessary requests when a user have no role at all


lk77:
Hello,

On one of our interfaces we don't have the use of roles for some of the users,
and it seems that  the packet is still doing a number of requests (35 in my case)
that are totally useless because the absence of roles.

The problem reside in the getAllPermissions() and getPermissionsViaRoles(),
the permissions of the user and the permissions of the different roles the user have, are being merged.
But when the user doesn't have roles, it should stop querying them i think.

Exemple of requests:
```
select `roles`.*, `model_has_roles`.`model_id` as `pivot_model_id`, `model_has_roles`.`role_id` as `pivot_role_id`, `model_has_roles`.`model_type` as `pivot_model_type` from `roles` inner join `model_has_roles` on `roles`.`id` = `model_has_roles`.`role_id` where `model_has_roles`.`model_id` in ('3') and `model_has_roles`.`model_type` = 'App\Models\User'
```

perhaps counting the role before trying to recover them,
i don't have a solution in mind,

tell me what you think.

thanks.

drbyte:
Do you envision something like this? 

```diff
    public function getPermissionsViaRoles(): Collection
    {
+        $roles = $this->load('roles', 'roles.permissions')->roles;
+
+        if (! $roles) {
+            return collect();
+        }
+
-        return $this->load('roles', 'roles.permissions')
-            ->roles->flatMap(function ($role) {
+        return $roles->flatMap(function ($role) {
                return $role->permissions;
            })->sort()->values();
    }
```

lk77:
I don't know,
the problem for me is the use of load,
it's reloading the roles each time a permission is checked,
something like that will keep the result for further checks :
```
/**
     * Return all the permissions the model has via roles.
     */
    public function getPermissionsViaRoles(): Collection
    {
        $roles = $this->roles;

        if (! $roles) {
                    return collect();
        }

        return $roles->load('roles', 'roles.permissions')->flatMap(function ($role) {
            return $role->permissions;
        })->sort()->values();
    }
```

josepostiga:
@drbyte this issue is still open. Is there anyting missing?

drbyte:
No, it should have been closed.

PR #903, opened 2018-10-13
Avoid unnecessary queries of user roles when fetching all permissions

josepostiga:
Hi, everyone!

This PR addresses #886 by adding a condition to only call `getPermissionsViaRoles()`, via `getAllPermisisons()`, when a `User` has roles.


drbyte:
Thanks!

Closes #886

Issue #904, opened 2018-10-14
Lumen: Class 'Illuminate\Routing\Route' not found


moeen-basra:
After upgrading package for lumen 5.7 I'm facing this problem.

I think there is no support for macros in lumen router and Facade for router you are trying to use its for laravel not for lumen.

urukalo:
same for lumen 5.6

stefanvanherwijnen:
I am having the same issue with Lumen 5.6 and laravel-permission 2.22.0. Version 2.17.0 works fine.
Probably related to this commit:
https://github.com/spatie/laravel-permission/commit/69fba17fe864da9cb2e9b03e7d97bb7fced75595

PR #905, opened 2018-10-14
#904 call registerMacroHelpers inside not lumen condtion

moeen-basra:
This will register macros only in case of laravel


drbyte:
Thanks!

Closes #904

PR #916, opened 2018-10-29
Refresh model roles after assigning a role

### LINKED ISSUE #914: Role relation does not get updated after assigning role
consider the following example:

```
public function testRoles()
    {
        $user = User::first();
        $this->assertFalse($user->hasRole('admin'));
        $user->assignRole('admin');
        $this->assertTrue($user->hasRole('admin'));
    }
```

The second assertion fails. I'm not sure if there's something wrong with the laravel internals or this package didn't go in depth yet. But atleast i fixed it for now by overriding and extending the hasRoles trait and adding

`unset($this->roles);`

to the end of the assignrole method.

What could be causing this?

hulkur:
Fixes #914 

There is some side effect with tests. 
`it_can_assign_and_remove_a_role` will pass if there is no assert in the beginning and fail with it (in original code).

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -116,6 +116,7 @@ trait HasRoles
 
         if ($model->exists) {
             $this->roles()->sync($roles, false);
+            $model->load('roles');
         } else {
             $class = \get_class($model);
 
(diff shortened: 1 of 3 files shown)


deviouspk:
Can you do the same for removing roles & assigning/removing permissions please

drbyte:
Thanks!

drbyte:
Note: with further testing I've discovered that we probably should have also done the reload on "remove" operations as well, for parity. But that then makes some middleware tests fail.

PR #918, opened 2018-11-04
Avoid empty permission creation

bpeynet:
Null permissions string resulted in the creation of an empty permission
associated to the new created role.
New line added to the CommandTest file to test the correction.


drbyte:
Thanks!
I've updated it slightly to simplify the code by using an early return.

PR #921, opened 2018-11-07
Fix operator used on RoleOrPermissionMiddleware

josepostiga:
This PR fixes the logic operation inside the `RoleOrPermissionMiddleware`, where a `||` is used, instead of a `&&`.

It also adds an alias for `hasPermissionTo()`, `checkPermissionTo`, which catches any `PermissionDoesNotExist` exception and returning `false`, ensuring the return value to always be a boolean.


drbyte:
Thanks.
What's the intended outcome if ALL roles-and-permissions passed to this middleware don't actually exist?

josepostiga:
It should return a 403.

PR #926, opened 2018-11-08
Add model/permissions caching

### LINKED ISSUE #759: Needs an in memory caching strategy
I think this package needs a memory cache because everything is unserialized from external cache every time `hasPermissionTo` is called.

I have a system with 100+ permissions and roles (the cache file is 4.6mb) and a view with 75+ `@can` directives, the page was taking more than 20 seconds to load.

A simple solution like this got it down to 2 seconds.

```   
// PermissionRegistrar.php

    private $permissions;

    public function getPermissions(): Collection
    {
        if (!$this->permissions) {
            $this->permissions = $this->cache->remember($this->cacheKey, config('permission.cache_expiration_time'), function () {
                return app(Permission::class)->with('roles')->get();
            });
        }
        return $this->permissions;
    }
```

`PermissionRegistrar` is also being created every time, it should be registered as a singleton
`$this->app->singleton(PermissionRegistrar::class);`

Sorry for just throwing it here but I dont have the time for a PR right now

### LINKED ISSUE #789: 200+ permissions slowing down the website :(
Hi,

I have more than 200 permissions and SaaS solution. with only 50+ users, my permission_role table has already crossed 4000 entries. Basically in my webapp, every user can create his/her own roles then assign any of these 200 permissions. 

I applied solution mentioned in #759 but it didn't reduce the memory usage ( at the moment more than 70mb :( as per laravel debugger )

It is slowing down my entire web-app and causing lot of other issues. Any solutions or suggestions? :(
I don't mind working with freelancer to get this fixed but I really need some solution for this. Page should not take more than 2-3 seconds to load. 

Thanks.

fullstackfool:
This adds substantial speed increases by caching the associations between models and permissions. Part of the changes focus on the `PermissionRegistar`, specifically caching for `getPermissions()`, thanks to @klincheg for these changes.

The rest is focussed on caching calls to `HasPermissionTo()` in `HasPermissions.php`. 

The association is cached under a compound key comprised of the package key, the model and the permission:
`spatie.permission.cache.App.Models.User.12.Spatie.Permission.Models.Permissions.view-dashboard`

These values are also used as tags: 
`spatie.permission.cache`, 
`App.Models.User.12`, 
`Spatie.Permission.Models.Permissions.view-dashboard`
(if tags are not supported then the package continues as before, as they are essential to this functionality)

As we are still using the package cache key, then the already existing cache-busting should not need to be updated to account for these changes. However, the usage of the model and the permissions in the tags allow for future addition of model specific, or permission specific cache-busting, helping prevent the stampeding herd effect that larger applications may experience each time a permission or role is updated.

Performance tests involved seeding a database with 250 users, who were assigned between 2 - 4 roles, which were assigned between 30 - 500 permissions. Then each permission for each user was tested:
```php
$this->users->each(function (User $user) {
    $user->getAllPermissions()->each(function (Permission $permission) use ($user) {
        $this->assertTrue($user->can($permission->name));
    });
});
``` 

I went through different variations of the above, some including non-assigned permissions, etc. 

These changes increase the speed of the checks from several minutes down to 4 - 20 seconds (approx 32000 assertions in a couple of seconds in some tests). And the database calls went from 1800+ to 0.

Things to note/get feedback on are:
 - Is the current cache busting effective enough for these changes? I reviewed the tests and ran through some real-world examples and had no issues, but a second opinion would be great.
 - The changes to the config could be breaking
 - The naming of `checkHasPermissionTo()` inside `HasPermissions` could be confused with the newly added method `checkPermissionTo()`
 - Clutter. I'm concious this has added a fair bit of new methods and checks, and I don't want to clutter up the package.

Closes #732 
Fixes #759 
Fixes #789

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -3,6 +3,7 @@
 namespace Spatie\Permission;
 
 use Illuminate\Support\Collection;
+use Spatie\Permission\Contracts\Role;
 use Illuminate\Contracts\Auth\Access\Gate;
 use Illuminate\Contracts\Cache\Repository;
 use Spatie\Permission\Contracts\Permission;
@@ -18,14 +19,49 @@ class PermissionRegistrar
     protected $cache;
 
     /** @var string */
-    protected $cacheKey = 'spatie.permission.cache';
+    protected $permissionClass;
 
+    /** @var string */
+    protected $roleClass;
+
+    /** @var int */
+    public static $cacheExpirationTime;
+
+    /** @var string */
+    public static $cacheKey;
+
+    /** @var string */
+    public static $cacheModelKey;
+
+    /** @var bool */
+    public static $cacheIsTaggable = false;
+
+    /**
+     * PermissionRegistrar constructor.
+     *
+     * @param \Illuminate\Contracts\Auth\Access\Gate $gate
+     * @param \Illuminate\Contracts\Cache\Repository $cache
+     */
     public function __construct(Gate $gate, Repository $cache)
     {
         $this->gate = $gate;
-        $this->cache = $cache;
+        $this->permissionClass = config('permission.models.permission');
+        $this->roleClass = config('permission.models.role');
+
+        self::$cacheExpirationTime = config('permission.cache.expiration_time',
+            config('permission.cache_expiration_time'));
+        self::$cacheKey = config('permission.cache.key');
+        self::$cacheModelKey = config('permission.cache.model_key');
+        self::$cacheIsTaggable = ($cache->getStore() instanceof \Illuminate\Cache\TaggableStore);
+
+        $this->cache = self::$cacheIsTaggable ? $cache->tags(self::$cacheKey) : $cache;
     }
 
+    /**
+     * Register the permission check method on the gate.
+     *
+     * @return bool
+     */
     public function registerPermissions(): bool
     {
         $this->gate->before(function (Authorizable $user, string $ability) {
@@ -40,19 +76,75 @@ class PermissionRegistrar
         return true;
     }
 
+    /**
+     * Flush the cache.
+     */
     public function forgetCachedPermissions()
     {
-        $this->cache->tags($this->cacheKey)->flush();
+        self::$cacheIsTaggable ? $this->cache->flush() : $this->cache->forget(self::$cacheKey);
     }
 
-    public function getPermissions($params = null): Collection
+    /**
+     * Get the permissions based on the passed params.
+     *
+     * @param array $params
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    public function getPermissions(array $params = []): Collection
     {
-        return $this->cache->tags($this->cacheKey)->remember($this->cacheKey.($params ? '.'.implode('.', array_values($params)) : ''), config('permission.cache_expiration_time'), function () use ($params) {
-            if ($params) {
-                return app(Permission::class)->where($params)->with('roles')->get();
-            } else {
-                return app(Permission::class)->with('roles')->get();
+        $permissions = $this->cache->remember($this->getKey($params), self::$cacheExpirationTime,
+            function () use ($params) {
+                return $this->getPermissionClass()
+                    ->when(($params && self::$cacheIsTaggable), function ($query) use ($params) {
+                        return $query->where($params);
+                    })
+                    ->with('roles')
+                    ->get();
+            });
+
+        if (! self::$cacheIsTaggable) {
+            foreach ($params as $attr => $value) {
+                $permissions = $permissions->where($attr, $value);
             }
-        });
+        }
+
+        return $permissions;
+    }
+
+    /**
+     * Get the key for caching.
+     *
+     * @param $params
+     *
+     * @return string
+     */
+    public function getKey(array $params): string
+    {
+        if ($params && self::$cacheIsTaggable) {
+            return self::$cacheKey.'.'.implode('.', array_values($params));
+        }
+
+        return self::$cacheKey;
+    }
+
+    /**
+     * Get an instance of the permission class.
+     *
+     * @return \Spatie\Permission\Contracts\Permission
+     */
+    public function getPermissionClass(): Permission
+    {
+        return app($this->permissionClass);
+    }
+
+    /**
+     * Get an instance of the role class.
+     *
+     * @return \Spatie\Permission\Contracts\Role
+     */
+    public function getRoleClass(): Role
+    {
+        return app($this->roleClass);
     }
 }
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -42,13 +42,17 @@ class PermissionRegistrar
 
     public function forgetCachedPermissions()
     {
-        $this->cache->forget($this->cacheKey);
+        $this->cache->tags($this->cacheKey)->flush();
     }
 
-    public function getPermissions(): Collection
+    public function getPermissions($params = null): Collection
     {
-        return $this->cache->remember($this->cacheKey, config('permission.cache_expiration_time'), function () {
-            return app(Permission::class)->with('roles')->get();
+        return $this->cache->tags($this->cacheKey)->remember($this->cacheKey.($params ? '.'.implode('.', array_values($params)) : ''), config('permission.cache_expiration_time'), function () use ($params) {
+            if ($params) {
+                return app(Permission::class)->where($params)->with('roles')->get();
+            } else {
+                return app(Permission::class)->with('roles')->get();
+            }
         });
     }
 }
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -9,6 +9,7 @@ use Spatie\Permission\PermissionRegistrar;
 use Spatie\Permission\Contracts\Permission;
 use Spatie\Permission\Exceptions\GuardDoesNotMatch;
 use Illuminate\Database\Eloquent\Relations\MorphToMany;
+use Spatie\Permission\Exceptions\PermissionDoesNotExist;
 
 trait HasPermissions
 {
@@ -21,6 +22,15 @@ trait HasPermissions
 
             $model->permissions()->detach();
         });
+    }
+
+    public function getPermissionClass()
+    {
+        if (! isset($this->permissionClass)) {
+            $this->permissionClass = app(PermissionRegistrar::class)->getPermissionClass();
+        }
+
+        return $this->permissionClass;
     }
 
     /**
@@ -98,13 +108,45 @@ trait HasPermissions
     /**
      * Determine if the model may perform the given permission.
      *
-     * @param string|\Spatie\Permission\Contracts\Permission $permission
+     * @param string|int|\Spatie\Permission\Contracts\Permission $permission
+     * @param string|null $guardName
+     *
+     * @return bool
+     * @throws \Exception
+     */
+    public function hasPermissionTo($permission, $guardName = null): bool
+    {
+        if (! is_string($permission) && ! is_int($permission) && ! $permission instanceof Permission) {
+            throw new PermissionDoesNotExist;
+        }
+
+        if (! PermissionRegistrar::$cacheIsTaggable) {
+            return $this->hasUncachedPermissionTo($permission, $guardName);
+        }
+
+        return cache()
+            ->tags($this->getCacheTags($permission))
+            ->remember(
+                $this->getCacheKey($permission),
+                PermissionRegistrar::$cacheExpirationTime,
+                function () use ($permission, $guardName) {
+                    return $this->hasUncachedPermissionTo($permission, $guardName);
+                }
+            );
+    }
+
+    /**
+     * Check the uncached permissions for the model.
+     *
+     * @param string|int|Permission $permission
      * @param string|null $guardName
      *
      * @return bool
      */
-    public function hasPermissionTo($permission, $guardName = null): bool
+    public function hasUncachedPermissionTo($permission, $guardName = null): bool
     {
+        $permissionClass = $this->getPermissionClass();
+
         if (is_string($permission)) {
             $permission = app(Permission::class)->findByName(
                 $permission,
@@ -113,10 +155,101 @@ trait HasPermissions
         }
 
         if (is_int($permission)) {
-            $permission = app(Permission::class)->findById($permission, $this->getDefaultGuardName());
+            $permission = $permissionClass->findById(
+                $permission,
+                $guardName ?? $this->getDefaultGuardName()
+            );
+        }
+
+        if (! $permission instanceof Permission) {
+            throw new PermissionDoesNotExist;
         }
 
         return $this->hasDirectPermission($permission) || $this->hasPermissionViaRole($permission);
+    }
+
+    /**
+     * An alias to hasPermissionTo(), but avoids throwing an exception.
+     *
+     * @param string|int|\Spatie\Permission\Contracts\Permission $permission
+     * @param string|null $guardName
+     *
+     * @return bool
+     *
+     * @throws \Exception
+     */
+    public function checkPermissionTo($permission, $guardName = null): bool
+    {
+        try {
+            return $this->hasPermissionTo($permission, $guardName);
+        } catch (PermissionDoesNotExist $e) {
+            return false;
+        }
+    }
+
+    /**
+     * Construct the key for the cache entry.
+     *
+     * @param null|string|int|\Spatie\Permission\Contracts\Permission $permission
+     *
+     * @return string
+     */
+    protected function getCacheKey($permission = null)
+    {
+        $key = PermissionRegistrar::$cacheKey.'.'.$this->getClassCacheString();
+
+        if ($permission !== null) {
+            $key .= $this->getPermissionCacheString($permission);
+        }
+
+        return $key;
+    }
+
+    /**
+     * Construct the tags for the cache entry.
+     *
+     * @param null|string|int|\Spatie\Permission\Contracts\Permission $permission
+     *
+     * @return array
+     */
+    protected function getCacheTags($permission = null)
+    {
+        $tags = [
+            PermissionRegistrar::$cacheKey,
+            $this->getClassCacheString(),
+        ];
+
+        if ($permission !== null) {
+            $tags[] = $this->getPermissionCacheString($permission);
+        }
+
+        return $tags;
+    }
+
+    /**
+     * Get the key to cache the model by.
+     *
+     * @return string
+     */
+    private function getClassCacheString()
+    {
+        return str_replace('\\', '.', get_class($this)).'.'.$this->getKey();
+    }
+
+    /**
+     * Get the key to cache the permission by.
+     *
+     * @param string|int|\Spatie\Permission\Contracts\Permission $permission
+     *
+     * @return mixed
+     */
+    protected function getPermissionCacheString($permission)
+    {
+        if ($permission instanceof Permission) {
+            $permission = $permission[PermissionRegistrar::$cacheModelKey];
+        }
+
+        return str_replace('\\', '.', Permission::class).'.'.$permission;
     }
 
     /**
@@ -125,6 +258,7 @@ trait HasPermissions
      * @param array ...$permissions
      *
      * @return bool
+     * @throws \Exception
      */
     public function hasAnyPermission(...$permissions): bool
     {
@@ -139,6 +273,29 @@ trait HasPermissions
         }
 
         return false;
+    }
+
+    /**
+     * Determine if the model has all of the given permissions.
+     *
+     * @param array ...$permissions
+     *
+     * @return bool
+     * @throws \Exception
+     */
+    public function hasAllPermissions(...$permissions): bool
+    {
+        if (is_array($permissions[0])) {
+            $permissions = $permissions[0];
+        }
+
+        foreach ($permissions as $permission) {
+            if (! $this->hasPermissionTo($permission)) {
+                return false;
+            }
+        }
+
+        return true;
     }
 
     /**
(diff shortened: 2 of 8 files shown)


src/Traits/HasPermissions.php
```diff
@@ -52,7 +52,7 @@ public function permissions(): MorphToMany
     /**
      * Scope the model query to certain permissions only.
      *
-     * @param \Illuminate\Database\Eloquent\Builder $query
+     * @param \Illuminate\Database\Eloquent\Builder                                               $query
```
drbyte: While I too like having PHPStorm align variable names when reformatting, one thing it does poorly is this scenario when it shoves in 30+ spaces and actually creates more confusion instead of improving readability.
So in this package we opt to turn off that "alignment" setting. 
Please remove the extra spaces added in docblocks.

drbyte:
@fullstackfool Thanks for this.

I think your questions/areas requesting more feedback all have merit.
I'm debating whether to tag a new major version release for them. 

I want to test it a bit before making such decisions though.

src/Traits/HasPermissions.php
fullstackfool: Apologies - tried very hard not to press my auto-format shortcut while doing this, apparently, I failed.

drbyte: No worries - I'm always catching myself doing that too!

drbyte:
Additional questions that come to mind:
- Should there be more tests added to cover testing the tagging features?
- Should the duplication of code for caching be consolidated somehow?
- Should the `tag()` method be added into the PermissionRegistrar:getPermissions() method?

drbyte:
Thinking aloud: Should we make it assume a taggable cache being a first-class citizen? ie: refactor to stuff the less-wanted behavior inside the "if" ("if it **isn't** taggable, then do X")

fullstackfool:
> Should the duplication of code for caching be consolidated somehow?

Indeed. I've attempted to tidy up a little in `PermissionsRegistrar` - what do you think about the below?

```php
<?php

namespace Spatie\Permission;

use Illuminate\Contracts\Auth\Access\Authorizable;
use Illuminate\Contracts\Auth\Access\Gate;
use Illuminate\Contracts\Cache\Repository;
use Illuminate\Support\Collection;
use Spatie\Permission\Contracts\Permission;
use Spatie\Permission\Contracts\Role;
use Spatie\Permission\Exceptions\PermissionDoesNotExist;

class PermissionRegistrar
{
    /** @var \Illuminate\Contracts\Auth\Access\Gate */
    protected $gate;

    /** @var \Illuminate\Contracts\Cache\Repository */
    protected $cache;

    /** @var string */
    protected $permissionClass;

    /** @var string */
    protected $roleClass;

    /** @var string */
    protected $cacheKey;

    /** @var int */
    protected $cacheMinutes;

    /** @var bool */
    public static $cacheIsTaggable = false;


    /**
     * PermissionRegistrar constructor.
     *
     * @param \Illuminate\Contracts\Auth\Access\Gate $gate
     * @param \Illuminate\Contracts\Cache\Repository $cache
     */
    public function __construct(Gate $gate, Repository $cache)
    {
        $this->gate = $gate;
        $this->cacheKey = config('permission.cache.key');
        $this->roleClass = config('permission.models.role');
        $this->permissionClass = config('permission.models.permission');
        $this->cacheMinutes = config('permission.cache.expiration_time');
        $this->cache = self::$cacheIsTaggable ? $cache->tags(config('permission.cache.key')) : $cache;

        self::$cacheIsTaggable = ($cache->getStore() instanceof \Illuminate\Cache\TaggableStore);
    }

    /**
     * Register the permission check method on the gate.
     *
     * @return bool
     */
    public function registerPermissions(): bool
    {
        $this->gate->before(function (Authorizable $user, string $ability) {
            try {
                if (method_exists($user, 'hasPermissionTo')) {
                    return $user->hasPermissionTo($ability) ?: null;
                }
            } catch (PermissionDoesNotExist $e) {
            }
        });

        return true;
    }

    /**
     * Flush the cache.
     *
     * @return bool
     */
    public function forgetCachedPermissions(): bool
    {
        return self::$cacheIsTaggable ? $this->cache->flush() : $this->cache->forget($this->cacheKey);
    }

    /**
     * Get the permissions based on the passed params.
     *
     * @param array $params
     *
     * @return \Illuminate\Support\Collection
     */
    {
        $permissions = $this->cache->remember($this->getKey($params), $this->cacheMinutes, function () use ($params) {
            return self::$cacheIsTaggable ? $this->getTaggedPermissions($params) : $this->getUntaggedPermissions();
        });

        if (!self::$cacheIsTaggable) {
            foreach ($params as $attr => $value) {
                $permissions = $permissions->where($attr, $value);
            }
        }

        return
        $permissions;
    }

    /**
     * Get the permissions for a taggable cache.
     *
     * @param array $params
     *
     * @return \Illuminate\Support\Collection
     */
    public function getTaggedPermissions(array $params): Collection
    {
            ->when($params, function ($query) use ($params) {
                return $query->where($params);
            })
            ->with('roles')
            ->get();
    }

    /**
     * Get the permissions for a non-taggable cache.
     *
     * @return \Illuminate\Support\Collection
     */
    public function getUntaggedPermissions(): Collection
    {
        return $this->getPermissionClass()->with('roles')->get();
    }


    /**
     * Get the key for caching.
     * 
     * @param $params
     *
     * @return string
     */
    public function getKey(array $params): string
    {
        if ($params && self::$cacheIsTaggable) {
            return $this->cacheKey . '.' . implode('.', array_values($params));
        }

        return $this->cacheKey;
    }

    /**
     * Get and instance of the permission class.
     *
     * @return \Spatie\Permission\Contracts\Permission
     */
    public function getPermissionClass(): Permission
    {
        return app($this->permissionClass);
    }

    /**
     * Get and instance of the role class.
     *
     * @return \Spatie\Permission\Contracts\Role
     */
    public function getRoleClass(): Role
    {
        return app($this->roleClass);
    }
}
```

fullstackfool:
> getTaggedPermissions vs getUntaggedPermissions ... if $params is null, aren't they identical (ie: the when would be skipped, thus leaving them the same)? Couldn't there be just one version of this? It wouldn't need to be a separate method then either.
Well, that's embarrassing...

> In your revised forgetCachedPermissions(), you removed the ->tags() call in the flush. Was that intentional? is that because of your changed declaration of $this->cache in the constructor?
return self::$cacheIsTaggable ? $this->cache->flush() : $this->cache->forget($this->cacheKey);
Thats the idea - do you think it's too confusing a way of doing it?

drbyte:
Turns out we can't just check for `when($params)`. Needs to also test for taggable.
This seems to work better across various cache stores: 
`->when($params && self::$cacheIsTaggable, function ($query) use ($params) {`

For the tag() thing, I think what you're proposing is fine. I was just trying to follow the bouncing ball while comparing diffs. I don't think it's wrong (it works); it's just different than what you'd done before, so I wanted to be sure I followed your thinking.

drbyte:
> The naming of checkHasPermissionTo() inside HasPermissions could be confused with the newly added method checkPermissionTo()

I wonder if renaming the new `checkHasPermissionTo` to `uncachedHasPermissionTo` would suit. It's more explicit on its intent so doesn't require documenting.

fullstackfool:
> > The naming of checkHasPermissionTo() inside HasPermissions could be confused with the newly added method checkPermissionTo()
> 
> I wonder if renaming the new `checkHasPermissionTo` to `uncachedHasPermissionTo` would suit. It's more explicit on its intent so doesn't require documenting.

Forgot to show the changes for that, came up with `hasUncachedPermissionTo` - either will do I think 😄

drbyte:
> Is the current cache busting effective enough for these changes? I reviewed the tests and ran through some real-world examples and had no issues, but a second opinion would be great.

I *think* so. 

> The changes to the config could be breaking

Yes, they are. Especially since you're referencing the config values as array elements. If those are missing, there will be errors. The use of the array is speedier than calling the `config()` helper repeatedly, but it prevents the ability to declare any fallbacks.

> Clutter. I'm concious this has added a fair bit of new methods and checks, and I don't want to clutter up the package.

Ya, my eye is on the many added cache-related helper methods in the HasPermissions trait. I find myself itching for a separate class for cache handling, but wonder if they fit in the Registrar just as well. Unfortunately since they're all context-relevant to Permissions, they do also fit equally well there in the trait ... it's just that it's getting so big ...

fullstackfool:
> Ya, my eye is on the many added cache-related helper methods in the HasPermissions trait. I find myself itching for a separate class for cache handling, but wonder if they fit in the Registrar just as well. Unfortunately since they're all context-relevant to Permissions, they do also fit equally well there in the trait ... it's just that it's getting so big ...

It is. And it would get even bigger once we consider adding permission/model specific cache-busting methods. And there is also role/model caching, which could also be added. Ok, so maybe this all needs to be extracted then.

drbyte:
What if we downgrade slightly and undo the breaking change on the cache config file.
Then we could merge as-is without tagging a major version.

Then we could extract things along with the breaking changes?

drbyte:
A couple nitpicks with the posted code today:

Can't return bool here. Returns void.
`    public function forgetCachedPermissions(): bool`

"and" should be "an":
`* Get and instance of`

fullstackfool:
Cheers for the pointers, and the config suggestion is a good first step.

Signing off for the night, so I'll sort the changes tomorrow and if you're happy with the revised changes I'll pull them into this branch for the pr. 

Thanks for your help so far!

fullstackfool:
> Can't return bool here. Returns void.

`Illuminate\Contracts\Cache\Store` and `Illuminate\Contracts\Cache\Repository`, as well as all the cache drivers have the return type set as `bool` - are you sure we can't add the return type here?

drbyte:
>are you sure we can't add the return type here?

I got errors while executing, saying "return type bool expected, void received".

fullstackfool:
I'll revert the change.

The package tests, project tests and manually playing around with it manually works fine and returns consistent results. I don't doubt your own findings, I'm actually concerned about my own testing.

fullstackfool:
Happy for me to pull changes into this PR?

```php
class PermissionRegistrar
{
    //

    /**
     * PermissionRegistrar constructor.
     *
     * @param \Illuminate\Contracts\Auth\Access\Gate $gate
     * @param \Illuminate\Contracts\Cache\Repository $cache
     */
    public function __construct(Gate $gate, Repository $cache)
    {
        $this->gate = $gate;
        $this->cacheKey = config('permission.cache_key');
        $this->roleClass = config('permission.models.role');
        $this->permissionClass = config('permission.models.permission');
        $this->cacheMinutes = config('permission.cache_expiration_time');

        self::$cacheIsTaggable = ($cache->getStore() instanceof \Illuminate\Cache\TaggableStore);
        $this->cache = self::$cacheIsTaggable ? $cache->tags(config('permission.cache_key')) : $cache;
    }

    //

    /**
     * Flush the cache.
     */
    public function forgetCachedPermissions()
    {
        self::$cacheIsTaggable ? $this->cache->flush() : $this->cache->forget($this->cacheKey);
    }

    /**
     * Get the permissions based on the passed params.
     *
     * @param array $params
     *
     * @return \Illuminate\Support\Collection
     */
    {
        $permissions = $this->cache->remember($this->getKey($params), $this->cacheMinutes, function () use ($params) {
                ->when(($params && self::$cacheIsTaggable), function ($query) use ($params) {
                    return $query->where($params);
                })
                ->with('roles')
                ->get();
        });

        if (!self::$cacheIsTaggable) {
            foreach ($params as $attr => $value) {
                $permissions = $permissions->where($attr, $value);
            }
        }

        return $permissions;
    }

    /**
     * Get the key for caching.
     *
     * @param $params
     *
     * @return string
     */
    public function getKey(array $params): string
    {
        if ($params && self::$cacheIsTaggable) {
            return $this->cacheKey . '.' . implode('.', array_values($params));
        }

        return $this->cacheKey;
    }

    /**
     * Get an instance of the permission class.
     *
     * @return \Spatie\Permission\Contracts\Permission
     */
    public function getPermissionClass(): Permission
    {
        return app($this->permissionClass);
    }

    /**
     * Get an instance of the role class.
     *
     * @return \Spatie\Permission\Contracts\Role
     */
    public function getRoleClass(): Role
    {
        return app($this->roleClass);
    }
}
```

config:
```php
return [

    //

    /*
     * By default all permissions will be cached for 24 hours unless a permission or
     * role is updated. Then the cache will be flushed immediately.
     */

    /*
     * The key to use when tagging and prefixing entries in the cache.
     */

    'cache_key' => 'spatie.permission.cache',

    /*
     * When checking for a permission by passing a Permission instance,
     * this key determines what attribute on the Permissions model is
     * used to cache against.
     *
     * Ideally, this should match your preferred way of checking permissions, eg:
     * `$user->can('view-posts')` would be 'name'.
     */

    'cache_default_model_attribute' => 'name',
];
```

drbyte:
While looking at this I realized that flatting the `cache` config introduces more complication for upgraders in the future, because here we're adding 2 new keys which would later get relocated. And that's all to save the pain of referencing 1 key that we're trying to relocate now. Your original approach was indeed the better one, and I went in search of trying to figure out a way to minimize the pain. 

In the end I stumbled on a great discovery:
https://github.com/spatie/laravel-permission/blob/74eea48352d9d5931a034054814f818efaca5167/src/PermissionServiceProvider.php#L48-L51

This auto-merges the configs, which ends up adding the missing values from the default config stub.
Given that this avoids throwing errors due to missing keys, I'm satisfied with tagging a non-major release.  There's the side caveat that people who *have* customized the expiry time will find it being ignored in favor of the value inside the cache sub array.  But we could mitigate in the constructor by doing a `config('old-key', config('new-key'))` lookup instead of just new-key.

cbernard73:
I've only just started using this package so I don't know the details as well as either of you, but I thought I'd give my opinion on the caching strategy.

It seems that the current strategy is to cache the entire permissions table under a single cache key. I haven't read through the code in detail, but I came to that conclusion by looking at the cache table after multiple users login. I think that is sub-optimal, and I will tell you why.

I would guess that in most use cases, you are checking the permissions for a single user (i.e. a user wants to run a report, a user wants to edit a resource). I think you will rarely want to check the permissions for multiple users at the same time. In fact, the only situation I can think of is on a user management screen.

Therefore I think there should be a cache entry for each users' set of permissions. I think this would be relatively simple to implement and it would make a massive improvement to performance. 

The other issue with the current strategy is that you need to bust the cache whenever any users permissions change, or even, when a new user is created. 

Is my argument make sense, or have I misunderstood how the system is working today?

fullstackfool:
@cbernard73 Your argument makes perfects sense, and the solution you suggested is the purpose of this PR. If you take a look through the above conversation, especially the first comment, you can see the plan is to cache the model/permission combination, under tagged keys.

Let me know if you have any idea/questions

fullstackfool:
> In the end I stumbled on a great discovery:
> [laravel-permission/src/PermissionServiceProvider.php](https://github.com/spatie/laravel-permission/blob/74eea48352d9d5931a034054814f818efaca5167/src/PermissionServiceProvider.php#L48-L51)

A nice solution. The only catch I can find is that it uses `array_merge`, anyone declaring the `cache` array in a local `permissions` config they need to ensure they declare all the keys, for instance:

```php
'cache' => [
    'expiration_time' => 60 * 24,
],
```

will generate errors. It would need to be:
```php
'cache' => [
    'expiration_time' => 60 * 24,
    'key' => 'spatie.permission.cache',
    'default_model_attribute' => 'name',
],
```

I feel this is acceptable as it is a caveat in any package using `mergeConfigFrom()`.

I will update the PR with the agreed changes so we can have a final check-over.

drbyte:
> A nice solution. The only catch I can find is that it uses array_merge, so only merges one level deep. Anyone declaring the cache key in a local permissions config will need to ensure they declare all the sub-keys

Argh. Good catch.
If we add a default in our use of the `config()` helper, that should provide a reasonable safety net, since in this case the keys we're adding are new.
I'll document the new keys in the release notes as well, as a help to bring awareness to the matter.

fullstackfool:
The usages of `$config`  in `HasPermissions`, are we reverting those to `config()` calls? I'd like to keep the optimised (however small an optimisation it may actually be) array version, but if we are concerned about fallbacks then they'll need to be changed.

drbyte:
In both of your code posts above you had declared a `$cacheMinutes` property in `PermissionRegistrar`. What if we continue with that and read it in the trait? Either with a getter or by using a class constant like the $cacheIsTaggable.
While we'd still need to do the config() call in `hasPermissionTo`, we could drop it in `getAllPermissions`.

fullstackfool:
Had the same thought, the use of statics has cleaned up all the `$config` I was passing around in `HasPermissions`

I've noticed that in ServiceProvider we are [registering the config twice](https://github.com/spatie/laravel-permission/commit/23ccf38a44cc213779c722a5dce97dc18d8a6bfd)? Once in `boot` and once in `register`

```php
    public function boot(PermissionRegistrar $permissionLoader)
    {
        if (isNotLumen()) {
            $this->publishes([
                __DIR__ . '/../config/permission.php' => config_path('permission.php'),
            ], 'config');
        //
        }
        //
    }

    public function register()
    {
        if (isNotLumen()) {
            $this->mergeConfigFrom(
                __DIR__.'/../config/permission.php',
                'permission'
            );
        }

        $this->registerBladeExtensions();
    }
```
I can't find any reason to have it twice and [the docs](https://laravel.com/docs/5.7/packages#configuration) advise registering the config in `register`. If removed from `register` the tests fail, and nothing happens if removed from `boot()`. So I'm pretty sure the config was already being merged 😄

drbyte:
Does it make sense to sort out the duplication in a separate PR?  Or does fixing it here have a material benefit?

cbernard73:
@fullstackfool thanks for your confirmation. I've read back through the whole conversation now that I have a slightly better understanding, and I have one question about the approach being taken. 

If you use package.model.permission as they cache key, won't you end up with 1:1 relationship between the Model_has_Permission table and the cache? Would it better to limit the key to just cache.model?

fullstackfool:
> @fullstackfool thanks for your confirmation. I've read back through the whole conversation now that I have a slightly better understanding, and I have one question about the approach being taken.
> 
> If you use package.model.permission as they cache key, won't you end up with 1:1 relationship between the Model_has_Permission table and the cache? Would it better to limit the key to just cache.model?

Do you mean cache all the permissions for each model, under one key? We could, but then we would still have to loop over them again and check each one against the assertion. My thinking is that the indexing/searching of the cache store will be more efficient than doing it ourselves? 

I can certainly give it a test and compare the testing times, if you wish?

fullstackfool:
> Yes, I mean cache all permissions for each model under one key. I know it adds more work to the application layer (and I'll admit that I don't know the consequences of that right now), but if you use model.permission as the key, you are leaving the work at the database layer, in which case it seems to me like you're just moving the problem from the Model_Has_Permissions table to the Cache table.

The problem isn't so much the database structure, it's the repeated database connections and the parsing of large sets or permissions to find a specific key. Ultimately what we are interested in is can `a` do `b`? A single 1-to-1 relationship with a boolean result.

As it stands there are a lot of loops in the current structure, especially if the call to `hasPermissionTo` is made using a permission name/id, instead of an instance.

The structure you suggested is certainly possible, but I feel there is a couple of issues: 
1) It would require a fair amount of jigging around inside `HasPermissions` to make it work, so might be more appropriate for a major release? This PR is intended to be as non-disruptive as possible, to enable a minor release, with a mind to review the caching as a whole as release a more designed system later on.
2) Memcached and Redis, which I imagine has the lions share of Laravel caching, are both perfectly suited to handling hundreds of _millions_ of rows efficiently, I'm not sure about moving that processing to php because we want to save a few tens of thousand rows from entering the cache?

> In my application, the dashboard screen is the heaviest user of the permissions system. There are around 50 @can calls, but they're all for the same user. If we store all of a user's permissions in one single cache entry, then I'd get the permissions in one request to the cache. 
Similiar situation on my test project, although the permission checks breach 200 on the dashboard. You're right that you'd get all the permissions in one go, and could loop them and check. This _might_ be quicker than multiple calls to the cache, but on projects with larger sets of permissions and pages with less checks the efficiancy of this method decreases, as it could have been accomplished in few calls to the cache. Ultimately some side-by-side tests would be good, are you interested in forking and seeing what you can come up with? I'm unlikely to have time to do it in the next week.

> You mentioned in your first post that during your tests the queries to the database went from 1800+ to 0. What are you using for your cache in those tests?
Tested with memcached, redis, array and file, although the last 2 are largely pointless. Both memcached and redis gained a ~9x speed increase, and cut the calls to the database to 0 for permission checks.

cbernard73:
Yes, I figured it would cause significant changes to hasPermissionTo. I think your decision is correct.

fullstackfool:
@drbyte The config changes are in the PR now. Still feels a bit weird to be passing around the `PermissionRegistrar` like a party favor, but `HasPermissions` is much cleaner now, and ultimately, as you said, we can come back to this for a proper look at caching in a dedicated handler.

Would you mind taking another look when you are able?

src/PermissionRegistrar.php
```diff
@@ -16,29 +18,50 @@ class PermissionRegistrar
...
     public function __construct(Gate $gate, Repository $cache)
     {
         $this->gate = $gate;
-        $this->cache = $cache;
         $this->permissionClass = config('permission.models.permission');
         $this->roleClass = config('permission.models.role');
-        $this->cacheKey = config('permission.cache.key');
 
+        self::$cacheExpirationTime = config('permission.cache.expiration_time',
+            config('permission.cache.expiration_time'));
```
drbyte: The second one should be `cache_expiration_time` with underscore, right?

drbyte:
I ran it through a bunch of other cache tests I have for testing array/file/database/redis, and they all pass.

In reviewing the code there's just the one config default param to fix, and then I think it's good for merging!

Did you wanna squash it down into 2 or 3 commits (to preserve authorship of @ klincheg which you were kind enough to do!) instead of 20 commits?

src/PermissionRegistrar.php
fullstackfool: Good spot, thank you.

fullstackfool:
I'll have a go at squashing the commits now, although a bleary-eyed attempt at the weekend somehow removed one of your commits too.

fullstackfool:
Got it there in the end! Because the original commits from @klincheg were so long ago I had to pull in master, but that meant there were a ton of commits splitting his and mine when squashing.

Got it down to 4 though, would this be ok?

drbyte:
Ya, 4 is great!
Reviewing now...

fullstackfool:
Fantastic! Thanks to @klincheg for laying the groundworks, and @carlcassar for some pointers along the way.

Ping me if you are at all interested in collaborating on a larger caching structure for the next release

Thanks for your patience

EzrealR:
When I have a lot of permissions and roles, the redis cache is as large as tens of megabytes. It takes time for redis to read each time. Is there any solution?

fullstackfool:
#1626

PR #973, opened 2018-12-11
Use config in flushing cache in migration file

lloricode:
Cache key on migration in from config.

### DIFF
--- a/database/migrations/create_permission_tables.php.stub
+++ b/database/migrations/create_permission_tables.php.stub
@@ -78,13 +78,9 @@ class CreatePermissionTables extends Migration
 
             $table->primary(['permission_id', 'role_id']);
             
-            $cacheConfig = config('permission.cache');
-            $cacheStore = $cacheConfig['store'];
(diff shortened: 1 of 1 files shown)


database/migrations/create_permission_tables.php.stub
```diff
@@ -78,7 +78,8 @@ class CreatePermissionTables extends Migration
 
             $table->primary(['permission_id', 'role_id']);
 
-            app('cache')->forget(config('permission.cache.key'));
+            $cacheConfig = config('permission.cache');
+            app('cache')->store($cacheConfig['store'])->forget($cacheConfig['key']);
```
drbyte: This will probably work better:
```php
            app('cache')
            ->store(config('permission.cache.store') != 'default' ? config('permission.cache.store') : null)
            ->forget(config('permission.cache.key'));
```

lloricode: much better :)

drbyte:
Thanks!

PR #987, opened 2019-01-02
Added @hasanypermission blade directive. 

pkeogan:
Added the following blade directives:
- `@hasanypermission`
- `@endhasanypermission`

```
in *.blade.php file
...
@hasanypermission(['view_foobar', 'edit_foo'])
...
```

It just added the directives in the service provider. 

```
PermissionServiceProvider.php
...
            $bladeCompiler->directive('hasanypermission', function ($permissions) {
                return "<?php if(Auth::user()->hasAnyPermission({$permissions})): ?>";
            });
            $bladeCompiler->directive('endhasanypermission', function () {
                return '<?php endif; ?>';
            });
```

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -91,13 +91,18 @@ class PermissionRegistrar
      *
      * @return bool
      */
-    public function registerPermissions()
+    public function registerPermissions(): bool
     {
-        $this->getPermissions()->map(function ($permission) {
-            $this->gate->define($permission->name, function ($user) use ($permission) {
-                return $user->hasPermissionTo($permission);
-            });
+        $this->gate->before(function (Authorizable $user, string $ability) {
+            try {
+                if (method_exists($user, 'hasPermissionTo')) {
+                    return $user->hasPermissionTo($ability) ?: null;
+                }
+            } catch (PermissionDoesNotExist $e) {
+            }
         });
+
+        return true;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


drbyte:
Hi,
Thanks for your contribution.
Please remove the changes you made to `PermissionRegistrar`, as they are breaking the tests, and would be better as a separate PR anyway.

pkeogan:
@drbyte Thanks for the info. I'll remove the `PermissionRegistrar` changes, as that was not intended on this PR. I will also create a unit test for the new blade directive and resubmit into this PR.

drbyte:
> I'll remove the `PermissionRegistrar` changes, as that was not intended on this PR.

Tip for future: instead of using `master`, create a separate branch for each PR, so that other changes you explore don't get accidentally included into the PR.

pkeogan:
@drbyte I'll close this PR, untill I have the unit testing done, I'll also use a branch like you suggested. That makes alot more sense. is there any naming in the prefix I should use?
such as:
`dev-fix-1`

drbyte:
As for naming branches, my suggestion is just to use something meaningful, which indicates context. 
In the case of this pr, perhaps `add-hasanypermission-directive`.

Some people follow the "git-flow" naming conventions to varying degrees, where they prefix things with the word "feature-". However, in the long run in my experience it has little impact if your commit subject-lines and the PR title are properly explanatory.

PR #991, opened 2019-01-06
Updated UnauthorizedException

jnbn:
$requiredRoles and $requiredPermissions are now accessible and available for use on any ExceptionHandler.

ie. if you need to decide different actions for different permissions or roles; you can now access it via $exception->requiredPermissions or $exception->requiredPermissions on any handlers render method.


drbyte:
I'm confused. Why not use the built-in public accessor methods `getRequiredPermissions()` and `getRequiredRoles()`?

jnbn:
oh. my bad. Didn't see that it already exists. Thanks @drbyte

drbyte:
No worries. Thanks for offering your contribution!

PR #1000, opened 2019-01-16
Updated migration stub to don't recreate on vendor:publish

lbenevenuto:
Updated migration stub to don't recreate on vendor:publish

@spatie thanks to the amazing job.


drbyte:
Hi,
Thanks for offering a contribution.
Spatie generally prefers to use non-dated migration stubs, inserting the date only when publishing.

Please explain what specific problem you're trying to solve here. What's your workflow which made this necessary for you?

PR #1016, opened 2019-02-01
Remove use of array_wrap helper

Te7a-Houdini:
array helpers are being deprecated in L5.8

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\Permission\Traits;
 
+use Illuminate\Support\Arr;
 use Spatie\Permission\Guard;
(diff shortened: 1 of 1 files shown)


drbyte:
Two challenges to be sorted out here:

1. The tests are failing for Laravel 5.5
2. Lumen doesn't use facades unless they're manually enabled, and it's preferable not to force them to be turned on just for this package, especially when the only place they'd be used is this one line.

Thoughts?

drbyte:
Argh - ignore the comment about facades. I clearly had something else on my mind when I wrote that.

Still have to sort out the 5.5 issue.

drbyte:
Thanks!

(I adjusted it slightly for L5.3 compatibility.)

PR #1017, opened 2019-02-03
Add custom guard query to role scope

mo7amed-3bdalla7:
get users with specific roles and specific guard not default guard.

`$users = User::role('writer')->get(); // Returns only users with the role 'writer' and default guard 'web' `
But we Can use:
`$users = User::role('writer','api')->get(); // Returns only users with the role 'writer' and guard 'api' `

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -56,7 +56,7 @@ trait HasRoles
      *
      * @return \Illuminate\Database\Eloquent\Builder
      */
-    public function scopeRole(Builder $query, $roles): Builder
+    public function scopeRole(Builder $query, $roles, $guard = null): Builder
     {
         if ($roles instanceof Collection) {
             $roles = $roles->all();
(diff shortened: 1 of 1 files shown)


drbyte:
@mo7amed-3bdalla7 Thanks for this.

Where in your app are you using this? Is it in controlling access to things, or is it in a UI where you are assigning/removing roles to users? or something else?

mo7amed-3bdalla7:
> @mo7amed-3bdalla7 Thanks for this.
> 
> Where in your app are you using this? Is it in controlling access to things, or is it in a UI where you are assigning/removing roles to users? or something else?

I use when i want to get roles of specific guard name

drbyte:
Thanks. 
I have no problem with this PR, other than that there’s no test for this feature addition. 

I’m asking questions because the next major version is likely dropping guard-specific pairings in the db, which will require leaving the guard logic to your app.

Issue #978, opened 2018-12-18
RoleMiddleware checks for default guard user only 


MohamedQabbani:
Hello 

in the RoleMiddleware file it checks for the Auth::guest(), since I'm using different guard for my backend dashboard it always return true even when i'm logged in with different guard. 

```php
public function handle($request, Closure $next, $role)
    {
        if (Auth::guest()) {
            throw UnauthorizedException::notLoggedIn();
        }

        $roles = is_array($role)
            ? $role
            : explode('|', $role);

        if (! Auth::user()->hasAnyRole($roles)) {
            throw UnauthorizedException::forRoles($roles);
        }

        return $next($request);
    }
```

Before I submit my PR, would this change be acceptable? 

```php

public function handle($request, Closure $next, $role, $guard = null)
    {
        if (is_null($guard))
        {
            $guard = config('auth.defaults.guard');
        }
        if (Auth::guard($guard)->guest()) {
            throw UnauthorizedException::notLoggedIn();
        }

        $roles = is_array($role)
            ? $role
            : explode('|', $role);

        if (! Auth::guard($guard)->user()->hasAnyRole($roles)) {
            throw UnauthorizedException::forRoles($roles);
        }

        return $next($request);
    }

```

Thanks!

drbyte:
Hey, thanks for this.

I like your suggestion.

I think we should also apply it to the other middleware, since they have the same limitation.
Some of them use `app('auth')->` instead of the `Auth::` facade syntax, but the change is basically the same.

But we should also add some tests for this. Are you comfortable writing a test that demonstrates the need for this change?

helmut:
Happy to submit a pull request for this if you are still interested? I also came across this where having multiple logged in users from different guards gave unexpected access.

drbyte:
Certainly!

PR #1038, opened 2019-02-21
Added guard to middleware and added tests

helmut:
This is just a first draft of the changes required allow users to optionally specify the guard when using the permission middleware. As discussed here: https://github.com/spatie/laravel-permission/issues/978

```php
// Example without guard should use default guard from config file (usually web)
Route::group(['middleware' => ['role:auditor']], function () {
    //
});
// Example of middleware with guard specified
Route::group(['middleware' => ['role:auditor,admin']], function () {
    //
});
```

During the implementation of this feature I identified an issue with the current middleware that occurred when having multiple users logged into the system with different guards combined with running the `auth` middleware which updates the default guard in the config file during runtime.

```php
Auth::login($user);
Auth::guard('admin)->login($adminUser);

// I don't think this should pass because the permission middleware 
// did not specify the admin guard but it does because the auth:admin 
// middleware updated the default guard config file
Route::get('/test', function() {
   return 'success!'
})->middleware('auth:admin', 'permission:admin-only-permission');

// This should pass because we specified the guard for an admin permission
Route::get('/test', function() {
   return 'success!'
})->middleware('auth:admin', 'permission:admin-only-permission,admin');
```

The cause of the issue is that the Laravel configuration value `auth.defaults.guard` can be changed during runtime. It seems that when you perform a check on a route using the `auth` middleware a method is called that updates the config value. This appears to be so that subsequent calls to the auth factory default to that guard. That's handy in some situations however in our situation it meant that we had unexpected results from our middleware checks. By not providing a guard and therefore technically passing `null` as a guard, Laravel falls back to the default. And like I said, this value can change.

For example see this test which I have included:

```php
public function logging_in_both_users_and_calling_the_authenticate_middleware_with_a_guard_results_in_a_change_at_runtime_of_the_default_guard_config_value()
{
    Auth::login($this->testUser);

    $this->runMiddleware(
        $this->authenticateMiddleware
    );

    $this->assertEquals('web', config('auth.defaults.guard'));

    Auth::guard('admin')->login($this->testAdmin);

    $this->runMiddleware(
        $this->authenticateMiddleware, 'admin'
    );

    $this->assertEquals('admin', config('auth.defaults.guard'));
}
```

When combined with our middleware that uses the default guard we encounter an issue:

```php
public function ensure_that_the_role_middleware_is_not_affected_by_a_change_at_runtime_of_the_default_guard_config_value()
{
    Auth::login($this->testUser);

    $this->testAdmin->assignRole('testAdminRole');
    Auth::guard('admin')->login($this->testAdmin);

    // Simulate a runtime update of the config value
    config(['auth.defaults.guard' => 'admin']);

    $this->assertEquals(
        $this->runMiddleware(
            $this->roleMiddleware, 'testAdminRole'
        ), 403);

    $this->assertEquals(
        $this->runMiddleware(
            $this->roleMiddleware, 'testAdminRole', 'admin'
        ), 200);            
}
```
This test fails with the existing middleware implementation. The first assertion passes giving access to an admin role when it should be using the default guard which was originally `web`. So in this case I've updated the config value at runtime manually but that can happen if you use the `auth` middleware as shown in the first test.

My solution to this is for the middleware to fetch the default guard directly from the config file so we can be sure we get the original value. 

```php
if (is_null($guard)) {
    $guard = (require config_path('auth.php'))['defaults']['guard'];
}
```

This is just a draft but I hope I explained it in a way that can be understood. If not just let me know.

### DIFF
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -3,13 +3,18 @@
 namespace Spatie\Permission\Middlewares;
 
 use Closure;
+use Illuminate\Support\Facades\Auth;
 use Spatie\Permission\Exceptions\UnauthorizedException;
 
 class PermissionMiddleware
 {
-    public function handle($request, Closure $next, $permission)
+    public function handle($request, Closure $next, $permission, $guard = null)
     {
-        if (app('auth')->guest()) {
+        if (is_null($guard)) {
+            $guard = (require config_path('auth.php'))['defaults']['guard'];
+        }
+
+        if (Auth::guard($guard)->guest()) {
             throw UnauthorizedException::notLoggedIn();
         }
 
@@ -18,7 +23,7 @@ class PermissionMiddleware
             : explode('|', $permission);
 
         foreach ($permissions as $permission) {
-            if (app('auth')->user()->can($permission)) {
+            if (Auth::guard($guard)->user()->can($permission, $guard)) {
                 return $next($request);
             }
         }
@@ -26,3 +31,4 @@ class PermissionMiddleware
         throw UnauthorizedException::forPermissions($permissions);
     }
 }
+
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -23,7 +23,7 @@ class PermissionMiddleware
             : explode('|', $permission);
 
         foreach ($permissions as $permission) {
-            if (Auth::guard($guard)->user()->can($permission, $guard)) {
+            if (Auth::guard($guard)->user()->can($permission)) {
                 return $next($request);
             }
         }
--- a/src/Middlewares/PermissionMiddleware.php
+++ b/src/Middlewares/PermissionMiddleware.php
@@ -31,4 +31,3 @@ class PermissionMiddleware
         throw UnauthorizedException::forPermissions($permissions);
     }
 }
-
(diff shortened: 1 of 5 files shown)


helmut:
On further thought it does seem to be that Laravel relies heavily on the (changing) default guard throughout the code. For example the `@can` blade directives and the `can` middleware completely rely on the default guard because they fetch the default user from auth.

Perhaps this is not something that should be added to the middleware for this package after all. I'm coming to the conclusion that adding guards to the middleware adds more problems than it solves.

drbyte:
@helmut Thanks for your work on this. And the research you've done. And for explaining it in detail!

> The cause of the issue is that the Laravel configuration value `auth.defaults.guard` can be changed during runtime.

> On further thought it does seem to be that Laravel relies heavily on the (changing) default guard throughout the code. 
>For example the `@can` blade directives and the `can` middleware completely rely on the default guard because they fetch the default user from auth.

Yes, and this is a mismatch in implementation expectations many have complained about, because this package (well, v2.x anyway) expects a more rigid implementation of the "default" guard, not a "default" that's ever-changing as Laravel does it. We haven't changed/"fixed" it because it's a BC change.

> Perhaps this is not something that should be added to the middleware for this package after all. I'm coming to the conclusion that adding guards to the middleware adds more problems than it solves.

I tend to agree.

The plan is to have v3 of this package remove the guard-specific matching since for majority of implementations it's overkill and just duplicates what can be done in less code directly in the app already.

helmut:
No problems. Happy to help! Well sort of... haha!

Issue #1031, opened 2019-02-20
Permissions may not be cached if permission name contains a space


matthewgoslett:
Version: 2.32.0
Cache Driver: memcached

This is dependent on the cache driver, and specifically, what the cache driver deems a safe cache key.

The cache read/write & subsequent read will never hit if the permission contains a space: eg: "Manage Users".  The cache key in this case is "spatie.permissions.Manage Users".
I'd suggest hashing the key in the `getKey` method of the `PermissionRegistrar`.

matthewgoslett:
The `getPermissionCacheKey` in the `HasPermissions` trait is also affected by this.  I haven't dug too deeply yet, so not sure if there any other areas in the code which touch the cache.

matthewgoslett:
I'm busy testing a WIP fork here - https://github.com/spatie/laravel-permission/compare/master...MasterStart:master

In addition to this, and whilst a separate issue, but checking for the same permission multiple times, when the cache driver is memcached, will have a significant impact on performance.

ie: If you have a page which spits out a table of users, and for each row in the table, you want to check that the user can `View User`, can `Edit User`, can `Change User Password`, etc.  200 rows of data * 3 permissions = 600 memcached calls.  A solution previously suggested is to also implement an in-memory object cache.

drbyte:
On the "permission name contains a space" I get all the package tests passing for other drivers, but need to re-install memcached to test that.

On the separate performance comment, #804 began some discussion and investigation on improving performance, but it's been stalled awhile.

drbyte:
Confirmed. Spaces are indeed a problem when using memcached.

drbyte:
.. and array or redis ... :(

drbyte:
@matthewgoslett I'm thinking the following changes to CacheTest.php should be a suitable test for a possible fix:

```diff
    /** @test */
    public function has_permission_to_should_use_the_cache()
    {
+        app(Permission::class)->create(['name' => 'with spaces']);
-        $this->testUserRole->givePermissionTo(['edit-articles', 'edit-news']);
+        $this->testUserRole->givePermissionTo(['edit-articles', 'edit-news', 'with spaces']);
        $this->testUser->assignRole('testRole');

        $this->resetQueryCount();
        $this->assertTrue($this->testUser->hasPermissionTo('edit-articles'));
        $this->assertQueryCount($this->cache_init_count + $this->cache_load_count + $this->cache_run_count + $this->cache_relations_count);

[...]

Issue #1030, opened 2019-02-19
Web app is slow after giving permissions to roles


Zolak94:
I have around 50 permissions and 4 roles, after giving permissions to roles (around 150 in total in role_has_permissions table), my web app is really slow. My way of checking for permissions is in almost every function that returns view in my controllers Iwould have something like this:

```
if(!Auth::user()->hasPermissionTo('show location')) {
    return redirect()->back()->with('fail', 'You have no permission for that.');
}
```

Am I doing something wrong or?

Zolak94:
One more thing, I use a lot of @can in my navbar and I manage to get the time of loading and its 300+ms before it was only 20-30ms.

JinalSomaiya:
i have also same issue more query listed in laravel-debugbar. so, what to do?

drbyte:
Given that #1031 suggests there may be an issue with caching for certain drivers, a couple questions:
1. What cache driver are you using?
2. Does downgrading to version 2.25.0 improve the performance issue? (the caching logic was changed in 2.26.0)

JinalSomaiya:
i am using
1. CACHE_DRIVER=array
2.  "spatie/laravel-permission": "^2.15",

drbyte:
@JinalSomaiya is this query issue "new" for you? Or when did it start?

JinalSomaiya:
when i add more times check more query added in laravel-debugbar. 
like this:
`@if(auth()->user()->can(\App\Domain\Permission\Permission::PERMISSION_VIEW_AND_MANAGE_LEADS) || auth()->user()->can(\App\Domain\Permission\Permission::PERMISSION_ALLOCATE_LEADS))
                <li class="nav-item {{ request()->is('leads*') ? 'active' : '' }}">
                    <a href="{{ route('dealer:leads') }}" class="nav-link">
                        <i class="sidebar-icon lead-inquiries-icon"></i>
                        <p>Lead inquiries</p>
                    </a>
                </li>
            @endif`
if i am add more if condition then more query added. so, what to do?

matthewgoslett:
I've noticed the same huge performance drop after updating to the latest version (which has the caching tweaks & optimisations).

As an example, we have a page which displays a table of data.  For each row in the table, we check against various permissions.  In this particular sample, we land up doing 1740 synchronous calls (387ms) to a memcached server.

[...]

PR #1039, opened 2019-02-21
add in-memory permission caching, switch to single cache key + fix bu…

matthewgoslett:
This PR is an attempt to address both https://github.com/spatie/laravel-permission/issues/1031 and https://github.com/spatie/laravel-permission/issues/1030

* Add in-memory permission object caching
* Switch to single cache key for all permissions
* Fix bug where permissions containing spaces aren't cached

**This definitely needs eyes from someone who's more familiar with the codebase than myself.**

I setup a local testing env with a replica of my production database.

**Before Patch**

A page with a table of results, where each row checked against 4 to 5 permissions, resulted in roughly 500 calls to a memcached server taking up approximately 350+-ms.

**After Patch**

It does only 1 call to the memcached server for the single cache key `spatie.permission.cache`.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -122,20 +122,7 @@ trait HasPermissions
             throw new PermissionDoesNotExist;
         }
 
-        $registrar = app(PermissionRegistrar::class);
-        if (! $registrar::$cacheIsTaggable) {
-            return $this->hasUncachedPermissionTo($permission, $guardName);
-        }
-
-        return $registrar->getCacheStore()
-            ->tags($this->getCacheTags($permission))
-            ->remember(
-                $this->getPermissionCacheKey($permission),
-                $registrar::$cacheExpirationTime,
-                function () use ($permission, $guardName) {
-                    return $this->hasUncachedPermissionTo($permission, $guardName);
-                }
-            );
+        return $this->hasUncachedPermissionTo($permission, $guardName);
     }
 
     /**
@@ -206,27 +193,6 @@ trait HasPermissions
         }
 
         return md5($key);
-    }
-
-    /**
-     * Construct the tags for the cache entry.
-     *
-     * @param null|string|int|\Spatie\Permission\Contracts\Permission $permission
-     *
-     * @return array
-     */
-    protected function getCacheTags($permission = null)
-    {
-        $tags = [
-            PermissionRegistrar::$cacheKey,
-            $this->getClassCacheString(),
-        ];
-
-        if ($permission !== null) {
-            $tags[] = $this->getPermissionCacheString($permission);
-        }
-
-        return $tags;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


drbyte:
@matthewgoslett Thanks for this.

I've been reviewing and testing a bit.

Curious: in your testing, while you said you observed time improvements, what's the RAM impact? Granted it probably requires a larger dataset to compare outcomes.

matthewgoslett:
> @matthewgoslett Thanks for this.
> 
> I've been reviewing and testing a bit.
> 
> Curious: in your testing, while you said you observed time improvements, what's the RAM impact? Granted it probably requires a larger dataset to compare outcomes.

I'm yet to roll this out on production as I've been pretty swamped at work.  I'll try do some profiling with New Relic late this week / early next week.

drbyte:
To explore: Given that this PR moves to a single cache key for everything, I wonder if we can safely remove all the tagged-cache infrastructure too?

src/Traits/HasPermissions.php
```diff
@@ -205,7 +205,7 @@ protected function getPermissionCacheKey($permission = null)
             $key .= $this->getPermissionCacheString($permission);
         }
 
-        return $key;
+        return md5($key);
```
drbyte: Given this is returning a hash of the entire key string, we probably don't need the `PermissionRegistrar::$cacheKey.'.'` prefix included a couple lines above, right?

drbyte:
To investigate: why do the cache hit counts (tests) break when `database` is selected as the cache driver?

matthewgoslett:
I'm running this on production and am getting some horrific read time to memcached.

screenshot_1085

screenshot_1086

I'm trying to re-produce locally.

screenshot_1087

If I repl and try read the cache key, I just get this
```
>>> \Cache::get('spatie.permission.cache');
=> null
```
The value is always null, and `\Cache::has` returns false, yet DebugBar shows a cache hit, so I'm not exactly sure what's going on here.

matthewgoslett:
Ah re the value always being null, I didn't realise that the cache store is decorated by a TaggedCache.

src/Traits/HasPermissions.php
matthewgoslett: I've pushed a new commit now where I've removed use of tags (since we use a single cache key).
I'm pushing a 2nd commit now to remove the redundant code, `getPermissionCacheKey`, `getClassCacheString` and `getPermissionCacheString`.

matthewgoslett:
Here are the latest results from my production tests:

screenshot_1089

screenshot_1090

drbyte:
👍 Thumbs up!

I came up with pretty much the same while toying with it yesterday. I didn't have time then to sort out the failing tests that happen in database caching mode, but I'll merge this and then make those adjustments separately.

Thanks!

matthewgoslett:
Hey All

Have any changes been made into master since this release?

I switched from my fork to latest release here, and I'm back to original perf issues again.

matthewgoslett:
@drbyte I can confirm something iffy must've made it's way into master.

Here's my results from this morning:

screenshot_1092

drbyte:
#1054 removed an extra loop iteration. I wonder how that would be affecting things?
#1056 is not merged yet, but I'd be interested in your feedback about its impact

v16Studios:
> @matthewgoslett Thanks for this.
> 
> I've been reviewing and testing a bit.
> 
> Curious: in your testing, while you said you observed time improvements, what's the RAM impact? Granted it probably requires a larger dataset to compare outcomes.

In reference to this, we updated recently and are indeed hitting some memory issues with the single cache key.  We use memcached, 128M limit, at the moment 26 permissions, approx 300 roles (it's a multi-app management system where a single user can create roles for any number of apps with any combination of permissions), 1600 entires in the model_has_roles table and 7.6k entires in the role_has_permissions table.

Possibly happening in getPermissions()  ?  Anything we can try to mitigate this without having to bump the memory limits up?

v16Studios:
@drbyte Hi!  I commented on this issue a while ago as we do have a large dataset, and were running into mem issues, ended up disabling the cache to get it working.

PR #1042, opened 2019-02-26
Add unique columns

lloricode:
This will make sure a unique values in two columns, a name and guard_name

### DIFF
--- a/database/migrations/create_permission_tables.php.stub
+++ b/database/migrations/create_permission_tables.php.stub
@@ -20,6 +20,7 @@ class CreatePermissionTables extends Migration
             $table->increments('id');
             $table->string('name');
             $table->string('guard_name');
+            $table->unique(['name', 'guard_name'],'p_n_g_unique');
             $table->timestamps();
         });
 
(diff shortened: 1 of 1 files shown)


drbyte:
Please explain "why".

The package already enforces unique combinations, via code.

lloricode:
HI @drbyte,

Yes it already has, I made this PR to also optimize migration according to code implementation.

drbyte:
I'm not convinced this is necessary. Particularly when there are future plans to drop guard-specific db storage.

How in your application is there a need for this?

lloricode:
> Particularly when there are future plans to drop guard-specific db storage.

@drbyte you right thanks :+1:

ZaidBarghouthi:
@drbyte 
> ... future plans to drop guard-specific db storage.

Can you elaborate on this please.

drbyte:
> ... future plans to drop guard-specific db storage.

@ZaidBarghouthi I mean the majority of apps I've seen people using this for are assigning the *same* permissions to both `web` and `api` guards (and any other guards), meaning there's functionally *no* need to define a specific role+guard or permission+guard relation.  So, dropping the `guard_name` field from both tables is what `v3` will be doing. 
There are some other Issues where this has been discussed in more detail.

PR #1048, opened 2019-02-28
new getPermissionNames() function

highstrike:
Hello,

I've noticed the `getRoleNames()` function on the HasRole trait but the counter-part for permissions missing, so i've added the `getPermissionNames()` function in the HasPermission trait.

I've also written a test for it, let's see if it passes Travis CI.


drbyte:
thanks

PR #1050, opened 2019-03-01
null-filtering permissions inconsistency against roles

highstrike:
Noticed an inconsistency between roles and permissions and applied the fix :)

see https://github.com/spatie/laravel-permission/commit/7656d0c2c17b8183ae4029607cf6576ef8c8eeb6


drbyte:
Thanks

PR #1054, opened 2019-03-03
[Performance] avoid to iterate the list twice and returns on first found permission

vkarampinis:
Hello,

I have a project with 350 permissions. I sidebar menu checks permission to create the proper items for the user. There are around 350 checks in there.
With the current implementation (latest version) for each permission check in `getPermissions` the permission list iterates twice with two `wheres` (one for `guard_name` and one for `id` or `name`).
On my situation, each check takes around `6ms` to find the proper permission. With around 350 checks on my menu, I get a delay around 2 seconds to generate.

With this patch, the `getPermissions` iterates the permission list once and stops on the first found permission. This drops my times under `300ms`.

I believe this is a good performance improvement.
Any suggestions, thoughts are welcome.
Thanks

All tests are green.

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -137,11 +137,15 @@ class PermissionRegistrar
 
         $permissions = clone $this->permissions;
 
-        foreach ($params as $attr => $value) {
-            $permissions = $permissions->where($attr, $value);
-        }
+        return collect([$permissions->first(function ($item) use ($params) {
+            foreach ($params as $key => $value) {
+                if ($item[$key] !== $value) {
+                    return false;
+                }
+            }
+            return true;
+        })]);
 
-        return $permissions;
     }
 
     /**
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -143,9 +143,9 @@ class PermissionRegistrar
                     return false;
                 }
             }
+
             return true;
         })]);
-
     }
 
     /**


src/PermissionRegistrar.php
```diff
@@ -137,11 +137,15 @@ public function getPermissions(array $params = []): Collection
 
         $permissions = clone $this->permissions;
 
-        foreach ($params as $attr => $value) {
-            $permissions = $permissions->where($attr, $value);
-        }
+        return collect([$permissions->first(function ($item) use ($params) {
```
drbyte: Thanks for submitting this.

Question: since this `getPermissions()` call is initiated by the `Permission` model (ie: typically via `findByName`), which immediately calls `first()` again on this response, is there yet another opportunity here to improve performance even further? I'd be curious about your benchmark results.

vkarampinis: Hello,

we can't have any performance boost in `findByName` or in `findById`, because `getPermissions()` always returns one item. So the `first()` is happening almost instant.

vkarampinis:
Some benchmarks:

with `master` branch it takes `~1000ms` to find the proper permission in a list with 350 permissions and 350 calls.

with this patch with the same list and same calls:
if the looking permission is in the **first** place of the list, it needs `~33ms`
if the looking permission is in the **last** position of the list (worst case scenario), it takes `~360ms`

---

Playing around I came up with another solution

```
{
    if ($this->permissions === null) {
        $this->permissions = $this->cache->remember(self::$cacheKey, self::$cacheExpirationTime, function () {
                ->with('roles')
                ->get();
        })->reduce(function($carry, $item){
            $carry[$item->id . '.' . $item->guard_name] = $item;
            $carry[$item->name . '.' . $item->guard_name] = $item;
            return $carry;
        }, collect());
    }

    $key = implode('.', $params);
    return collect([$this->permissions[$key] ?? null]);
}
```
with this approach it takes `29ms` for the same checks, without care if the looking item is in the first or the last place.
Of course, this solution is a bit ugly and has the cons that the `$params` should always have the (`id` or `name` ) and `guard_name` attributes.
Also, the permission list consumes twice memory than before.

I don't believe this solution is so elegant.
In my situation, I will gain an average of `150ms` (assuming that some permissions are in first places and others are at the last) but I don't like the cons I mentioned above.

---

Any thoughts or suggestions are welcome.

src/PermissionRegistrar.php
drbyte: True

drbyte:
I agree - the cons aren't ideal.

drbyte:
While you're benchmarking, does removing `->with('roles')` change performance much in your app?

vkarampinis:
I only gain `~20ms` overall but this is because the object that gets from Redis is much smaller and deserialization is faster. This is logical.

But we can't remove `->with('roles')` as a lot of tests are failing.

drbyte:
Ya, those tests fail just because they're based on actual numbers of what's needed in order to do the relation load. But removing that should be a separate PR anyway.

drbyte:
This looks good. 
Thanks!

drbyte:
@vkarampinis if I recall correctly I reverted it because of the post from @matthewgoslett at 
https://github.com/spatie/laravel-permission/pull/1039#issuecomment-469565714 ... although I haven't seen any update for further verification.

Ideally I'd like to see the 3 of us collaborate to get all these changes cohabitating happily!

vkarampinis:
@drbyte oh this is very strange. We definitely need some feedback because I have the opposite results.

@matthewgoslett can you give some feedback?
How are your graphs with version `2.36.1` and how are with `2.36.0` ?
How many roles and how many permissions do you have in your app?

PR #1058, opened 2019-03-05
Revert #1054 "[Performance] avoid to iterate the list twice and returns on first found permission"

drbyte:
Reverts spatie/laravel-permission#1054


vkarampinis:
Hello @drbyte, can you share some info on what it breaks and forced to revert?

PR #1066, opened 2019-03-19
return roles after removing role

nachmanrosen:
Return model after removing roles to mimic other assigning roles and permission methods.
Authors: @tschams, @nachmanrosen


drbyte:
Thanks!

PR #1090, opened 2019-04-29
Add orchestra/testbench 3.8 support

CmdrSharp:
Since other dependencies specify support for Laravel 5.8, `orchestra/testbench` should also allow for the 3.8 release.


CmdrSharp:
Nevermind, forgot that 3.8 requires rewrites to match PHPUnit 8 spec. I see why it was left a version behind.

PR #1093, opened 2019-05-05
Speeding up performance during attach/detach-ing permissions.

Amirhb:
…enhance the performance of attach/detach-ing permissions.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -383,6 +383,10 @@ trait HasPermissions
     {
         $permissionClass = $this->getPermissionClass();
 
+        if (is_a($permissions, $permissionClass)) {
+            return $permissions;
+        }
+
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for your contribution!

Do you have measurements of performance change by doing this? 

Strictly speaking, is it really saving any database lookups? If `$permissions` is a class object, then `is_string` and `is_numeric` and `is_array` are going to be skipped, which moves to the `return` anyway. Or am I misreading your proposed code?

Amirhb:
Hi,

It speeded up the process up to 8-12 times when I passed more than 400 Permissions object in my application( yes it was a super-admin!). Before that it took more than 1 minutes and 20 seconds to sync the permissions to the role in the production server. After the update it took only 10 seconds.

And you're right when the right object is present there will be no more select query.

drbyte:
Closed due to inactivity.

PR #1105, opened 2019-05-18
Add check on helper functions

moeen-basra:
Add check on helper functions  if function already exists.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -10,3 +10,4 @@
 /tests              export-ignore
 /.editorconfig      export-ignore
 /docs               export-ignore
+/.styleci.yml       export-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 composer.lock
 vendor
 tests/temp
+.idea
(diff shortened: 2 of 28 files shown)


drbyte:
Can you explain the scenario where these helper methods conflicted? Which specific packages are conflicting?

moeen-basra:
> Can you explain the scenario where these helper methods conflicted? Which specific packages are conflicting?

Hi, it could conflict with every application which provide support for both Laravel and Lumen.
One of them is https://github.com/moeen-basra/photon

drbyte:
> One of them is https://github.com/moeen-basra/photon

Thanks.

> it could conflict with every application which provide support for both Laravel and Lumen.

Yes, of course. It's a much heated debate around Laravel right now.

I'm not convinced that merely checking `function_exists` is the ultimate best way, since you can't count on the pre-existing function doing what your own app/package/etc expects it to do. Namespacing is safer, and preferable. (So is limiting scope to not supporting things that aren't needed; think YAGNI.)

I'd prefer that we namespace this.

drbyte:
While this file will probably be removed in the next version, will merge for now.
Thanks.

PR #1109, opened 2019-05-20
Load roles relationship only when missing

sergiz:
Currently, if you use getPermissionsViaRoles or any related methods (like getAllPermissions), it will query the DB on every call. My PR makes sure that the DB is queried only, if permissions via roles have not been loaded yet.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -262,10 +262,17 @@ trait HasPermissions
      */
     public function getPermissionsViaRoles(): Collection
     {
-        return $this->load('roles', 'roles.permissions')
-            ->roles->flatMap(function ($role) {
-                return $role->permissions;
-            })->sort()->values();
+        $relationships = ['roles', 'roles.permissions'];
+
+        if (method_exists($this, 'loadMissing')) {
+            $this->loadMissing($relationships);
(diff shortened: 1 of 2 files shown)


sergiz:
I see someone already tried to make this fix in #903, but reverted it because of backwards compatibility. Since versions of Laravel, that don't support LoadMissing are already long time EOL, intentionally keeping these "bugs" around seems a bit off to me, but that's up to you.

BastianErlerArbeit:
Why not just check if the loadMissing exists?

```
    /**
     * Return all the permissions the model has via roles.
     */
    public function getPermissionsViaRoles(): Collection
    {
        if (method_exists($this, 'loadMissing')) {
            $this->loadMissing('roles', 'roles.permissions');
        } else {
            $this->load('roles', 'roles.permissions');
        }
        return $this->roles->flatMap(function ($role) {
            return $role->permissions;
        })->sort()->values();
    }
```

If you change the function in your PR it should be backwards compatible and still be able to eager load the permissions

sergiz:
@BastianErlerArbeit that's a great idea, thanks! Made the necessary changes, so now everything is in order.

drbyte:
Thanks!

Issue #1102, opened 2019-05-13
Why is role_has_permissions a separate table?


sergiz:
Hello.

Is there a specific reason why role's permissions are not saved in model_has_permissions table and require a dedicated table?

drbyte:
A different design approach could put everything in one table, but then you'd have less flexibility on how you use the data.

As it is right now, you can use roles-only, or permissions-only, or the combination of both.

What problem is this creating for you?  Why do you find yourself wanting them in the other table?

sergiz:
I am about to use this package and wanted to get to know it better first.

How come you wouldn't be able to use "roles-only, or permissions-only, or the combination of both", if role's permissions would be simply in model_has_permissions, like for every other model?

drbyte:
Apologies - I didn't read your post as deeply as I ought.

You raise a good point. There might be room for an efficiency gain there. I haven't investigated that ... but it requires mapping out all the polymorphic relations happening there.

sergiz:
Yeah, that I was thinking about. It seems to me that a separate table is redundant in this case, since model_has_permissions already includes everything we need to also store role's permissions.

I glanced over the implementation and it doesn't seem to be too hard to make a transition. I can look into it later and make a PR

PR #1112, opened 2019-05-22
Unified morph table for permissions

sergiz:
This is a follow-up from #1102.

I made it completely optional. This change does not make any difference for existing (or future) users, if they do not opt-in for it.

It  basically checks, if the config value 'permission.table_names.role_has_permissions' exists and, if it does there is no difference from now. If this value is not set, then the role_has_permissions table will not be created (and used) and model_has_permissions will be used instead (like for any other model).

It gives the end user completely free hands to decide whether they want completely separate table to handle role's permissions, or to use the existing model's table.


drbyte:
@sergiz Thanks for this.
I keep meaning to sit and inspect whether any additional tests should be included for this change, but haven't had the time to think it through. 
Your thoughts?

sergiz:
Yeah, I could make create some tests. Hopefully I find time to do that soon.

drbyte:
Given MorphToMany inherits from BelongsToMany, can we just leave the BelongsToMany as the return type?

sergiz:
Sorry that it's taking so long, but I'm fully occupied until October. I will look into all these issues then.

sergiz:
Okay, I addressed your concerns now.

a) BelongsToMany return type declaration cannot be left there, as PHP requires it to exactly match the returned instance, so MorpthToMany will not satisfy it. This is one of the reasons why I even made this PR, as I couldn't make this change on my local copy without maintaing my own fork of this package.

b) I added a few tests to make sure that everything is working.

drbyte:
@sergiz What are your thoughts on switching this to be the default, and the split tables be the optional mode?

sergiz:
I think it would make sense since there is no real benefit of having two tables that do the same thing. However, I would consider this a breaking change that requres a new major version.

If you are going to release a new major version with this, then I would also think about simply removing this option all together and simply make it that unified table is the only option.

drbyte:
> I think it would make sense. However, I would consider this a breaking change that requres a new major version.

Yes, that is what I'm aiming for.

> If you are going to release a new major version with this, then I would also think about simply removing this option all together and simply make it that unified table is the only option.

Thoughts on best practices for people migrating their data from the old table into the new?

sergiz:
There could be a section in the upgrade guide, how to make a migration. Eventually you would have to select relations from role_has_permissions and insert them to model_has_permissions. "models.role" config should be taken into consideration, including possible custom morph map.

This can be achieved on the DB level using a "INSERT INTO SELECT" query.

sergiz:
I think this PR should be reopened since the final decision hasn't been made yet.

freekmurze:
I've decided to not merge this in, to keep that package as small as it can be.

sergiz:
> 
> 
> I've decided to not merge this in, to keep that package as small as it can be.

So 5 tables is less than 4? Interesting.

PR #1122, opened 2019-06-06
Explain using package with laravel's authorize method

cbernard73:
There's not much in the documentation about using the package from within a Controller. I've added a paragraph to address that area.


drbyte:
Hi, 
With your edits, are you simply trying to say "if your user has a permission named the same as the policy method (ie: "show" in this case), then that permission will take precedence, and the policy won't trigger"?

cbernard73:
Yes, I guess I am.

Regards,
Ciaran.

---
Ciaran Bernard
Partner
Outbound Assist by Ascend
ph: +353-87-927-2160

drbyte:
Cool. Thanks for submitting this. 
I've simplified the description a little to make it more concise.

PR #1131, opened 2019-06-20
Flush permissions cache when removing a role from a model

rb1193:
According to the docs the hasRoles trait's removeRole() method should clear the permissions cache, but currently it doesn't seem to be. Hopefully this PR should resolve

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -150,6 +150,8 @@ trait HasRoles
 
         $this->load('roles');
 
+        $this->forgetCachedPermissions();
+
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


src/Traits/HasRoles.php
```diff
@@ -150,6 +150,8 @@ public function removeRole($role)
 
         $this->load('roles');
 
+        $this->forgetCachedPermissions();
```
drbyte: Would it make sense for the `forget` go before the re-`load`?

drbyte:
Thanks for catching this scenario.

Just one adjustment, noted in the code above.

src/Traits/HasRoles.php
rb1193: I don't think it particularly matters? The load call doesn't hit the permissions cache, so far as I can tell. I notice that assignRole also leaves the cache forget until last, if that's any sort of guide.

PR #1170, opened 2019-07-25
add timestamp in model_has_permissions & model_has_roles tables

aligilan:
In order to find, when assigned role/permission to related model.


sergiz:
I don't think most users would find this useful. Probably the reason why Morph relationships don't have timestamps by default.

drbyte:
@aligilan Thanks for submitting this.

This is the sort of change that is probably best implemented in your own app, not in the package. There you can implement it in your migrations and override the methods as necessary.

Wishing you the best of success with your apps!

aligilan:
Ok, thank you for this fantastic package.

PR #1176, opened 2019-08-01
Update middleware example

akkhan20:
There is no point of using "role_or_permission" if we only have to define a role or a permission on it as they both have dedicated middlewares for that.


drbyte:
Thanks!

PR #1177, opened 2019-08-07
add unique modifier to name attribute of roles and permissions table

murtazasultani:
I am just adding the unique modifier to the name attribute of roles and permissions table because of when checking for a specific role or permission when there are two different roles with the same name, here I have been faced with such an issue. the same issue exists with the permissions table.


drbyte:
Thanks for your submission.
However, since this version allows to have roles of the same name on different guards, this unique index would break that.

Further, the application already enforces non-duplicates at the code level if you use the provided methods instead of directly adding records to the tables.

PR #1197, opened 2019-09-01
Add Display Name & Description

sibche2013:
Add Display Name & Description For Every Roles


drbyte:
Thanks for your submission.

We are not planning to implement additional fields on the tables at this time.

However, you can certainly add them to your own migrations in your own app, and then also add the necessary code to interact with them.

PR #1206, opened 2019-09-06
add UNIQUE constraint to name/guard_name column pair for roles and pe…

dennis-git:
…rmissions tables

add UNIQUE constraint to name/guard_name column pair for roles and permissions tables in the migration to prevent creating duplicate Roles and Permissions per guard, which can break someone's code quite easily.

### DIFF
--- a/database/migrations/create_permission_tables.php.stub
+++ b/database/migrations/create_permission_tables.php.stub
@@ -21,6 +21,7 @@ class CreatePermissionTables extends Migration
             $table->string('name');
             $table->string('guard_name');
             $table->timestamps();
+            $table->unique(['name', 'guard_name']);
         });
 
         Schema::create($tableNames['roles'], function (Blueprint $table) {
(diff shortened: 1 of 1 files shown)


drbyte:
How are you inserting these duplicates? 
This package opted to enforce that integrity in code, rather than in the database.

dennis-git:
ok, that's fair but I personally prefer to have it also enforced on the database layer. This solution has another benefit of adding a UNIQUE index on the name/guard column pair to speed up queries. The code does throw RoleAlreadyExists and PermissionAlreadyExists Exceptions. I personally added the UNIQUE constraint to my codebase to be 100% sure no duplicate Roles/Permissions per Guard name can exist that can create havoc or security issue at any point in the future and at the same time speed up queries.

drbyte:
Yes, there's some value in the speed improvement especially if you're doing lookups outside of what the current user has access to.

I'm going to close this for now. 
One can always add those indices to their own migrations where it benefits them.

Still, thanks for submitting: it's always helpful to know how people are using the package, how they're tweaking it, and what their pain points are!

Related: #1042, #1177

PR #1211, opened 2019-09-10
Fix hasAnyRole in docs

Jubeki:
hasAnyRole requires an array instead of multiple parameters


drbyte:
Thanks.

drbyte:
Aside: you can also pass a pipe-delimited string in cases where an array isn't ideal.

PR #1240, opened 2019-10-07
Add timestamps columns to pivot tables

### LINKED ISSUE #1228: Timestamps on the pivots
We use this package on a fairly large app and find it super helpful!

There has been numerous occasions where it would be immensely helpful to have timestamps on the pivots. Specifically in regards to using a role or permission to isolate a specific group in your app.

For instance, we developed a new business segment and permitted people to register and designated them as an "invite" with a role for this new business case. Some invites include users who are already using another one of our products. We used a role to track which users have requested an invite.

In the case of the new users, we have tracking for when the request was made, in other cases, a pivot timestamp would have really come in handy.

Thank you for the fantastic packages.

josepostiga:
This PR adds support for timestamps columns in the pivot tables.

To avoid adding breaking changes, I've added a new migration stub to update the relevant tables, adding the necessary timestamp columns.

Fixes #1228 

This PR was done in cooperation with @infus0815.


drbyte:
@josepostiga Thanks for this.
What are your thoughts on people wanting helper methods to retrieve timestamp details?
Or do you think simply documenting how to get time details through the models would be enough? Care to write those simple details into the docs?

josepostiga:
@drbyte honestly, I haven't thought much about helpers to access the pivot's timestamps. I can't see a really good case to use it, unless a very specific need, like the one described in the related issue (#1228 ).

I feel almost the same about documenting that you can access the timestamps, as it's explained in the framework documentation site itself.

What are your thoughts about it?

drbyte:
Ya, I think a simple entry to the /docs/advanced/other.md file with an example showing how to retrieve it is probably sufficient.

josepostiga:
@drbyte I've updated the doc file you pointed out. Care to take a look?

sergiz:
This was already rejected in #1170. I think that only handful of users would find these timestamps useful (that's why Laravel also disabled them by default), and those can add them by themselves.

drbyte:
@sergiz is correct. That's why I haven't merged this yet.

Still weighing out the pros and cons of having this in the core of this package.
It's been requested half a dozen times. Which means there's some interest, but it's small.

I'm toying with options:
1. Merge this and force it on everyone, even those who don't want the extra overhead on thousands of records. (Overhead = extra data storage and data transfer and CPU time.)
2. Rework it to make it supported in-package but configurable (on/off in config file).
3. Simply document how to do it in your own app if you need it, as part of the Advanced section of the docs, on Extending the Package.

/ping @danrichards @josepostiga @sergiz ... and anyone else: thoughts?

sergiz:
There are thousands of users of this package, so if only a couple of those people need it, I think it's redundant to force it on everyone.

If I have to choose one option, it would be 3. However, this modification is so simple (and not even directly related to this package, but Laravel in general), so ...

josepostiga:
I really don't have much more to say. I simply saw an issue tagged as "enhancement" and contributed to that. 

Since another similar PR was closed, and since I can agree is something that depends on specific use cases, maybe it's ok to close it. Users of this package can use their own models, that can extend this package' ones, and then add the extra `withTimestamps()` and table columns, too.

danrichards:
Whelp, if an on / off switch solves the overhead problem. Then I don't see any reason not to make the improvement.

For more perspective, besides what is outlined in https://github.com/spatie/laravel-permission/issues/1228 you could consider why we have roles and permissions in the first place, **security**.

Example:

If you discover a data liability in system and know only 3 people had access at the time, instead of the active set (say, 3000, these are just arbitrary numbers), it could mean a big difference in how or if you push a remediation or make an internal decision. And how you notify the impacted group.

PR #1249, opened 2019-10-16
Added publish shortcut console commands.

joskolenberg:
This PR adds shortcut commands to make publishing the config file and migration more easily:

- `permission:publish` as a shortcut to `vendor:publish --provider="Spatie\Permission\PermissionServiceProvider"`
- `permission:migration` as a shortcut to `vendor:publish --provider="Spatie\Permission\PermissionServiceProvider" --tag="migrations"`
- `permission:config` as a shortcut to `vendor:publish --provider="Spatie\Permission\PermissionServiceProvider" --tag="config"`

Since the console doesn't provide autocompletion I often find myself copying these lines from the docs.
This way the commands are easier to remember or can be looked up using the `artisan:list` command.

Maybe three commands is overkill, another option would be to use a single command with a tag parameter like
- `permission:publish`
- `permission:publish --tag="migrations"`
- `permission:publish --tag="config"`

Love to discuss this one, would be interesting for other packages as well imho.

### DIFF
--- a/docs/installation-laravel.md
+++ b/docs/installation-laravel.md
@@ -23,7 +23,7 @@ The service provider will automatically get registered. Or you may manually add 
 You must publish [the migration](https://github.com/spatie/laravel-permission/blob/master/database/migrations/create_permission_tables.php.stub) with:
 
 ```bash
-php artisan vendor:publish --provider="Spatie\Permission\PermissionServiceProvider" --tag="migrations"
+php artisan permission:migration
 ```
 
 If you're using UUIDs or GUIDs for your `User` models you can update the `create_permission_tables.php` migration and replace `$table->unsignedBigInteger($columnNames['model_morph_key'])` with `$table->uuid($columnNames['model_morph_key'])`.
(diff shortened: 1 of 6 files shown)


drbyte:
@joskolenberg Thanks for this.
I worry that adding more commands to the artisan console is contributing to a problem I wish Laravel didn't have: clutter with too many things in that list already.

I too have never been a fan of the need to type out (copy/paste) the long service-provider name and then a tag.

Nowadays I almost always just run `php artisan vendor:publish` and just choose from the list shown. It's faster even than reading through the other commands to find the one I can't remember!

That said, I'm open to others' opinions on the merits of including this, or not.

joskolenberg:
@drbyte Thanks for the response.

I think you're right about too much clutter in that list. The option to have three separate commands indeed is overkill, the `permission:publish` command with optional `--tag` option still could be helpful imo.

I'll wait for more opinions here. If there is enough interest to merge this I'll update the code to have a single command with `--tag` option.

sergiz:
I am against this. You will run this command exactly once per project and it takes you literally less than 20 seconds to go to this github page and copy it (even less, if you save it to you local notes or even create a bash alias).

It adds unnecessary clutter, even if it's only one and it doesn't contribute anything to your project when in production.

joskolenberg:
Yeah, I guess you’re right.. Thanks for your thoughts.
Started having doubts about it too, think we can close this one.

PR #1254, opened 2019-10-21
Remove version checks

sergiz:
Now that this package only supports modern versions of Laravel, these old version checks are not needed anymore.


drbyte:
Thanks!

PR #1256, opened 2019-10-24
Use checkPermissionTo method at gate check

sergiz:
Since we have the checkPermissionTo method which already handles the possible exception, it makes no sense to use hasPermissionTo at the gate check.


drbyte:
Thanks

PR #1259, opened 2019-10-24
hasDirectPermission cleanup

sergiz:
This PR makes two changes:

a) It deletes some unreachable code. findByName and findById throw an exception, so the following if clause was never fulfilled.

b) If instance of Permission model cannot be retrieved, exception should be thrown and not simply false returned. This is clearly indicated by the tests names (which were also properly fixed).

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -246,7 +246,7 @@ trait HasPermissions
         }
 
         if (! $permission instanceof Permission) {
-            return false;
+            throw new PermissionDoesNotExist;
         }
 
         return $this->permissions->contains('id', $permission->id);
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -239,16 +239,10 @@ trait HasPermissions
 
         if (is_string($permission)) {
             $permission = $permissionClass->findByName($permission, $this->getDefaultGuardName());
-            if (! $permission) {
-                return false;
-            }
         }
 
         if (is_int($permission)) {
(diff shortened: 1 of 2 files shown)


src/Traits/HasPermissions.php
```diff
@@ -239,20 +239,14 @@ public function hasDirectPermission($permission): bool
...
         if (is_int($permission)) {
             $permission = $permissionClass->findById($permission, $this->getDefaultGuardName());
-            if (! $permission) {
-                return false;
-            }
         }
 
         if (! $permission instanceof Permission) {
-            return false;
+            throw new PermissionDoesNotExist;
```
drbyte: To what extent is this a breaking change?

sergiz: I consider this a bugfix really.

When PR #820 was made, I guess that the author simply glanced over the function and concluded that it was returning false on the previous two conditions. Since that code is unreachable, in reality it was already throwing PermissionDoesNotExist exception because that what findByName/Id method does.
So what we have here is a complete mess because the function throws PermissionDoesNotExist, if we pass non-existant permission as an integer or a string, but returns false, if we pass anything else that is not a valid Permission instance.

hasPermissionTo already has the same behaviour that this function would have after this PR is merged.

Whoever is using hasDirectPermission method directly, must already be handling this exception. I don't think this change would break anything. Furthermore, whoever differentiates between this method throwing an exception or returning boolean, must now do a check manually, to see, if the method is dealing with non-existant permission, or the model actually does not have a permission, which is actually existant.

drbyte:
Thanks

PR #1261, opened 2019-10-25
Enforce unique constraints on database level

sergiz:
Role and Permission unique name <-> guard constraints are already enforced on the PHP level, so it only makes sense to also enforce them on database level.

https://github.com/spatie/laravel-permission/blob/e3821559c69b2b1ea8fdd967fa89ba13a01ead78/src/Models/Role.php#L37
https://github.com/spatie/laravel-permission/blob/e3821559c69b2b1ea8fdd967fa89ba13a01ead78/src/Models/Permission.php#L40


drbyte:
Merging this time due to agreement with Dennis's comment in #1206 regarding the improved index performance.
Ref #1206 #1042

PR #1291, opened 2019-12-01
Update installation-laravel.md

erlangp:


erlangp:
we must generate & configure the config first, because when i try to run migration:

```
php artisan migrate
Migrating: 2019_12_01_205310_create_permission_tables

   Illuminate\Database\QueryException  : SQLSTATE[42000]: Syntax error or access violation: 1103 Incorrect table name '' (SQL: create table `` (`id` bigint unsigned
```

docs/installation-laravel.md
```diff
@@ -166,7 +166,13 @@ return [
 ];
 ```
 
-After the config & migration have been published you can create the role- and permission-tables by running the migrations:
+Then, apply the configuration:
+
+```bash
+php artisan config:cache
```
drbyte: Why run config:cache here?

erlangp: without run that command, i got SQL error like i mention before

erlangp:
step i have tried:
1. i publish migration
2. i run migration migrate
3. i got SQL error

4. i  publish config
5. i run migration migate
6. i got SQL error

7. i run config:cache
8. i run migration migrate
9. i got migration Success

erlangp:
so the happy path:
1. publish migration
2. publish config
3. run config:cache
4. run migration migrate

(this i tried on my local computer, windows 8)

drbyte:
I think you're better to run `config:clear` in dev. There's NO need to run config:cache in dev. That's only for production environments.

But I wouldn't put anything about config caching in this package's docs. Those are dev skill things, not package-specific things.

drbyte:
With this package, as long as you've published the config and the migration (which can be done with vendor:publish and choose this package from the list), then when you run the migration things will be fine.

docs/installation-laravel.md
erlangp: must i change with `config:clear` sir?

erlangp:
i removed that line sir

erlangp:
thanks sir :D

PR #1306, opened 2019-12-20
Adding more documentation to the code

claretnnamocha:


drbyte:
Thanks for your submission.

We removed most of these comments because we're using PHP 7's declaration of return types, so the docblocks are redundant.

claretnnamocha:
Okay then @drbyte, thanks for the response

PR #1312, opened 2019-12-28
Add guard for middleware

masihjay-z:
Hi
I add choosing specific guard for middleware.
i had really needed it.

### DIFF
--- a/docs/basic-usage/middleware.md
+++ b/docs/basic-usage/middleware.md
@@ -53,6 +53,21 @@ Route::group(['middleware' => ['role_or_permission:super-admin|edit articles']],
     //
 });
 ```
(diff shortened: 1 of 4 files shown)


drbyte:
Hi @masihjay-z 
Thanks for sharing your idea.  Unfortunately the tests are failing.

While looking into your suggestion it strikes me that (while this package has some known limitations related to matching up of guards) the middleware guard checks might be better left to a different section of the code, instead of specifying a guard in the name of the permission you're passing to the middleware.

Wouldn't it make better sense to let Laravel's Authentication layer validate the user using your configured guards, which then sets the guard that the user is using, and then have the middleware check to see if there's a match for the specified permission-name and the user's guard (instead of telling the middleware which guard to use, as part of the permission name)?

PR #1316, opened 2020-01-06
Added missing `guardName` to Exception `PermissionDoesNotExist`

musapinar:
Referenced here: https://github.com/spatie/laravel-permission/blob/master/src/Models/Permission.php#L110


drbyte:
Thanks

PR #1324, opened 2020-01-14
Update CacheTest.php

musapinar:
Fixed random test failure adding ```->sort()->values()``` to ```$actual``` Collection in ```get_all_permissions_should_use_the_cache()```

https://i.imgur.com/MoS3erH.png

PR #1325, opened 2020-01-15
Update HasPermissionsTest.php

musapinar:
Fixed random test failure adding ```->sort()->values()``` to ```pluck()```
Exact same issue as https://github.com/spatie/laravel-permission/pull/1324

### DIFF
--- a/tests/HasPermissionsTest.php
+++ b/tests/HasPermissionsTest.php
@@ -376,7 +376,7 @@ class HasPermissionsTest extends TestCase
 
         $this->assertEquals(
             collect(['edit-articles', 'edit-news']),
-            $this->testUser->getAllPermissions()->pluck('name')
+            $this->testUser->getAllPermissions()->pluck('name')->sort()->values()
         );
     }
 


drbyte:
I've never seen these "random test failures" that you mention, neither in local dev nor in CI.
What's causing it for you?

musapinar:
> 
> 
> I've never seen these "random test failures" that you mention, neither in local dev nor in CI.
> What's causing it for you?

As you can see in the screenshot here: https://i.imgur.com/MoS3erH.png
The Collection's key/value pair may be different.

Sometimes I have :
```
Illuminate\Support\Collection {#1638
  #items: array:2 [
    0 => "Edit Articles"
    1 => "Edit News"
  ]
}
```

Some other times I have : 
```
Illuminate\Support\Collection {#1638
  #items: array:2 [
    0 => "Edit News"
    1 => "Edit Articles"
  ]
}
```
Which causes the subsequent assertion to fail randomly.
So I am forcing alphabetical order with ```->sort()->values()```

Context:
Laravel 6.10
PHP 7.3

After further investigations, I figured out it's actually because I am using UUIDs instead of this package's default bigIncrement IDs. ```pluck()``` orders by ID ASC indeed by default, but in my case the UUIDs are generated on the fly via a model observer so it makes no sense to sort by UUID and explains perfectly why it randomly failed.

TL;DR : False flag. Sorry about that.

drbyte:
For consistency with the other PR already merged, I'll merge this too. 
While there are no plans to natively force UUIDs in this package, this is not a problematic change.

PR #1326, opened 2020-01-15
Add splat operator for hasAnyRole()

musapinar:
Added splat operator to ```hasAnyRole()```

Fixes the following : https://github.com/spatie/laravel-permission/blob/master/tests/HasRolesTest.php#L492

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -218,7 +218,7 @@ trait HasRoles
      *
      * @return bool
      */
-    public function hasAnyRole($roles): bool
+    public function hasAnyRole(...$roles): bool
     {
         return $this->hasRole($roles);
     }


drbyte:
> Fixes the following : https://github.com/spatie/laravel-permission/blob/master/tests/HasRolesTest.php#L492

Are you saying that that line of that test is failing in your environment? 
Or are you saying that the test itself is wrong? (And therefore the test should be updated too?)

musapinar:
> 
> 
> > Fixes the following : https://github.com/spatie/laravel-permission/blob/master/tests/HasRolesTest.php#L492
> 
> Are you saying that that line of that test is failing in your environment?
> Or are you saying that the test itself is wrong? (And therefore the test should be updated too?)

https://github.com/spatie/laravel-permission/blob/master/src/Traits/HasRoles.php#L221
Method ```hasAnyRole()``` accepts only 1 parameter, ...

https://github.com/spatie/laravel-permission/blob/master/tests/HasRolesTest.php#L492
... while method call in the test suggests in cake take 2. The test currently passes because it is "poorly" written (```assertFalse```)

I cannot think of any reason for why they did not add the splat operator for that particular function.
If you merge the PR you do not need to update that test. 

However, if you are incline to merge the PR, I'd also suggest adding following additional tests, like : 
```
/** @test */
public function it_can_check_against_any_multiple_roles_using_multiple_arguments()
{
    this->testUser->assignRole('testRole');
    $this->assertTrue($this->testUser->hasAnyRole($this->testAdminRole, ['testRole'], ['I Am An' => 'Associative Array Wtf'], 'I Am A Role That Does Not Even Exist Lol'));
}

/** @test */
public function it_returns_false_instead_of_an_exception_when_checking_against_any_undefined_roles_using_multiple_arguments()
{
    $this->assertFalse($this->testUser->hasAnyRole('I Am A Role That Does Not Even Exist Lol', $this->testAdminRole, ['I Am An' => 'Associative Array Wtf']));
}
```

drbyte:
I agree with your points.
I think the splat operator wasn't used because the function is (almost) an alias to `hasRole()`, which can't use the splat operator.
That said, given their function signatures are intentionally different I think it does make sense to do as you propose.

PR #1328, opened 2020-01-16
Remove duplicate scope lookups

michaelr0:
Something that caught my eye is that scopeRole will continue to query for each (int, string or collection) parameter that has been passed to it.

In a scenario where the a role exists with the id of 1 and a name of 'User'
`App\User::role([1, 'User'])->first();` will cause two queries to be fired off, however both will return the same result.

``select * from `roles` where `id` = 1 and `guard_name` = 'web' limit 1`` Finished in 390us.
``select * from `roles` where `name` = 'User' and `guard_name` = 'web' limit 1`` Finished in 370us.

It might be worthwhile to change the original scopeRole to combat duplicate lookups.
```
    /**
     * Scope the model query to certain roles only.
     *
     * @param \Illuminate\Database\Eloquent\Builder $query
     * @param string|array|\Spatie\Permission\Contracts\Role|\Illuminate\Support\Collection $roles
     * @param string $guard
     *
     * @return \Illuminate\Database\Eloquent\Builder
     */
    public function scopeRole(Builder $query, $listOfRoles, $guard = null): Builder
    {
        if ($listOfRoles instanceof Collection) {
            $listOfRoles = $listOfRoles->all();
        }

        if (!is_array($listOfRoles)) {
            $listOfRoles = [$listOfRoles];
        }

        $guard = $guard ?: $this->getDefaultGuardName();

        $listOfRoleNames = [];
        $roles = [];

        foreach ($listOfRoles as $role) {
            if ($role instanceof Role) {
                $roles[$role->id] = $role;
            } else {
                $method = is_numeric($role) ? 'findById' : 'findByName';

                if (!is_numeric($role) && !isset($listOfRoleNames[$role])) {
                    $thisRole = $this->getRoleClass()->{$method}($role, $guard);
                    $roles[$thisRole->id] = $thisRole;
                    $listOfRoleNames[$thisRole->name] = null;
                } elseif (is_numeric($role) && !isset($roles[$role])) {
                    $thisRole = $this->getRoleClass()->{$method}($role, $guard);
                    $roles[$thisRole->id] = $thisRole;
                }
            }
        }

        return $query->whereHas('roles', function ($query) use ($roles) {
            $query->where(function ($query) use ($roles) {
                foreach ($roles as $role) {
                    $query->orWhere(config('permission.table_names.roles') . '.id', $role->id);
                }
            });
        });
    }
```

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -18,7 +18,7 @@ trait HasPermissions
     public static function bootHasPermissions()
     {
         static::deleting(function ($model) {
-            if (method_exists($model, 'isForceDeleting') && !$model->isForceDeleting()) {
+            if (method_exists($model, 'isForceDeleting') && ! $model->isForceDeleting()) {
                 return;
             }
 
@@ -28,7 +28,7 @@ trait HasPermissions
 
     public function getPermissionClass()
     {
-        if (!isset($this->permissionClass)) {
+        if (! isset($this->permissionClass)) {
             $this->permissionClass = app(PermissionRegistrar::class)->getPermissionClass();
         }
 
@@ -69,7 +69,7 @@ trait HasPermissions
             $query->whereHas('permissions', function ($query) use ($permissions) {
                 $query->where(function ($query) use ($permissions) {
                     foreach ($permissions as $permission) {
-                        $query->orWhere(config('permission.table_names.permissions') . '.id', $permission->id);
+                        $query->orWhere(config('permission.table_names.permissions').'.id', $permission->id);
                     }
                 });
             });
@@ -77,7 +77,7 @@ trait HasPermissions
                 $query->orWhereHas('roles', function ($query) use ($rolesWithPermissions) {
                     $query->where(function ($query) use ($rolesWithPermissions) {
                         foreach ($rolesWithPermissions as $role) {
-                            $query->orWhere(config('permission.table_names.roles') . '.id', $role->id);
+                            $query->orWhere(config('permission.table_names.roles').'.id', $role->id);
                         }
                     });
                 });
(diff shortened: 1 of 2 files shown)


michaelr0:
> How does one find himself scoping for both the `id` and the `name` of the same role?
> Right now I personally think we should keep this function as simple as it is.

Hi @musapinar,

Thanks for your reply, in theory my guess is that you wouldn't, however that was just provided as an example.
`role(['User', 'User'])` or `role([1, 1])` would also result in the same duplication of the ``select * from `roles` where ...`` query, and there is currently nothing in place to keep that under control.

While probably unlikely it is currently possible to do something like,  
`User::role(['User', 'User, 'User', 'User', 'User', 'User'])->first();`

Which would result in the execution of 6 of the following query.
``select * from `roles` where `name` = 'User' and `guard_name` = 'web' limit 1``

Resulting in the following query.
``select * from `users` where exists (select * from `roles` inner join `model_has_roles` on `roles`.`id` = `model_has_roles`.`role_id` where `users`.`id` = `model_has_roles`.`model_id` and `model_has_roles`.`model_type` = 'App\User' and (`roles`.`id` = 1 or `roles`.`id` = 1 or `roles`.`id` = 1 or `roles`.`id` = 1 or `roles`.`id` = 1 or `roles`.`id` = 1)) limit 1``

I think if it was going to happen it would likely be when passing data into the function, perhaps during some report generation effort?
Something like

`
$roles = ... data from somewhere, perhaps a loop/DB query 

User::role($roles)->with('orders')->where('status', 'active')->get();
`

While I agree that in the above scenario, the developer of the application should be doing something to handle these kinds of problems, I also think that the spatie/laravel-permission package could also help avoid these kinds of possibilities either by not executing the duplicate queries or perhaps returning an exception if a duplicate role name or id has been passed into it.

It might even be as simple as running `collect()->unique()` on the $role(s) that are not an instance of Role

Thanks @musapinar keen to hear back from you.

musapinar:
@michaelr0 I do not have strong opinion on this. I'd like to hear from other people. It just seems over-engineered to me.

michaelr0:
Closing this particular PR, will revisit this later.

PR #1333, opened 2020-01-17
Updated HasPermissions.php

musapinar:
- Removed ```@throws \Exception``` in PHPDoc because IDE was complaining about unhandled Exception when using ```getAllPermissions()```, while in the worst case scenario, as far as I can tell, it will just return an empty Collection.
- Added IDE helper for ```$permissions```

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -265,11 +265,10 @@ trait HasPermissions
 
     /**
      * Return all the permissions the model has, both directly and via roles.
-     *
-     * @throws \Exception
      */
     public function getAllPermissions(): Collection
     {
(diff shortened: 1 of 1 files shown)


src/Traits/HasPermissions.php
```diff
@@ -265,11 +265,10 @@ public function getPermissionsViaRoles(): Collection
 
     /**
      * Return all the permissions the model has, both directly and via roles.
-     *
-     * @throws \Exception
```
drbyte: Yes, this was a fragment from #926 which should have been removed during the refactor in #1039

musapinar: @drbyte Same goes for ```hasAnyPermission()``` I believe.

PR #1334, opened 2020-01-18
Updated `scopeRole()` in trait `HasRoles.php`

musapinar:
- faster/prettier than multiple ```orWhere```


drbyte:
Thanks for these!

PR #1336, opened 2020-01-18
Add scopeWhereRole to hasRoles trait

michaelr0:
Hi there,

It appears that scopeRole fires off a query to check each string or int that has been passed to it, scopeWhereRole aims to reduce these additional queries by only using the query builder.

Examples:
`App\User::role(['User', 'Admin'])->get();` will run the following queries:

``select * from `roles` where `name` = 'User' and `guard_name` = 'web' limit 1``
``select * from `roles` where `name` = `Admin' and `guard_name` = 'web' limit 1``

``select * from `users` where exists (select * from `roles` inner join `model_has_roles` on `roles`.`id` = `model_has_roles`.`role_id` where `users`.`id` = `model_has_roles`.`model_id` and `model_has_roles`.`model_type` = 'App\User' and (`roles`.`id` = 1 or `roles`.`id` = 2)) and `users`.`deleted_at` is null``

While `App\User::whereRole(['User', 'Admin'])->get();` will only run the one query:
``select * from `users` where exists (select * from `roles` inner join `model_has_roles` on `roles`.`id` = `model_has_roles`.`role_id` where `users`.`id` = `model_has_roles`.`model_id` and `model_has_roles`.`model_type` = 'App\User' and ((`roles`.`name` = 'User' and `roles`.`guard_name` = 'web') or (`roles`.`name` = 'Admin' and `roles`.`guard_name` = 'web'))) and `users`.`deleted_at` is null``

It seems that the reason for the additional ``select * from `roles` where ...`` queries in the original scopeRole function are because it does a check to see if the role exists or not, if the role is not already an `instanceof Role` then it will query to see if the int (id) or string (name) are roles that actually exist, if any do not exist then an exception will be triggered.

Personally I'd prefer that scopeRole worked in the way that this PR has for scopeWhereRole, as I'd be willing to lose the exception in favour of reducing additional queries, however I understand not everyone would want to, hence opting to instead call the function scopeWhereRole.

Thanks.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -118,13 +118,11 @@ trait HasRoles
         $guard = is_null($guard) ? $this->getDefaultGuardName() : $guard;
 
         return $query->whereHas('roles', function ($query) use ($roles, $guard) {
-            $query->where(function ($query) use ($roles, $guard) {
+            $query->where(config('permission.table_names.roles').'.guard_name', $guard);
+            return $query->where(function ($query) use ($roles) {
                 foreach ($roles as $role) {
                     $column = is_numeric($role) ? 'id' : 'name';
-                    $query->orWhere([
-                        [config('permission.table_names.roles').".{$column}", '=', $role],
-                        [config('permission.table_names.roles').'.guard_name', '=', $guard],
-                    ]);
+                    $query->orWhere(config('permission.table_names.roles').".{$column}", $role);
                 }
             });
         });
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -129,7 +129,7 @@ trait HasRoles
             });
         });
     }
-    
+
     /**
      * Assign the given role to the model.
      *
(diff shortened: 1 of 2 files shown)


musapinar:
Since the Guard is common, maybe it can be put outside the loop?
```
return $query->whereHas('roles', function (Builder $query) use ($roles, $guard) {
    $query
        ->where(function (Builder $subQuery) use ($roles) {
            foreach ($roles as $role) {
                $column = is_numeric($role) ? 'id' : 'name';
                $subQuery->orWhere(config('permission.table_names.roles').".{$column}", $role);
            }
        })
        ->where(config('permission.table_names.roles').'.guard_name', $guard);
});
```

I don't like the naming. I think it should be ```scopeRoleSomething``` to keep consistency.
Also I'm certainly not the only one already prefixing all scopes such as "scopeWith..." or "scopeWhere..." by default, so I find it confusing.

michaelr0:
Hi @musapinar,

Thanks I think it is a good idea to move the guard out of the foreach,

As for the naming, I feel whereRoleShortcut is too long, I had opted for whereRole, because it does exactly what the naming suggests.

Are there any other ideas on what might be possible name to avoid name clashing but still be simple and concise?

Thanks.

michaelr0:
Closing due to lack of interest and wanting to clean up my Git Repos.

PR #1341, opened 2020-01-21
Refactor hasAnyPermission to flatten collection instead of array

musapinar:
If accepted, proposition should be enforced with a test such as ```$this->assertFalse($this->testUser->hasAllPermissions(['edit-articles', 'edit-news'], 'edit-blog'));``` 
with User having permission for `edit-article` and `edit-news` indeed, but no permission for `edit-blog`


drbyte:
Thanks!

Tests added, and applied the same change to other newly added methods.

(Failing tests are a result of a bug in Laravel 6.15.0 and 6.15.1, not with this PR.)

PR #1342, opened 2020-01-22
Proposition to create many roles

musapinar:
Model Role:
- Updated ```create()``` so that it (optionally) returns the existing model instead of throwing a ```RoleAlreadyExists``` Exception. It still does return Exception by default, same as prior to current change proposal.
- Added ```createMany``` to create many roles in one go. Usage example : 
```
    $rolesToCreate = [
        [
            'name' => 'Super Admin',
            //'description' => 'Musa',
            //'guard_name' => 'admin',
        ],
        [
            'name' => 'Tenant',
            //'description' => 'Dirigeant de l\'Enseigne',
            //'guard_name' => 'web',
        ],
    ];
    Role::createMany($rolesToCreate);
```
or 
```
    $roleAdmin = [
        'name' => 'Super Admin',
        //'description' => 'Musa',
        //'guard_name' => 'admin',
    ];

    $roleTenant = [
        'name' => 'Tenant',
        //'description' => 'Şirket yöneticisi',
        //'guard_name' => 'web',
    ];

    $rolesToCreate = collect([$roleAdmin, $roleTenant]);
    Role::createMany($rolesToCreate);
```

It would make seeding easier.

- Minor style changes.

### DIFF
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -48,7 +48,7 @@ class Role extends Model implements RoleContract
 
         return static::query()->create($attributes);
     }
-	
+
     /**
      * Create many roles.
      *
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -39,7 +39,10 @@ class Role extends Model implements RoleContract
         ])->first();
 
         if ($existingRole) {
-            if ($withException) throw RoleAlreadyExists::named($attributes['name'], $attributes['guard_name']);
+            if ($withException) {
+                throw RoleAlreadyExists::named($attributes['name'], $attributes['guard_name']);
+            }
+
             return $existingRole;
         }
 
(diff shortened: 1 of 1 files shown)


drbyte:
@dennis-git wrote:
> You can call createMany() directly in your code, no need for it to be in the library

Agreed.
I'm not seeing the need for this.

musapinar:
> 
> 
> You can call createMany() directly in your code, no need for it to be in the library

You cannot call ```createMany()``` on the Role model directly, you would need the related model (either a ```BelongsToMany``` or a ```HasOneOrMany``` relationship).

The following is wrong:
```
(new App\Role)->createMany($roles);
```

The following is what you meant I guess, but that does not satisfy the purpose : 
```
(new App\User)->roles()->createMany($roles);
```

One could use the ```insert()``` from the ```\Database\Query\Builder``` eventually but it won't handle timestamps autocompletion (and won't even fire any model events I believe).

That being said, I can understand this might be a specific need of mine only so feel free to close this @drbyte , no worries.

PR #1356, opened 2020-02-08
Add roleorpermission blade directive

Te7a-Houdini:
this **PR** adds **@roleorpermission** directive . that do the same behavior of **RoleOrPermissionMiddleware**

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -114,6 +114,15 @@ class PermissionServiceProvider extends ServiceProvider
             $bladeCompiler->directive('endunlessrole', function () {
                 return '<?php endif; ?>';
             });
+
+            $bladeCompiler->directive('roleorpermission', function ($arguments) {
+                list($roleOrPermission, $guard) = explode(',', $arguments.',');
+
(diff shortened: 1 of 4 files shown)


drbyte:
Hi @Te7a-Houdini - thanks.

Can you describe a use-case for this?

I very rarely see a case where either/or makes sense for this when an app is structured with partials related to specific permissions.

Te7a-Houdini:
@drbyte  sure . am currently building an app for pharmacy and the admin of websites assigns dynamic roles and permissions . so i wanted this blade directive to show some dashboard based on role or assigned permission tabs like this screenshot

Screenshot from 2020-02-08 22-59-30

drbyte:
I see. 

I find a more efficient way to handle a Super-Admin role is to actually grant it all permissions, and then the app can organically just test for permissions. Less clutter and less cognitive load that way.

https://docs.spatie.be/laravel-permission/v3/basic-usage/super-admin/
https://docs.spatie.be/laravel-permission/v3/best-practices/roles-vs-permissions/

Te7a-Houdini:
@drbyte aha got it . thanks a lot for help . so i think this PR can be closed

drbyte:
Great. If a stronger use-case comes up, we can revisit it. 

Best of success with your app.

PR #1377, opened 2020-02-18
Add cascade-permission-middleware

drbyte:
Allows checking of dot-notated permissions in a cascading manner.

* checks the permissions in the following dot-notation-nested order to find first match
     * admin
     * admin.auth
     * admin.auth.users
     * admin.auth.users.modify
     * admin.auth.users.modify.create

Heavily inspired by contribution from @eduardoarandah

### DIFF
--- a/docs/basic-usage/middleware.md
+++ b/docs/basic-usage/middleware.md
@@ -15,7 +15,7 @@ Route::group(['middleware' => ['can:publish articles']], function () {
 
 ## Package Middleware
 
-This package comes with `RoleMiddleware`, `PermissionMiddleware` and `RoleOrPermissionMiddleware` middleware. You can add them inside your `app/Http/Kernel.php` file.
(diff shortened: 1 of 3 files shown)


eduardoarandah:
Awesome!

if anybody's interested, here's the idea:

https://github.com/eduardoarandah/spatie-laravel-permission-permission-middleware

https://github.com/spatie/laravel-permission/issues/440

drbyte:
I'm trying to decide whether I like the suggested name of `permissionsCascade` or whether something else is a better suggestion. (Thinking shorter is better. But still want it expressive.)

drbyte:
@benyanke

drbyte:
@eduardoarandah 
@stefanriehl

What are your thoughts on the "need" for this if #1381 is merged?
I worry that it will introduce unnecessary confusion to have this middleware doing nearly the same thing.

stefanriehl:
> I worry that it will introduce unnecessary confusion to have this middleware doing nearly the same thing.

I agree. Would be a bit confusing. Especially when you have wildcard permissions enabled.

PR #1378, opened 2020-02-18
clear class permissions var on boot

### LINKED ISSUE #1365: Swoole and permissions
Laravel version: 5.8.36
Package version: 3.6.0
PHP version: 7.4.0
laravel-swoole: 2.6.66

Part of swoole config:
```
'providers' => [
    Illuminate\Pagination\PaginationServiceProvider::class,
    Illuminate\Auth\AuthServiceProvider::class,
    App\Providers\AuthServiceProvider::class,
    Spatie\Permission\PermissionServiceProvider::class,
    Tymon\JWTAuth\Providers\LaravelServiceProvider::class,
],
```

When adding or removing a permission from a role, the permission is not reflected in the app for the user. The app has the permission set in the route. Restarting swoole after a permission change fixes things.

When debugging getPermissions() in src/PermissionRegistrar.php, it appears that $this->permissions is being reused between requests. The first request after a restart gets/puts the cache, but in all subsequent requests $this->permissions is not null and used. So even clearing the cache causes it to happily use the first copy of permissions it got.

When not running in swoole, getPermissions() hits the cache for each request.

Any idea how to remedy this?

haught:
For long running processes like swoole, PermissionRegistrar $permissions needs to be cleared when the service provider is booted.

Fixes #1365

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -105,6 +105,16 @@ class PermissionRegistrar
     }
 
     /**
+     * Clear class permissions.
+     */
+    public function clearClassPermissions()
+    {
+        $this->permissions = null;
+
+        return true;
+    }
+
+    /**
      * Get the permissions based on the passed params.
      *
      * @param array $params
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -110,8 +110,6 @@ class PermissionRegistrar
     public function clearClassPermissions()
     {
         $this->permissions = null;
-
-        return true;
     }
 
     /**
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -106,6 +106,8 @@ class PermissionRegistrar
 
     /**
      * Clear class permissions.
+     * This is only intended to be called by the PermissionServiceProvider on boot,
+     * so that long-running instances like Swoole don't keep old data in memory.
      */
     public function clearClassPermissions()
     {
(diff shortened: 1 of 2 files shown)


drbyte:
While this approach is very clear and explicit, and easy to follow, and is probably best in this case, I have a tech question (call it a learning moment about Swoole ;) ): 
would merely initializing it to `null` via `protected $permissions = null;` instead of just `protected $permissions;` be sufficient, instead of calling a method to do it?

src/PermissionRegistrar.php
```diff
@@ -104,6 +104,16 @@ public function forgetCachedPermissions()
...
     }
 
+    /**
+     * Clear class permissions.
+     */
+    public function clearClassPermissions()
+    {
+        $this->permissions = null;
+
+        return true;
```
drbyte: Is a `return` value needed here? Let's exclude it if it's not really required.

(I know `registerPermissions()` returns `bool`, so maybe you were just following that pattern, but that's an artifact that was kept as part of the public method for backward compatibility amidst other changes. 
I plan to remove it in the next major release.)

haught: The PermissionRegistrar class is only initialized once when swoole loads, so setting the protected var doesn't help. I thought the same but after I sprinkled \Log::debug all over the place I saw this behavior. 

I am thinking that this behavior might be related to the use of the singleton as I don't see it elsewhere in __construct in my app models. It seems like the singleton is reused even if the PermissionServiceProvider boots on every request.

I did add the return to match the pattern, it is not needed, I will remove.

drbyte: Thanks!

PR #1381, opened 2020-02-20
Add wildcard permissions

stefanriehl:
Allows using wildcard permissions based on the default wildcard permission implementation of [Apache Shiro](https://shiro.apache.org/permissions.html).

Enable wildcard permissions in the config file:

```php
// config/permission.php
'enable_wildcard_permission' => true,
```

When enabled, wildcard permissions offers you a flexible representation for a variety of permission schemes.

Some examples:

```php
// user can only do the actions create, update and view on both resources posts and users
$user->givePermissionTo('posts,users.create,update,view');

// user can do the actions create, update, view on any available resource
$user->givePermissionTo('*.create,update,view');

// user can do any action on posts with ids 1, 4 and 6 
$user->givePermissionTo('posts.*.1,4,6');
```

See full documentation at

```php
// docs/basic-usage/wildcard-permissions.md
```

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -93,8 +93,7 @@ return [
     'display_permission_in_exception' => false,
 
     /*
-     * By default, wildcard permission usage is disabled.
-     * When set to true, wildcard permission handling is enabled.
+     * By default wildcard permission lookups are disabled.
      */
 
     'enable_wildcard_permission' => false,
--- a/config/permission.php
+++ b/config/permission.php
@@ -92,6 +92,13 @@ return [
 
     'display_permission_in_exception' => false,
 
+    /*
+     * By default, wildcard permission usage is disabled.
+     * When set to true, wildcard permission handling is enabled.
+     */
+
+    'enable_wildcard_permission' => false,
+
     'cache' => [
 
         /*
--- a/docs/basic-usage/wildcard-permissions.md
+++ b/docs/basic-usage/wildcard-permissions.md
@@ -27,7 +27,7 @@ this is the common use-case, representing {resource}.{action}.{target}.
 
 ### Using Wildcards
 
-Each part can also contains wildcards (*). So let's say we assign the following permission to a user:
+Each part can also contain wildcards (*). So let's say we assign the following permission to a user:
 
 ```php
 $user->givePermissionTo('posts.*');
--- a/docs/basic-usage/wildcard-permissions.md
+++ b/docs/basic-usage/wildcard-permissions.md
@@ -1,0 +1,62 @@
+---
+title: Wildcard permissions
+weight: 3
+---
+
+Wildcard permissions can be enabled in the permission config file:
+
+```php
+// config/permission.php
+'enable_wildcard_permissions' => true,
+```
+
+When enabled, wildcard permissions offers you a flexible representation for a variety of permission schemes. The idea
+ behind wildcard permissions is based on the default permission implementation of 
+ [Apache Shiro](https://shiro.apache.org/permissions.html).
+
+A wildcard permission string is made of one or more parts separated by dots (.).
+
+```php
+$permission = 'posts.create.1';
+```
+
+The meaning of each part of the string depends on the application layer. 
+
+> You can use as many parts as you like. So you are not limited to the three-tiered structure, even though 
+this is the common use-case, representing {resource}.{action}.{target}.
+
+### Using Wildcards
+
+Each part can also have wildcards (*). So let's say we assign the following permission to a user:
+
+```php
+$user->givePermissionTo('posts.*');
+// is the same as
+$user->givePermissionTo('posts');
+```
+
+Everyone who is assigned to this permission will be allowed every action on posts. It is not necessary to use a 
+wildcard on the last part of the string. This is automatically assumed.
+
+```php
+// will be true
+$user->can('posts.create');
+$user->can('posts.edit');
+$user->can('posts.delete');
+``` 
+
+Besides the use of parts and wildcards, subparts can also be used. Subparts are divided with commas (,). This is a 
+powerful feature, that let's you create complex permission schemes.
+
+```php
+// user can only do the actions create, update and view on both resources posts and users
+$user->givePermissionTo('posts,users.create,update,view');
+
+// user can do the actions create, update, view on any available resource
+$user->givePermissionTo('*.create,update,view');
+
+// user can do any action on posts with ids 1, 4 and 6 
+$user->givePermissionTo('posts.*.1,4,6');
+```
+
+> As said before, the meaning of each part is determined by the application layer! So, you are free to use each part as you like. And you can use as many parts and subparts as you want.
--- a/docs/basic-usage/wildcard-permissions.md
+++ b/docs/basic-usage/wildcard-permissions.md
@@ -7,11 +7,11 @@ Wildcard permissions can be enabled in the permission config file:
 
 ```php
 // config/permission.php
-'enable_wildcard_permissions' => true,
+'enable_wildcard_permission' => true,
 ```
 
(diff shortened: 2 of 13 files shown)


tests/HasPermissionsTest.php
```diff
@@ -234,15 +234,15 @@ public function it_throws_an_exception_when_the_permission_does_not_exist()
     {
         $this->expectException(PermissionDoesNotExist::class);
 
-        $this->testUser->hasPermissionTo('does-not-exist');
+        $this->testUser->hasPermissionTo(7892);
```
drbyte: Were the two changes to a numeric permission name intentional?

drbyte:
@stefanriehl This looks great.

Question: all the wildcard lookup logic is agnostic of the user's current guard. Is that intentional?

config/permission.php
```diff
@@ -92,6 +92,13 @@
 
     'display_permission_in_exception' => false,
 
+    /*
+     * By default, wildcard permission usage is disabled.
+     * When set to true, wildcard permission handling is enabled.
```
drbyte: I'm not a fan of the word "usage" here.
Perhaps the two sentences can be combined? Or maybe "By default wildcard permission lookups are disabled." and skip the second line?

docs/basic-usage/wildcard-permissions.md
```diff
@@ -0,0 +1,62 @@
...
+wildcard on the last part of the string. This is automatically assumed.
+
+```php
+// will be true
+$user->can('posts.create');
+$user->can('posts.edit');
+$user->can('posts.delete');
+``` 
+
+Besides the use of parts and wildcards, subparts can also be used. Subparts are divided with commas (,). This is a
```
drbyte: Perhaps add a `### Sub-Parts` heading here?

src/Traits/HasPermissions.php
```diff
@@ -113,10 +114,21 @@ public function hasPermissionTo($permission, $guardName = null): bool
...
-            );
+            try {
+                $permission = $permissionClass->findByName(
+                    $permission,
+                    $guardName ?? $this->getDefaultGuardName()
+                );
+            } catch (PermissionDoesNotExist $e) {
+                // the permission as string is not present in the database
+                // so check wildcard permission if enabled in config
+                // Use sensitive default for backwards compatibility
```
drbyte: Perhaps reword for brevity and clarity? Something like:
```
// Arriving here means the permission is not found
// so check wildcard permission if enabled.
```

docs/basic-usage/wildcard-permissions.md
```diff
@@ -0,0 +1,62 @@
...
+
+```php
+// will be true
+$user->can('posts.create');
+$user->can('posts.edit');
+$user->can('posts.delete');
+``` 
+
+Besides the use of parts and wildcards, subparts can also be used. Subparts are divided with commas (,). This is a 
+powerful feature, that let's you create complex permission schemes.
```
drbyte: 2 fixes to this line:
```diff
- powerful feature, that let's you create complex permission schemes.
+ powerful feature that lets you create complex permission schemes.
```

tests/HasPermissionsTest.php
stefanriehl: My mistake. I forgot to undo that. Will change.

stefanriehl:
> @stefanriehl This looks great.
> 
> Question: all the wildcard lookup logic is agnostic of the user's current guard. Is that intentional?

Good point. You're right. Will change.

stefanriehl:
@drbyte To be honest, I am not yet completely satisfied with the implementation. Why? Well, let me show an example:

```php
public function it_can_verify_permission_instances_not_assigned_to_user()
{
    $userPermission = Permission::create(['name' => 'posts.*']);
    $permission = Permission::create(['name' => 'posts.create']);

    $user->givePermissionTo([$userPermission]);

    // all these asserts should be true, but are currently false

    $this->assertTrue($user->hasPermissionTo('posts.create'));
    // because permission exists in db and wildcard permission won't be called

    $this->assertTrue($user->hasPermissionTo($permission));
    $this->assertTrue($user->hasPermissionTo($permission->id));
    // because we can only handle strings right now
}
``` 

That doesn't seem logical to me! So, I would like to change the implementation:

> If wildcard permissions are enabled -> the application can verify permissions by id, instance or string against user permissions (if id does not exist, an exception will be thrown). 

I already did the implementation with more tests, but I would like to get your thoughts before updating.

src/Traits/HasPermissions.php
stefanriehl: pls see my new comment

drbyte:
>  I would like to change the implementation

Makes sense.

I haven't seen your proposed new approach, but perhaps it makes sense to just be binary about it: either the wildcard switch is on and that's the only matching that's used, or it's off and it defaults to the prior matching logic.

stefanriehl:
> either the wildcard switch is on and that's the only matching that's used, or it's off and it defaults to the prior matching logic.

That's what I did. I will push tomorrow. Thanks!

docs/basic-usage/wildcard-permissions.md
```diff
@@ -27,7 +27,7 @@ this is the common use-case, representing {resource}.{action}.{target}.
 
 ### Using Wildcards
 
-Each part can also have wildcards (*). So let's say we assign the following permission to a user:
+Each part can also contains wildcards (*). So let's say we assign the following permission to a user:
```
drbyte:
```diff
-contains
+contain
```

drbyte:
@stefanriehl I do like this improvement. Feels more intuitive that the switch is all-or-nothing, not hybrid.

Thoughts about performance? caching needs?

stefanriehl:
> @stefanriehl I do like this improvement. Feels more intuitive that the switch is all-or-nothing, not hybrid.
> 
> Thoughts about performance? caching needs?

Well, I did some performance tests. Wildcard permissions vs. DB permissions:

- MBP 16GB
- Local Docker with PHP 7.3.13 / Ubuntu 16.04
- sqlite
- 1 User has n permissions
- we're using permission caching
- the last inserted permission will be checked, so that we can be sure that all wildcard permissions will be tested. That is of course the worst case.

| Permissions | DB      | Wildcard |
|-------------|---------|----------|
| 10          | 0.00015 | 0.0005   |
| 25          | 0.00024 | 0.00082  |
| 50          | 0.00041 | 0.0009   |
| 100         | 0.00074 | 0.00149  |
| 250         | 0.00173 | 0.0032   |
| 500         | 0.00385 | 0.00656  |

Wildcard permission check is ~ 2.2x slower than DB permission. With the restriction that probably the permission to be checked is not the last one in the permission list. In a real scenario the wildcard query surely would be faster. Interesting: the more permissions a user has, the lower the factor.

@drbyte What do you think? Is it necessary to consider caching for wildcard permissions? In my opinion caching would only make sense, if we'll cache the positive/negative checked permissions for a user.  But this implementation is more complex I think.

drbyte:
I agree: it's probably suitable to revisit caching requirements for this feature after feedback from use-in-the-wild, where specific use-cases can be cited, so we fix actual problems instead of over-engineering something that just takes more effort to support and maintain.

drbyte:
I'm satisfied that the public signature of these changes is minimal.
The tests seem to cover all use-cases that I can think of.

Are you satisfied with the completeness of this? Ready for merge?

docs/basic-usage/wildcard-permissions.md
stefanriehl: I'll change this.

stefanriehl:
@drbyte Yes I'm fine. I'll just correct the spelling mistake.

drbyte:
Thanks!

Issue #1313, opened 2019-12-31
Check the guard of logged in user


Daniyal-Javani:
> v2 guard lookup currently looks to the model's guard_name property, with fallbacks down to finally the first available guard in your config array of guards (not the one listed as "default"). Null doesn't disable this lookup.
> So, wherever you don't specify a guard when doing a permission check, it does the lookup according to that pattern I just described.

Shouldn't it check the guard of logged in user?
like this
```
function activeGuard()
{
    foreach (array_keys(config('auth.guards')) as $guard) {

        if (auth()->guard($guard)->check()) return $guard;
    }
    return null;
}
```

How can I check based on the guard of user?

_Originally posted by @Daniyal-Javani in https://github.com/spatie/laravel-permission/issues/892#issuecomment-569771727_

drbyte:
@Daniyal-Javani I'd like to see a future version of this package honor the guard that the user has been assigned during their login/authentication. 

A PR which updates that functionality would be welcome. Ideally with tests.

musapinar:
@drbyte 
That already is the case unless I'm wrong.
https://github.com/spatie/laravel-permission/blob/master/src/Guard.php#L47

Source: https://github.com/laravel/framework/issues/22096

drbyte:
@musapinar Yes, but the next line overrides it with the model's guard, which leaves many people confused about why the package checks against a different guard than the one the user is using (they would say "it checks against the wrong guard"). 
(See several issues about guards: most are complaining of this factor.)

It's a complex change to make, because it's messing with a matrix of values created by conflating authorization with authentication. I'd rather remove all mention of guards altogether, and let the application implement guard-related rules. 
But there's a large user base depending on the guard features of this package, even though they wish it worked in a way that honoured the user's logged-in-guard with more prominence and more consistency.

drbyte:
@musapinar I think the first starting point though is reviewing where it's not working as desired, writing the tests for that, and then getting those tests to pass.

Daniyal-Javani:
OK, what about this test
https://github.com/Daniyal-Javani/laravel-permission/commit/54dde601de025813c84926a0be151c421be84b31

musapinar:
@Daniyal-Javani @drbyte 
Test looks good to me.

[...]

PR #1384, opened 2020-02-26
feat: check permission based on logged in user guard

Daniyal-Javani:
@musapinar idea to fix https://github.com/spatie/laravel-permission/issues/1313


drbyte:
@Daniyal-Javani @musapinar 
Do you see any value in some reverse tests? 
Thoughts on whether any other cases need consideration?

Daniyal-Javani:
@drbyte Should I update the document like this?
```
diff --git a/docs/basic-usage/multiple-guards.md b/docs/basic-usage/multiple-guards.md
index 69db924..b244499 100644
--- a/docs/basic-usage/multiple-guards.md
+++ b/docs/basic-usage/multiple-guards.md
@@ -28,7 +28,7 @@ To check if a user has permission for a specific guard:
 
-> **Note**: When determining whether a role/permission is valid on a given model, it chooses the guard in this order: first the `$guard_name` property of the model; then the guard in the config (through a provider); then the first-defined guard in the `auth.guards` config array; then the `auth.defaults.guard` config.
+> **Note**: When determining whether a role/permission is valid on a given model, it chooses the guard in this order: first the guard of logged in user; then the `$guard_name` property of the model; then the guard in the config (through a provider); then the first-defined guard in the `auth.guards` config array; then the `auth.defaults.guard` config.
```

drbyte:
I'm not sure the update to the docs is correct.

I think the correct new explanation is more along the lines of the following (which also feels too verbose for the docs):
- build a list of guards to match against, in this order:
- $guard_name property of the model, if any
- guard names from configured auth providers matching the model, if any
And then find the first one that matches the user's logged-in guard, if any
If not matched, return the actual guard (which the "default" has been updated to, if the user is logged in).

Keep in mind that sometimes there won't be an actual match for the logged-in user: 
- perhaps because the a permissions-check test is triggered before login, 
- it's a create-perm/role call on another model. 

It's this last case (create) that worries me. What about an Administrator, where your app has used an Admin login guard, and you're creating permissions without specifying a guard, and now it's going to default to the admin guard instead of the web guard which non-admins use?

thoresuenert:
I think this solution is way too deep into the logic.

This can (maybe) conflict with admin panels where I want to create auth user independent roles and permission.

I would prefer to pass the auth user guard name during the gate check to `checkPermissionTo` and make it configurable to avoid breaking changes.

drbyte:
> I would prefer to pass the auth user guard name during the gate check to `checkPermissionTo` 

I'd be interested in seeing an implementation of that where the current tests pass. I recall exploring that sort of approach but finding several tests failing (where refactoring those tests wasn't relevant to the code change).

PR #1394, opened 2020-03-01
Add connection

leafling:
Added a 'connection' option to the config/permissions.php file and added code to __construct function of the Role and Permission models to set the Connection to the one specified in the config...

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -29,12 +29,14 @@ return [
     ],
 
     /*
-     * When using a database other than the default connection, you can assign
-     * the connection name value. To use the default connection, simply leave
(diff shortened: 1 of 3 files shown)


drbyte:
@leafling I appreciate that you've put a lot of time into this PR.

However, we've opted not to make the package so brittle.

Two related PRs, #719 and #411 which also included tests, were closed because the custom connection overrides needs of your application can better be captured by extending the models into your app's namespace. This also leaves less code for the package maintainers to have to maintain. :)

Again, thanks for submitting the PR and sharing your code. I understand the effort that goes into a PR, and that it can be disappointing for it not to be merged.

Issue #302, opened 2017-05-20
Remove Pivot From Json


bambamboole:
When return Json with this code:

```
		$users = User::with('roles')->latest()->paginate();

		if($request->wantsJson()){
			return response()->json($users, 200);
		}
```

then it returns json like this:

```
	{
id: 1,
name: "Magdalena Bednar",
email: "admin@admin.com",
avatar: "/storage/avatar/default.svg",
created_at: "2017-05-20 14:07:17",
updated_at: "2017-05-20 14:07:17",
roles: [
{
id: 1,
name: "Admin",
created_at: "2017-05-20 14:07:16",
updated_at: "2017-05-20 14:07:16",
pivot: {
user_id: 1,
role_id: 1
}
}
]
},
{
id: 2,
name: "Lurline Muller Sr.",
email: "user@user.com",
avatar: "/storage/avatar/default.svg",
created_at: "2017-05-20 14:07:17",
updated_at: "2017-05-20 14:07:17",
roles: [
{
id: 2,
name: "User",
created_at: "2017-05-20 14:07:17",
updated_at: "2017-05-20 14:07:17",
pivot: {
user_id: 2,
role_id: 2
}
}
]
},
```

It set on the User and the Role Model `protected $hidden = ['pivot'];` but that doesn't works...
Any suggestion why I can't remove it ?

freekmurze:
This problem is not related to this package. Not sure why adding it to `hidden` doesn't work.

Best to open up an issue at [Laravel framework](https://github.com/laravel/framework).

omarjebari:
Setting $hidden = ['pivot'] on the Spatie Role and Permission models does fix this problem but they are a part of the vendor packages so shouldn't be changed. The only existing solution is to extend these models but this seems excessive given that it's often only to accomplish this.
I think it is a good idea for the package to be modified to remove 'pivot' from the data returned as i can't imagine a scenario where users want this in the returned data.
The alternative is to add extra functions like 'getAllPermssionsWithoutPivot'.
Please implement a fix for this.

drbyte:
> @omarjebari wrote: 
> remove 'pivot' from the data returned as i can't imagine a scenario where users want this in the returned data.

Let's discuss: what are the negative side-effects of setting `$hidden = ['pivot']` on the role and permission models? What breaking changes could that introduce?

omarjebari:
Thanks for your quick response.
[...]

PR #1396, opened 2020-03-02
Exclude pivot details in json output

drbyte:
Requested various times, such as in #302


drbyte:
Unfortunately this conflicts with the purpose/benefit of #1240 

Alternative: It's easy to extend the model and add the `$hidden` property inside your own app if needed. Or build a ResourceCollection to filter the desired json.

PR #1401, opened 2020-03-05
Change permissionId typehint from int to string

simoebenhida:
It does not work when using an `uuid`!

### DIFF
--- a/src/Exceptions/PermissionDoesNotExist.php
+++ b/src/Exceptions/PermissionDoesNotExist.php
@@ -11,7 +11,7 @@ class PermissionDoesNotExist extends InvalidArgumentException
         return new static("There is no permission named `{$permissionName}` for guard `{$guardName}`.");
     }
 
-    public static function withId(int $permissionId, string $guardName = '')
+    public static function withId(string $permissionId, string $guardName = '')
     {
         return new static("There is no [permission] with id `{$permissionId}` for guard `{$guardName}`.");
     }


drbyte:
@simoebenhida 
Thanks for your submission.

Unfortunately making this change will break applications which use integers.

simoebenhida:
@drbyte I'm not sure that will be the case! If you pass an `integer` when waiting for a string it will be converted to a string! but when doing the opposite it will throw an error !! So I think it will work fine!

Or we can remove the typehint if you would like!

drbyte:
Those are valid initial thoughts. But further exploration suggests it's not realistic without other large architectural changes.

Question: how are you even triggering this situation?

Facts:
- this exception is only thrown by `findById`
- the Permission and Role contracts require an `int` parameter to `findById`
- all calls to `findById` are wrapped in an `is_int()` or `is_numeric()` check, so should never fire on a string

So I'm very curious what other changes you've made in order to trigger this situation where that exception method is ever called with a non-integer value.

simoebenhida:
@drbyte Thanks for looking into this!

It is because we are throwing the exception from inside our custom Permission Model!

drbyte:
Hmmm... a custom permission model is fine, except that many parts of the code depend on resolving the model from the container based on the contract. But if your custom model doesn't fully enforce the contract (ie: the signature for `findById` expects an `int` as the first parameter, according to the contract), then you're going to run into errors.

Hmmm...

simoebenhida:
We created a new method called `findByUuid` and we were throwing the `PermissionDoesNotExist` exception! 

We just created a new custom exception for now!
 
Thanks for looking into this @drbyte appreciate it! I think we should take `uuid` in consideration for future uses! changing `id` to a string will not break anything I guess!

PR #1443, opened 2020-04-15
Cleanup the config file.

omatamix:
Making it look nice.


drbyte:
@omatamix 
Thank you for the time you took to put this PR together. 

Unfortunately, in keeping with the standard format used for config files in other Spatie packages we're not going to merge this change.

Still, thanks for your time spent!

PR #1449, opened 2020-04-21
Implement `withoutRole` and `withoutPermission` scopes

cwilby:
This PR adds an inverse scope for the `role` and `permission` scopes.


drbyte:
@cwilby Thanks for this. Clearly this has taken a lot of time to put together.

As with any PR that introduces a lot of changes, or a lot of additional code to maintain, it's helpful to understand the thinking behind the changes proposed. Hence I have some questions:

- if this (the overall PR) weren't done, was a particular limitation encountered by just cloning the existing scopeRoles and scopePermissions and making small adjustments for the inverse? (obviously it would introduce duplications, and it looks like maybe most of your changes are just to reduce that duplication: is that a correct assumption? Is there more than that?)
- why relocate `getRoleClass`? It was semantically in the same file where it was primarily used. What problem was solved by changing it?

Note: My questions do NOT reflect an objection to the proposed changes. I'm seeking to understand first before I offer challenges or suggest alterations.

Thanks again!

cwilby:
@drbyte No worries!  Happy to take part in the process.

First, your assumption regarding duplication is correct, my main driver behind this PR is reducing duplicate code where possible.  "Measure twice, cut once."

I'm equally conflicted with the dependency relationship between `HasPermissions` and `HasRoles`, I'll do my best to explain.

The `permissions` scope in `HasPermissions` should filter users based on whether they are assigned a role that includes that permission.  For this reason, the trait needs to get the class used for the Role via the `PermissionRegistrar`.

Because `HasRoles` already uses the `HasPermissions` trait, it made sense at the time to move `getRoleClass` to `HasPermissions` and allow it to be inherited from there.

But even writing that, it feels like a mess.  I'm totally open to suggestions for refactoring.

drbyte:
Thanks. I don't have a fully developed response to write yet, but I did want to mention one piece of information to add to my last point about the getRoleClass location:
While it's rare, I've seen a few cases where `HasPermissions` has been used instead of `HasRoles` in order to skip the use of roles altogether. IIRC there are a few other adjustments requires to make that work, but I'd rather not introduce something that makes doing that even more difficult.
I haven't thought it through fully yet, but perhaps your change may actually make that simpler. 
Technically that whole concept is an unrelated discussion, but it's always in the back of my mind when considering changes to these traits.

More to ponder! ...

cwilby:
No problem, thanks for your extended review 👍

cwilby:
If `HasPermissions` is used instead of `HasRoles`, this would still work, but it would perform an unnecessary query to check role permissions.

Should HasRoles extends HasPermissions in that case?

ludo237:
Pretty much the same extension that I've written two days ago, more or less, for integrating those query scopes. Great work on this one and it's a nice addition to this package

irvine48:
i am looking to how to ban an account by give them a role "banned", seen this is a good one. Is the documentation updated? i not sure how to implement this.

ludo237:
@irvine48 I think that a banned role is not the right approach to this. I think you should use a `banned_at` date and handle that in a different way. 

If you really want to stick to the banned role you can create that role without any permission but your system must leverage permissions in order to work

cwilby:
@ludo237 mind dropping a link to your package for future readers?

drbyte:
@cwilby This looks good. I'm a fan of the reduction in duplicated code as well.

While it seems to be okay in review, I"m wondering if you have any observations about cache or memory impact of the rearrangements. I'm thinking it might actually even improve things even if only very slightly. Just curious your experience, assuming you've been using this in production since preparing the PR awhile back...?

src/Traits/HasPermissions.php
```diff
@@ -101,6 +135,15 @@ protected function convertToPermissionModels($permissions): array
         }, $permissions);
     }
 
+    protected function convertPermissionsToRoleIds($permissions)
```
cwilby, line 138: @drbyte I haven't had the chance to test this in production yet so nothing concrete to report.  In theory, introducing  `$query->pluck('id')` in this method (where `array_column('id', $roles)` was previously used in the calling method to calculate permission ids for new/existing scopes) could reduce memory usage.

Issue #1454, opened 2020-04-23
Permissions for CRUD


joserick:
In cases where permissions are needed for each part of a [resource controllers](https://laravel.com/docs/7.x/controllers#resource-controllers) in base CRUD.

I understand that I have to create each one of the permissions individually using the 'create' function, defined in 'Spatie\Permission\Models\Permission'
Example:
```php
Role::create(['name' => 'admin'])->givePermissionTo(
	Permission::create(['name' => 'index articles']),
	Permission::create(['name' => 'create articles']),
	Permission::create(['name' => 'store articles']),
	Permission::create(['name' => 'show articles']),
	Permission::create(['name' => 'edit articles']),
	Permission::create(['name' => 'update articles']),
	Permission::create(['name' => 'delete articles']),

	Permission::create(['name' => 'index authors']),
	Permission::create(['name' => 'create authors']),
	...
);
```
But since it is so repetitive, I considered feasible to create another function that performs the structure mentioned above based on laravel's CRUD (resource controllers), something like this:
```php
Role::create(['name' => 'admin'])->givePermissionTo(
	Permission::createResource('articles', 'authors')
);
```
And in 'Spatie\Permission\Models\Permission' add the createResource:
```php
public static function createResource(...$permissions): Collection
{
    return collect($permissions)->flatten()->map(function ($permission) {
        if (! is_string($permission)) {
            return false;
        }

        foreach (['index', 'create', 'store', 'show', 'edit', 'update', 'delete'] as $crud) {
            $array[] = ['name' => "$crud $permission"];
        }

        return $array;
    })->collapse()->filter()->map(function ($permission) {
        return static::create($permission);
    });
}
```
With this I think it can be a help when requiring creating permissions for Resource Controllers.

PR #1455, opened 2020-04-23
Added createResource

joserick:
This new function create all permissions necessary for "resource controllers".

 More information in the 'issues' #1454


drbyte:
@joserick Thank you for the time and work you put into contributing this.

We're still trying to keep this package as lean as possible, so not planning to merge this at this time.

One way someone can implement this is to extend the model and add your code to it. This way the get the benefit of it if they need it for themselves.

Thanks again for contributing!

PR #1468, opened 2020-05-06
Add `display_role_in_exception` key with default value `false` to permission config

### LINKED ISSUE #1467: missing `permission.display_role_in_exception`?
Hi,

tl;dr: `UnauthorizedException::forRolesOrPermissions` checks for non-existent configuration key `permission.display_role_in_exception`.

I'm updating this package from version `1.18` to the latest version in one of my company's projects. I downloaded the new `permission.php` file (since `v1.18` still uses the old `laravel-permission.php` file) and was comparing the old and new config files, to see if there were any custom values I needed to copy. I also went through the (source) code to check for the usages of the config values. 
In the `Spatie\Permission\Exceptions\UnauthorizedException` class I noticed that the `forRolesOrPermissions` method contained a condition that checked if two config settings were `true`:
```
if (config('permission.display_permission_in_exception') && config('permission.display_role_in_exception'))
```
I found the first config key in `permission.php`, but not the second. Did i stumble upon a bug/missing config key, or am I missing something here?

Kind regards,

Yoeri

Yoeriwalstra:
The `UnauthorizedException::forRolesOrPermissions` checks for the non-existent configuration key `display_role_in_exception`. This PR adds that config key with a default value of `false` to `config/permission.php`.

Fixes #1467

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -85,12 +85,20 @@ return [
     ],
 
     /*
-     * When set to true, the required permission/role names are added to the exception
+     * When set to true, the required permission names are added to the exception
      * message. This could be considered an information leak in some contexts, so
      * the default setting is false here for optimum safety.
      */
 
     'display_permission_in_exception' => false,
+
+    /*
+     * When set to true, the required role names are added to the exception
+     * message. This could be considered an information leak in some contexts, so
+     * the default setting is false here for optimum safety.
+     */
+
+    'display_role_in_exception' => false,
 
     /*
      * By default wildcard permission lookups are disabled.


drbyte:
@Yoeriwalstra Thanks!

The code change is good.

Before I merge ... One tip: if you edit the first post above, where you put your initial text when generating the PR, and add the word "Fixes " followed by the number of the Issue it solves, then when the PR is merged it'll automatically close the associated issue. You'll also see the cross-linking show up on both sides. It's one way you can help maintainers keep things tidied up.
(As a maintainer I can do that edit myself, but I'll let you do it so you can see how it works.)

Yoeriwalstra:
I think this is what you were looking for. I don't see a reference to the PR on the issue, or vice versa, though. I'll just assume the issue gets closed as soon as the PR is merged.

drbyte:
No, this is what I meant:

Image 2020-05-07 at 3 01 09 PM

PR #1484, opened 2020-05-24
Speed up permissions cache lookups

davejamesmiller:
Uses `$collection->first(...)` instead of `$collection->where(...)->where(...)->first()` to avoid creating two temporary arrays and stop as soon as a match is found.

In my project - PHP 7.4, Xdebug and Debugbar enabled, and 380 permissions - it reduced the `Permission::findByName()` time from 48.54ms average to 6.19ms. On a page with 48 permissions checks, that reduced the total find time from 2330ms to 297ms, and the overall page load **from 3.4 seconds to 1.1 seconds**.

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -36,7 +36,7 @@ class Permission extends Model implements PermissionContract
     {
         $attributes['guard_name'] = $attributes['guard_name'] ?? Guard::getDefaultName(static::class);
 
-        $permission = static::getPermissions(['name' => $attributes['name'], 'guard_name' => $attributes['guard_name']])->first();
+        $permission = static::getPermission(['name' => $attributes['name'], 'guard_name' => $attributes['guard_name']]);
 
         if ($permission) {
             throw PermissionAlreadyExists::create($attributes['name'], $attributes['guard_name']);
@@ -85,7 +85,7 @@ class Permission extends Model implements PermissionContract
     public static function findByName(string $name, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
+        $permission = static::getPermission(['name' => $name, 'guard_name' => $guardName]);
         if (! $permission) {
             throw PermissionDoesNotExist::create($name, $guardName);
         }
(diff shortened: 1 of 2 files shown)


davejamesmiller:
> If you feel this pull request deserves some attention from my human colleagues feel free to reopen it.

I don't have permission to reopen it? But yes I think it's still an issue.

drbyte:
@davejamesmiller 
Thanks for this.

After merging with recent updates and exploring your proposed changes, and reflecting on prior discussions around the `clone` and other things, I wonder if the same results can be accomplished, without negative side effect, with reducing this whole PR to just the following combined change?

```diff
   public function getPermissions(array $params = []): Collection
    {
        if ($this->permissions === null) {
            $this->permissions = $this->cache->remember(self::$cacheKey, self::$cacheExpirationTime, function () {
                return $this->getPermissionClass()
                    ->with('roles')
                    ->get();
            });
        }
        
-        $permissions = clone $this->permissions;
-        
-        foreach ($params as $attr => $value) {
-            $permissions = $permissions->where($attr, $value);
-        }
-        
-        return $permissions;

+        return $this->permissions->filter(static function ($permission) use ($params) {
+            foreach ($params as $attr => $value) {
+                if ($permission->getAttribute($attr) != $value) {
+                    return false;
+                }
+            }
+
+            return true;
+        });
    }
```

Curious your thoughts, and whether it still improves your benchmark results similarly to your original results.

davejamesmiller:
Hi @drbyte 

Thanks for taking a look.

Updated benchmarks for a full page load on the same app as above are:

- `master` - 3.3 seconds
- `speed-up-cache` (my PR) - 1.0 seconds
- [`speed-up-cache-2`](https://github.com/davejamesmiller/laravel-permission/commit/102a37d90bbf1520af3d05b859eea6227cf71899) (your suggestion) - 1.7 seconds

So I think that would be an improvement, but switching to `first()` would still be more efficient by a large enough margin to make it worthwhile.

drbyte:
Thanks for the updated benchmarks. That's helpful.

> but switching to `first()` would still be more efficient

Maybe I've been staring at code too long, but how are these different?

- getPermissions()->first()
 ... where getPermissions is returning the same collection that getPermission() refers to

- getPermission()
  ... where it calls ->first() on the same collection

davejamesmiller:
`getPermissions()` calls `filter()`, which scans through the entire array and returns a new collection (probably with 0 or 1 elements), then calls `first()` on that. So PHP still has to allocate a new array and a new object for every call.

`getPermission()` calls `first()` on the original collection, which stops as soon as a match is found and never creates a new collection or array.

davejamesmiller:
@drbyte Would you consider revisiting this? I still think it's needed. For now I have to override the class in every project. Thanks.

PS The bot says "feel free to reopen it" but we don't have permission to do that.

drbyte:
@davejamesmiller It's still on my radar. I've been meaning to revisit it in conjunction with other PRs and issues talking about cache and performance. My worry is that the various approaches don't address each others' problems and might even work against each other.  Given that I haven't run into big issues with these myself I haven't staged an environment that addresses them all.

davejamesmiller:
OK, thanks 👍🏻

drbyte:
@davejamesmiller much of your suggestion was incorporated into #1799 which has been merged and released in `4.3.0`.
Thanks for your contribution!

PR #1516, opened 2020-06-25
Feature/use auth guard for permission checks

### LINKED ISSUE #1515: Multiple guards are independent from the Auth::guard() which is used for login?
I wrap my head around the multiple guard usage and I am confused.
I thought initially I have to duplicate all permissions for both guards, but as long as I use the Same Model for a Guard it doesn't matter.

When I have the base setup with multiple guards as follows:

```php
// config/auth.php
'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],

        'api' => [
            'driver' => 'passport',
            'provider' => 'users',
            'hash' => false,
        ],
    ],
```

```php
// App\User
class user extends Authenticated
{
  public $guard_name = 'web';
}
```

This package don't care which guard I used to login.
In this scenario the `web` guard is used for roles and permissions in both scenarios:

- logged in by api guard
- logged in by web guard

Am I right?

PS: that's feels a bit odd :D we use this package several years.

thoresuenert:
Closes #1515 

As I mentioned in #1384 the solution provided by that PR is way to deep into the guard resolving strategy.

My solution to the problem starts on the gate level.
The package already supports to specify the guard name when we call `checkPermissionTo` on guard level.

For Backwards compatibility I added a feature flag to the config file.

**One Concern**
When the feature flag defaults to true there is one failing test:
```
Spatie\Permission\Test\MiddlewareTest::a_user_cannot_access_a_route_protected_by_the_permission_middleware_of_a_different_guard
```
To be honest I don't know what this test want to provide.
I don't understand which part from the guard guessing logic kicks in.
the user have only one permission so the guard_name is irrelevant for the test case.

To make the test pass with this new feature we just need to login the admin with the admin guard.
```
Auth::guard('admin')->setUser($this->testAdmin);
Auth::guard('web')->setUser($this->testUser);
```
The TestCase sets the guard providers as follows:
```
$app['config']->set('auth.guards.admin', ['driver' => 'session', 'provider' => 'admins']);
$app['config']->set('auth.providers.admins', ['driver' => 'eloquent', 'model' => Admin::class]);
```

So in the real world there is no way that the `$this->testAdmin` is logged in with the guard `web`.

### DIFF
--- a/src/helpers.php
+++ b/src/helpers.php
@@ -18,3 +18,22 @@ if (! function_exists('getModelForGuard')) {
             })->get($guard);
     }
 }
+
+if (! function_exists('currentAuthGuard')) {
+    /**
+     *
+     * @return string|null
+     */
+    function currentAuthGuard()
+    {
+        if(! config('permission.enable_dynamic_auth_guard_checks')) {
+            return null;
+        }
+
+        foreach (array_keys(config('auth.guards')) as $guard) {
+
+            if (auth()->guard($guard)->check()) return $guard;
+        }
+        return null;
+    }
+}
--- a/config/permission.php
+++ b/config/permission.php
@@ -106,6 +106,13 @@ return [
 
     'enable_wildcard_permission' => false,
 
+    /*
+     * By default the guard for permission checks is auto discovered
+     * if enable all permission checks will use the auth guard used for login
+     */
+
+    'enable_dynamic_auth_guard_checks' => false,
+
     'cache' => [
 
         /*
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -87,7 +87,7 @@ class PermissionRegistrar
     {
         app(Gate::class)->before(function (Authorizable $user, string $ability) {
             if (method_exists($user, 'checkPermissionTo')) {
-                return $user->checkPermissionTo($ability) ?: null;
+                return $user->checkPermissionTo($ability, currentAuthGuard()) ?: null;
             }
         });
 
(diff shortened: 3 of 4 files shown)


src/helpers.php
```diff
@@ -18,3 +18,22 @@ function getModelForGuard(string $guard)
             })->get($guard);
     }
 }
+
+if (! function_exists('currentAuthGuard')) {
+    /**
+     *
+     * @return string|null
+     */
+    function currentAuthGuard()
```
drbyte, line 27: Can this be done without adding a global scope helper function?

src/helpers.php
```diff
@@ -18,3 +18,22 @@ function getModelForGuard(string $guard)
...
+        if(! config('permission.enable_dynamic_auth_guard_checks')) {
+            return null;
+        }
+
+        foreach (array_keys(config('auth.guards')) as $guard) {
+
+            if (auth()->guard($guard)->check()) return $guard;
+        }
+        return null;
+    }
```
drbyte, line 38: If I recall correctly the reason why the Guard model's lookup function checks not only for the array_key but also for a valid (defined) provider is because Lumen does things somewhat differently from Laravel here, and without it Lumen wasn't compatible.

src/helpers.php
thoresuenert: A reasonable place could be `Spatie\Permission\Guard` as a static method.
Don't know if that is a good place to have a dependancy to the `Auth` Facade/Class.

src/helpers.php
thoresuenert: To be honest I copied that check from an other issue.
I didn't find an elegant way to get the active guard, is laravel missing a piece here?

drbyte: > I didn't find an elegant way to get the active guard, is laravel missing a piece here?

The only way is to read config(auth.guard.default): Laravel sets (replaces) it with the name of the active guard when the guard changes.

PR #1521, opened 2020-07-04
PermissionRegistrar::initializeCache public to easily reinitialize cache store

ivalkenburg:
Because a reference to the cache store gets stored during application boot makes it cumbersome to change for example any cache prefixes. Having public access to PermissionRegistrar::initializeCache() method makes it possible to easily manually reinitialize the stored cache store if you wish to use this package for example with multitenancy.

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -49,7 +49,7 @@ class PermissionRegistrar
         $this->initializeCache();
     }
 
-    protected function initializeCache()
+    public function initializeCache()
     {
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks for this. I see your point. 
And yet I wonder why do various multi-tenancy packages not need to do this?

ivalkenburg:
Well, in my situation i wish to use this package in conjunction with spatie's multitenancy package. Using the ability to change the cache prefix based on which tenant is active to scope the cache. But because permissions package stores a reference of the existing cache store with the old prefix from application boot having access to the initializeCache method lets me get the correct store which was changed.

PR #1523, opened 2020-07-06
Add @throws to migration docblocks

MinaseTaki:
Add @throws to docs.


drbyte:
Thanks for your PR.

While it is appreciated, we're not usually adding this for things that are generated from Laravel's stubs.

PR #1525, opened 2020-07-10
Addes description to roles/permissions table

hmrneves1:
Hello,

IMO, I think the package is missing the column description on roles and permissions table. It's always good to give a brief description, specially to a permission.

Regards


freekmurze:
Add this to the migration of your own project.

PR #1533, opened 2020-07-16
Update Spatie Logo

mdrathik:
here was a broken image link, just updated Spatie Logo to fix that

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 <table>
    <tr>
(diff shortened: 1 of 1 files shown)


drbyte:
@freekmurze do you have an updated `auth0` logo you want used here since it's part of the Sponsors block?

AlexVanderbist:
Thanks! Hadn't noticed it had gone missing yet 😅

PR #1538, opened 2020-07-26
Adapting creation commands to the models

JackieDo:
**Why need this feature?**

- Your package allows customizing (extending) the Role and Permission models.
- Your package allows changes to the database table structure (via migrations).

But once the database tables is changed, the creation commands (`php artisan permission:create-xxx`) will not work exactly because there are some fields in the database that require input but your commands doesn't do it. This wastes the value of your two commands.

This patch make creation commands adapt to the Role and Permission models when the database table structure is changed.

**Principles and purposes of this patch**

- Try not to change the methods in your package.
- Make your command arguments and options clearer (the `guard` and `permissions` arguments should be command options).
- The new fields in the database table will be automatically converted into arguments and options for your commands.

Because the contributor terms require testing before making a pull request, so I changed your `CommandTest.php` file to be compatible with the new way the commands work. I also made some changes to documentations to re-guide how to work with the new creation commands.


bhushan:
+

bhushan:
plus one

freekmurze:
Thanks for your work on this. For now, I'm going to pass on this to keep the package as simple as it can be.

Issue #495, opened 2017-10-08
Remove morphToMany return type for users() method at least?


ziming:
A project I'm currently working on had each user's username as their primary key field in their users table instead of an integer id. I tried to override the users().etc method but because a return type is specified (morphToMany), I couldn't return something else. I don't have the authority to modify the users table either.

So here's a suggestion to either remove the strict return type or upgrade the return type to belongsToMany at least, which is a superclass of morphToMany (I don't think you used any methods specific to morphToMany). Removing the return type would make things much more flexible too.

My current way is to create my own role/permission classes and wrap yours like an adapter for all the other methods. But it could be simpler. I could just extend your class if the morphToMany return type is not specified. Sure I could specify a method called myUsers() too but that also means I have a non working users() method I must always remember not to call.

AlexVanderbist:
Hi, what exactly is stopping you from changing the migration to allows strings as the `model_id` in all tables and extending the `users()`, `permsissions()` and `roles()` methods to use `username` as the `relatedKey`? 

For example when extending the `Permission` class and overwriting the `users()` method:
```php
    public function users(): MorphToMany
    {
        return $this->morphedByMany(
            getModelForGuard($this->attributes['guard_name']),
            'model',
            config('permission.table_names.model_has_permissions'),
            'permission_id',
            'model_id',
            'id',
            'username'
        );
    }
```

AlexVanderbist:
Closing this to due to inactivity. I'll update the return type to something sensible in v3 of the package.

ziming:
Ah, the project uses posgreSQL. Comparing an integer and a string will result in an error so if model_id is of string type and another table id field is an auto increment integer type. Trying to do comparison or joins will get error (model_id = something.id). It doesn't do loose comparison.

I settled the issue in the end currently by adding a separate field called username. Then since I can't override users() as it defined MorphToMany, I create my own model class and wrap around Spatie Role and Permission class like an adapter. then write my own users method but return type set to BelongsToMany. so model_type is App\User and username is the user's username while model_id is null for that record.

The methods on BelongsToMany and MorphToMany are the same I believe except that when you call attach(), save() on a MorphToMany, it fills in model_id and model_type. But that should be transparent behind the scene as MorphToMany is a sub class of BelongsToMany.

Of course I could just not name that method users() too and leave a note on the project not to call $role/permission->users().

Issue #395, opened 2017-08-03
Subclasses with global scope


OliverZiegler:
Trying to achieve the same as stated in #41.

Unfortunately the former BelongsToMany now is a MorpToMany relation.
And as we try to get the roles from a subclass will try to find the wrong morphType.

Any advice how to get this to work?

OliverZiegler:
So with some more googling I found:

```php
public function getMorphClass()
{
    return User::class;
}
```

with which it works perfectly....

PR #1545, opened 2020-08-11
Change users relationship type to BelongsToMany

WimWidgets:
Currently the relationship between users and both roles and permissions is of type `MorphToMany`, which is great if you need that flexibility. When you know you only have one kind of model (probably `User`), you don't need all that polymorphism.

I've tried to extend some classes to remove the morph, but this impossible because the return type can't be changed. However, the `MorphToMany` type is an extension of `BelongsToMany`. So in this PR I've changed the return type to that.

All the tests still pass and this should not introduce any breaking changes.

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -8,7 +8,6 @@ use Spatie\Permission\Traits\HasRoles;
 use Illuminate\Database\Eloquent\Model;
 use Spatie\Permission\PermissionRegistrar;
 use Spatie\Permission\Traits\RefreshesPermissionCache;
-use Illuminate\Database\Eloquent\Relations\MorphToMany;
 use Spatie\Permission\Exceptions\PermissionDoesNotExist;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Spatie\Permission\Exceptions\PermissionAlreadyExists;
@@ -59,7 +58,7 @@ class Permission extends Model implements PermissionContract
     /**
      * A permission belongs to some users of the model associated with its guard.
      */
-    public function users(): MorphToMany
+    public function users(): BelongsToMany
     {
         return $this->morphedByMany(
             getModelForGuard($this->attributes['guard_name']),
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -10,7 +10,6 @@ use Spatie\Permission\Exceptions\GuardDoesNotMatch;
 use Spatie\Permission\Exceptions\RoleAlreadyExists;
 use Spatie\Permission\Contracts\Role as RoleContract;
 use Spatie\Permission\Traits\RefreshesPermissionCache;
-use Illuminate\Database\Eloquent\Relations\MorphToMany;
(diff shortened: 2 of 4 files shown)


drbyte:
> Currently the relationship between users and both roles and permissions is of type `MorphToMany`, which is great if you need that flexibility. When you know you only have one kind of model (probably `User`), you don't need all that polymorphism.

Yes, version 1 of this package doesn't have all that polymorphism. I've unofficially been keeping it functional into newer Laravel versions, but not backporting v2 features to it. That unofficial upkeep may have to end soon since v1 exists primarily to support older Laravel versions, which are beholden to older PHP and other dependencies.
Anyway, it's currently still an option if you find the polymorphism in later versions to be too complex for your needs.

drbyte:
Notes:

There was some discussion about this same change in #495

And a related comment in #395

WimWidgets:
> Yes, version 1 of this package doesn't have all that polymorphism. I've unofficially been keeping it functional into newer Laravel versions, but not backporting v2 features to it. That unofficial upkeep may have to end soon since v1 exists primarily to support older Laravel versions, which are beholden to older PHP and other dependencies.
> Anyway, it's currently still an option if you find the polymorphism in later versions to be too complex for your needs.

If I don't want to use polymorphism I should use version 1, but support for that version will drop soon and it is missing features from version 2. So is that even a viable option?

> There was some discussion about this same change in #495
>
> [AlexVanderbist commented on Oct 10, 2017][1]
> I agree removing the strict return type would make things more flexible but it would also be a breaking change (I think). I'd rather address this in v3 of the package (#479).

There is agreement about the flexibility, but also fear of a possible breaking change (_"I think"_). I personally don't _think_ it's a breaking change due to inheritance of the relationship types. Either way, the test-suite should reflect the actual situation.

> And a related comment in #395

I don't quite see how this is related. I want to be able to remove the polymorphism _altogether_ if I so chose to.

---

Just to be clear, this is a nice to have. I'm using the package as is and that's fine. I just like to remove clutter everywhere I can when I know I don't need it. Removing the polymorphism would eliminate a redundant `_type` column and add a useful foreign key for example.

[1]: https://github.com/spatie/laravel-permission/issues/495#issuecomment-335498157

drbyte:
I agree. I'd like to merge this.

I'm trying to find my notes where I opted "not" to do this because it broke something. I just can't recall whether it triggered a test-suite failure (which would be useful to see), or if it was in a specific app (which is harder to track down, replicate, debug, and find out "why" and whether the problem was very app-specific or exposed a dependency issue present in larger applications that's not obvious within this package's test suite).

I guess merging it would quickly reveal situations where it's a problem!  :D 

I hope to do some more digging by the end of the week, and hopefully merge it then.

WimWidgets:
At this point the test-suite still passes, so the change doesn't seem to break the package as defined by its rules. It could be that a specific implementation will be broken. If that's the case and that's something you feel the package should support, than the test-suite should reflect that.

PR #1574, opened 2020-09-25
Add HasFactory trait to models.

yazeedobaid:
This PR adds the new `HasFactory` trait to the `Role` and `Permission` models. This will enable the creation of factory classes on consuming projects that use built-in models.

In the project that I'm working on, we have defined a `RoleFactory` and a `PermissionFactory` for the built-in models. While updating to Laravel 8 and converting the factories for the two models to classes, we discover that the models don't use the HasFactory trait. Adding this trait will enable the usage of factory classes on models using the factory method from the trait.

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\Permission\Models;
 
+use Illuminate\Database\Eloquent\Factories\HasFactory;
 use Spatie\Permission\Guard;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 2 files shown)


drbyte:
This breaks Laravel 7 compatibility.

yazeedobaid:
@drbyte you are right, I didn't think about this at the first. I guess we should go with creating new custom models for Permission and Role models and add that trait to them. But should this be mentioned in the docs in someway?

drbyte:
Yes, I've added a note to the docs.

PR #1591, opened 2020-10-14

EminRahmanov:
Role and permission names are unique with guard name.


drbyte:
Thanks.

This is a duplicate of 1261

PR #1608, opened 2020-11-04
Default guard should be config's default guard instead of first key from all available drivers.

A-H-Pooladvand:
After hours of searching to know that why when I set everything in config and database to `api` but my code doesn't work I found that in `getDefaultName` method instead of getting default guard from `defaults` we're getting it from first key of `guards` which is weird.

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -50,6 +50,6 @@ class Guard
     {
         $default = config('auth.defaults.guard');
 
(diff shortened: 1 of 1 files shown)


drbyte:
This change breaks the tests, and backward compatibility.

bhaveshdaswani93:
When working for default guard it should consider that one
I change my default guard to API and now on assign role it is giving an error role not found

PR #1615, opened 2020-11-09
run test suite with PHP8

Nielsvanpach:
- As discussed with @freekmurze , we can drop support on L5.8
- dropped deprecated `--no-suggest` option in composer v2

### DIFF
--- a/.github/workflows/run-tests-L7.yml
+++ b/.github/workflows/run-tests-L7.yml
@@ -9,16 +9,14 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        php: [7.4, 7.3, 7.2]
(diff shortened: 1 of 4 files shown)


drbyte (approved):

drbyte:
@Nielsvanpach 
Thanks for this.

I'm leaning towards releasing this as a 4.0 for the package, so that the 3.x will always work for Laravel 5.8 sites.

Plus, a couple open PRs related to guards are best-suited for a major version release, so I'm inclined to merge them into that before tagging it.

So ... while I'm merging this into master, I'm not planning to tag a release for it just yet. People wanting to test this with their app can use `dev-master`.

PR #1624, opened 2020-11-16
add guard_name constraints to the relations

ctf0:
this is to make sure each model gets his correct roles/permissions according to each `guard_name` prop, otherwise fall back to the default which is `web`

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -48,7 +48,7 @@ trait HasPermissions
             config('permission.table_names.model_has_permissions'),
             config('permission.column_names.model_morph_key'),
             'permission_id'
-        );
+        )->where('guard_name', $this->getDefaultGuardName());
     }
 
     /**
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -45,7 +45,7 @@ trait HasRoles
             config('permission.table_names.model_has_roles'),
             config('permission.column_names.model_morph_key'),
             'role_id'
-        );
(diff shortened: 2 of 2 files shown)


drbyte:
Should this filter be part of the userland app code instead?
What if someone actually wants to get *all* permissions back, regardless of guard?

ctf0:
if am missing the point plz advice, 

but doesnt the user assign roles & perms according to his gate otherwise you would get errors later on ? 
if yes then getting all the roles/perms is done through the model `Role/Permission::get()`.

drbyte:
> if am missing the point plz advice,
> 
> but doesnt the user assign roles & perms according to his gate "otherwise you would get errors later on" ?
> if yes then getting all the roles/perms is done through the model `Role/Permission::get()`.

I'm not sure I understand

drbyte:
Could you explain what you were doing in your app that made you decide this code change was necessary? 

What is the problem that this is trying to solve?

ctf0:
> What is the problem that this is trying to solve?

having models with different guards that require roles & perms ex.

| model | guard|
|------|-------|
| User | web|
| Agent | agent|
| Client | client|

so when trying to assign a role to one of them i was getting duplicate list of roles & permissions, because i have some roles & perms that the same names under each guard.

but with the constraints, i can 
- attach/detach the correct relations with the correct guard
- get the correct list of roles/permissions without duplicates or mismatched guard
- create accessors with ease

ctf0:
after more thinking, i think we dont really need it, its up to dev to show the correct roles & permissions according to the user guard name to select from.

Issue #1156, opened 2019-07-11
Tip: Wildcard guard for permissions and roles


mikemand:
I know there are already a lot of (closed) issues about the guards, but I thought I'd document my solution for v2. Hopefully, v3 is in the works and will either include a wildcard ability or remove guards altogether. :smiley:

I have a use case similar to #968, where I have to support the same permissions for `web` and `api` guards equally.

Here's what I've done:

1. Extend the default Role and Permission models, and add a property called `$guard_name`. It can be set to whatever you want the default "guard" to be called. Don't forget to update the config to use your new models.
	```php
	<?php
	
	namespace App;
	
	class Role extends \Spatie\Permission\Models\Role
	{
	    protected $guard_name = '*';
	}
	```
	```php
	<?php
	
	namespace App;
	
	class Permission extends \Spatie\Permission\Models\Permission
	{
	    protected $guard_name = '*';
	}
	```
1. In your model(s) that `HasRoles` or `HasPermissions`, overwrite the `hasPermissionTo` and `getDefaultGuardName` methods:
	```php
	<?php
	
	namespace App;
	
	class User extends ...
	{
		use ...HasRoles...;
	
		public function hasPermissionTo($permission, $guardName = '*'): bool
		{
			// Since this method comes from a trait,
			// you cannot simply `parent::hasPermissionTo($permission, '*')`.
			// You'll have to copy the entire body of the method into yours.
			// Just replace '*' with the "guard" name from above.
		}

		protected function getDefaultGuardName(): string
	    {
	        return '*';
	    }
	}
	```
1. Create all of your permissions and roles with `'guard_name' => '*'` (replace `'*'` with whatever you want the default "guard" to be.)

So far, this seems to be working. I have not seen any downsides, but if @drbyte or someone else knows any pitfalls or gotchas, please let me know.

drbyte:
Thanks @mikemand -- appreciate your sharing this simple workaround!

The guard-agnostic `v3` has been percolating on the back-burner for awhile, but hope to give it some love and a solid push in a couple months.

drbyte:
@mikemand 
In a future major version, I'm thinking perhaps we can allow (maybe via a configuration option) a form of wildcard guard support as default, but for those who really require the specificity they could enable the additional restrictions.

Given that you've been using a wildcarding approach for awhile now, do you have any suggestions for implementation? Any gotchas you've run into? Any interest in participating with coding?

mikemand:
Hi @drbyte,

[...]

PR #1628, opened 2020-11-20
allow checking for role_or_permission with multiple guards

ctf0:
atm in multi guards setup we cant check for the same role/perm with different guards ex.
```
// wont work as the middleware will bail on first error
'middleware' => [
     'role_or_permission:user|user.edit,web',
     'role_or_permission:user|user.edit,agent'
],
```

[nor its documented](https://spatie.be/docs/laravel-permission/v3/basic-usage/multiple-guards) on how to use the middleware in this scenarios.

so with the new changes you only need the role / perm ex.
```
'middleware' => 'role_or_permission:user|user.edit',
```

and the mw will take care of the rest as the guard is already resolved when the user logs in.

also i've read https://github.com/spatie/laravel-permission/issues/1156 over & over but cant find a way around for using the middleware with that setup.

if all good i can update the docs as well 🥇 .

### DIFF
--- a/src/Middlewares/RoleOrPermissionMiddleware.php
+++ b/src/Middlewares/RoleOrPermissionMiddleware.php
@@ -3,14 +3,15 @@
 namespace Spatie\Permission\Middlewares;
 
 use Closure;
-use Illuminate\Support\Facades\Auth;
 use Spatie\Permission\Exceptions\UnauthorizedException;
 
 class RoleOrPermissionMiddleware
 {
-    public function handle($request, Closure $next, $roleOrPermission, $guard = null)
+    public function handle($request, Closure $next, $roleOrPermission)
     {
-        if (Auth::guard($guard)->guest()) {
+        $user = $request->user();
+
+        if (!$user) {
             throw UnauthorizedException::notLoggedIn();
         }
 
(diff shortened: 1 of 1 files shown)


drbyte:
It seems your PR breaks the tests.

Also, won't your proposed changes break the functionality added in https://github.com/spatie/laravel-permission/pull/1565 ?

ctf0:
> won't your proposed changes break the functionality added

correct but honestly am not sure why do we need it
- with the new change u can check correctly against the logged in user whether u r in `multi tenant, guards, or single` usage
- the old functionality is only required if u want to check against a specific guard which is useless because u always want to check against the logged in user not to run tests.

as for tests, i will try to fix them.

drbyte:
> > won't your proposed changes break the functionality added by @lloricode in #1565 ?
> 
> correct but honestly am not sure why do we need it
> 
> * with the new change u can check correctly against the logged in user whether u r in `multi tenant, guards, or single` usage
> * the old functionality is only required if u want to check against a specific guard which is useless because u always want to check against the logged in user.
> * the PR doesnt specify what he is trying to solve

@lloricode thoughts? Will this PR still allow what you were trying to do?

lloricode:
> > > won't your proposed changes break the functionality added by @lloricode in #1565 ?
> > 
> > 
> > correct but honestly am not sure why do we need it
> > 
> > * with the new change u can check correctly against the logged in user whether u r in `multi tenant, guards, or single` usage
> > * the old functionality is only required if u want to check against a specific guard which is useless because u always want to check against the logged in user.
> > * the PR doesnt specify what he is trying to solve
> 
> @lloricode thoughts? Will this PR still allow what you were trying to do?

looks good to me, we can update our project if some changes made

lloricode (approved):

Issue #1632, opened 2020-11-30
How to use permission from model_has_permissions only?


ThunderBirdsX3:
My system can assign multiple role, And each Role has default permissions.

But I let user can choose, Which permission from Role they want to assign.

So, How to give multiple role to user, But use permissions only from `model_has_permissions` not `role_has_permissions`.

ThunderBirdsX3:
Should I set this code directly in User Model?
```
    /**
     * Determine if the model has, via roles, the given permission.
     *
     * @param \Spatie\Permission\Contracts\Permission $permission
     *
     * @return bool
     */
    protected function hasPermissionViaRole(Permission $permission): bool
    {
        return false;
    }
```

drbyte:
Closed in #1633

PR #1633, opened 2020-11-30
Add check permission via role config.

ThunderBirdsX3:
From #1632 

I add check permission via role config, In case  you want to give and check permissions from `model_has_permissions` only.
Not to check permissions from Role.

### ** Not support when check permission directly from role.

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -108,7 +108,7 @@ return [
 
     /*
      * When you use Model has multiple role.
-     * But you want to give permissions directly to Model, Not use default permissions from role.
+     * But you want to give and check permissions directly to Model, Not use default permissions from role.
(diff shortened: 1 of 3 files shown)


drbyte:
This appears to be a duplication of the functionality already provided by `hasDirectPermission()`.

ThunderBirdsX3:
> `hasDirectPermission`

Yes but it also fallback to `hasPermissionViaRole()` if it was false.

drbyte:
Maybe I'm not understanding the use-case fully, but it seems like this is a situation that is very specific to your app.

It adds more complexity to this package, when the package aims to be less complex so that app-specific needs can be easily implemented in userland.

PR #1669, opened 2021-01-14
Correctly finds the guard name when using users method

### LINKED ISSUE #1666: When roles relation is fetched guard_name property is not applied.
```
"message": "Undefined index: guard_name",
    "exception": "ErrorException",
    "file": "/Users/bhushan/Code/meds/meds-be/vendor/spatie/laravel-permission/src/Models/Role.php",
    "line": 66,
```

When we use $guard_name property in User model then we don't have access to it when we fetch roles relations with some other model. 

Ex. 
$company->roles->get(['roles.id', 'roles.name']);

I will share PR very soon :)

bhushan:
fixes #1666 

Correctly finds the guard name when using users method

### DIFF
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -61,7 +61,7 @@ class Role extends Model implements RoleContract
     public function users(): BelongsToMany
     {
         return $this->morphedByMany(
-            getModelForGuard($this->attributes['guard_name']),
+            getModelForGuard($this->attributes['guard_name'] ?? Guard::getDefaultName(static::class)),
             'model',
             config('permission.table_names.model_has_roles'),
             'role_id',


bhushan:
?

Gummibeer:
@drbyte could you check this PR? It was closed by bot because of inactivity. But as it's a fix for a reported bug I think that it should get attention.

drbyte:
@bhushan can you provide a test that demonstrates how this code change fixes the situation you're addressing?

bhushan:
@drbyte 
TBH 😅, I am not understanding my own words and unfortunately I don't have access to project where i was using this package due to client change. I will debug it more and if found anything, Will update PR with test..

drbyte:
@Gummibeer Have you seen this bug in a project?
While I'm not opposed to the fix, I'd prefer to see it with an accompanying test that explains the scenario, beyond just my guess at the sort of scenario where it's happening.

bhushan:
@drbyte
I am trying to remember scenario. I also wrote few tests with different scenarios locally, but it is passing, It was something related to `$gaurd_name` property, I had two Models having permissions `User` and `Company`, and for some reason I used `$gaurd_name` property in Models to set default guard for model.

And it was failing when I tried to fetch $company->roles()->get(['roles.id']) something like this.

with 

```php
"message": "Undefined index: guard_name",
    "exception": "ErrorException",
    "file": "/Users/bhushan/Code/meds/meds-be/vendor/spatie/laravel-permission/src/Models/Role.php",
    "line": 66,
```

this error, 

I have asked for access to client code, If I get any idea I will add a discussion comment.

MdMostaFizurRahaman:
Same problem

bhushan:
@MdMostaFizurRahaman could you send me steps to reproduce.. i will update this PR with test so this fix can be merged if approved ..

PR #1694, opened 2021-03-01
add uuid support.

shobeiry:
add uuid support.

### DIFF
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -114,7 +114,7 @@ class Role extends Model implements RoleContract
         return $role;
     }
 
-    public static function findById(int $id, $guardName = null): RoleContract
+    public static function findById($id, $guardName = null): RoleContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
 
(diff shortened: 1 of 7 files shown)


frital:
Hi @shobeiry !
I'm also need this, and would to start to work on a PR for one of the most popular role / permissions laravels packages (maybe this or laratrust). But in any case, in you PR, i think that a "static::creating" boot method is missed, or not? Because since you lost the auto-increment on db level, you will need generate and assign an uuid when a new model (role or permission) is created. Pls correct me if i am wrong.
Thx in advance!!

shobeiry (commented):
now auto generate uuid on create method added to models.

src/Models/Role.php
```diff
@@ -21,11 +21,32 @@ class Role extends Model implements RoleContract
...
         parent::__construct($attributes);
     }
 
+    public static function boot()
+    {
+        parent::boot();
+        if (config('permission.models.keys_type') == 'uuid') {
+            self::creating(function ($model) {
+                if (is_null($model->id)) {
+                    $model->id = Str::uuid()->toString();
```
drbyte, line 44: Should `Str` be imported?

shobeiry (commented):
i fix Str not importing in Role model.

freekmurze:
Hi, thanks for this. I'm going to pass on this for now, but I'll keep this feature in mind when working on the next major release of the package.

PR #1695, opened 2021-03-03
Custom define the DB indexes naming

andreikainer:
For organizations that are strict about index naming conventions :-)


drbyte:
Your point is noted.

I'm also wary that making this semi-dynamic instead of hard-coded to a fixed length opens it back up to 'name too long' errors as addressed in #1146 and #827. 

Is your case simply that you've opted to call it `model_uuid` instead of `model_id` ?

I suppose as long as it's avoiding the inclusion of the db table-prefix string perhaps it won't be an issue.

andreikainer:
Oh yes true. Forgot about the 'too long' error. You right.

PR #1715, opened 2021-03-28
Retrieve guard only once in middlewares

abenerd:
This should speed up the middlewares marginally as we don't have to retrieve the same guard several time over to act upon it.


drbyte:
Thanks

PR #1727, opened 2021-04-12
Add functionality for regular expression has checking

mylesduncanking:
Queries all permissions and filters based on a regular expression providing a solution for situations where you want to check if any permissions in a structured set are attached.

**Example**
If you want to hide/show an "Imports" menu item if they have access to at least 1 of the imports available. Instead of running `Auth::user()->hasAnyPermission('imports.foo', 'imports.bar', 'imports.apple', 'imports.banana')` you can run `Auth::user()->hasPregPermission('imports.')`

### DIFF
--- a/docs/basic-usage/direct-permissions.md
+++ b/docs/basic-usage/direct-permissions.md
@@ -53,6 +53,12 @@ $user->hasAnyPermission(['edit articles', 'publish articles', 'unpublish article
 $user->hasAllPermissions(['edit articles', 'publish articles', 'unpublish articles']);
 ```
 
+...or if a user has ANY permission matching a regular expression
+
+```php
+$user->hasPregPermission('publish articles'); // Matches "publish articles" AND "unpublish articles"
+```
+
 You may also pass integers to lookup by permission id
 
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks for the suggestion and the code.

How do you see this being purposefully different from the built-in Wildcard Permission feature?

mylesduncanking:
Thanks for commenting and appreciate everything you've built.

My scenario which caused me to look into this was I have currently got 8 different imports and each has a related permission. In future I'll be adding more in and may even rename / remove current ones. Task: avoid having to wrap the menu item in an if which lists them all out. If they can't do any imports, don't render the menu item.

The wildcard permission may be able to do that and I've just misread the docs. What I understood was you needed to have a specific permission created for "imports.*" and then when assigned gives permission to all imports.

Other scenarios may well exist too when just having the ability to use regex to check permissions is easier /  less maintenance than listing each permission out.

freekmurze:
I'm going to hold off this one for now. I think most people can already use the wildcard permissions. If you need this, add it to your own project.

PR #1739, opened 2021-05-12
Remove redundant code in getMigrationFileName() function

jochensengier:
Unless I'm mistaken, these lines can be safely deleted.

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -176,9 +176,6 @@ class PermissionServiceProvider extends ServiceProvider
         $filesystem = $this->app->make(Filesystem::class);
 
         return Collection::make($this->app->databasePath().DIRECTORY_SEPARATOR.'migrations'.DIRECTORY_SEPARATOR)
-            ->flatMap(function ($path) use ($filesystem) {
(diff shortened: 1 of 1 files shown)


drbyte:
It's true that those lines are not presently needed. 
They were added in order to support the addition of more migration files. Later the decision was to stick with one migration file.

drbyte:
I think there was some intention of using some of that logic to test whether a migration of the same name (but different date) already exists.

jochensengier:
No worries, it's not a big deal, I just wanted to point it out.
Feel free to close my PR.

dbakan:
I think, this should be reopened and merged. The current code does not work as promised and this PR fixes it.

Looks like the problem was introduced with [https://github.com/spatie/laravel-permission/commit/66f6e4a376fed2a3be11605daa4bd02c001c504e](https://github.com/spatie/laravel-permission/commit/66f6e4a376fed2a3be11605daa4bd02c001c504e). As @drbyte said, some lines where added to support multiple migrations, which is great. But it seems that the old lines were left in place unintentionally when they should have been removed.
The hard coded `_create_permission_tables.php` is still in there.

drbyte:
> The current code does not work as promised 

Please elaborate

PR #1747, opened 2021-05-21
index and primary naming the same as table name

shaul1991:
Modified to be registered with the same index and unique name as the custom table name in the config.


drbyte:
The proposed change is likely to cause the index names to be too long for allowed lengths. That's why the specific names were chosen.

So, closing this.

PR #1785, opened 2021-07-19
Update new-app.md

danolu:
The capitalization of super-admin in AuthServiceProvider change in case returns false in the Gate function, disallowing the super admin user to override permissions. This might be due to the fact that variable and constant names are case sensitive in PHP.
New adopters of Spatie/Permissions rely on the documentation for a guide and it has caused quite some issues. ==>
http://5.9.10.113/64808362/laravel-spatie-roles-super-admin-not-working
https://stackoverflow.com/questions/65213504/how-to-use-correctly-super-admins-role-in-laravel-spatie
https://www.bountysource.com/issues/98599883-defining-a-super-admin-not-working

### DIFF
--- a/docs/basic-usage/new-app.md
+++ b/docs/basic-usage/new-app.md
@@ -132,7 +132,7 @@ Super-Admins are a common feature. Using the following approach allows that when
 
 +        // Implicitly grant "Super Admin" role all permission checks using can()
 +        Gate::before(function ($user, $ability) {
-+            if ($user->hasRole('Super-Admin')) {
++            if ($user->hasRole('super-admin')) {
 +                return true;
 +            }
 +        });


drbyte:
Thanks for your submission.

If you want to use lowercase role names in your app, feel free to do so. 
But you must be consistent due to case-sensitivity.

Changing the example from Title-Case to Lowercase won't resolve the issues you linked to.

danolu:
Thank you for your response.
I wasn't insinuating that I prefer lowercase.
I'm only pointing out that there's a case inconsistency in the
documentation. People who follow it holistically (mostly new adopters) will
definitely run into the afore-attached issues.

The role assignment in the PermissionSeeder is in lower case while
the role-Gate assignment in the boot section of AuthServiceProvider is
capitalized.
One of them has to change to match the other to make the documentation apt
to the core.

drbyte:
Oh. I see. You're right. You hadn't mentioned the seeder. 

Will update. 

Thanks

PR #1799, opened 2021-07-27
Make cache more than 90% smaller and Speed up permissions cache lookups

### LINKED ISSUE #1626: The cache is too big
I have more than 1,000 permissions and dozens of roles, and the cache is as large as tens of megabytes, which seriously slows down my redis query. How can I solve this?

erikn69:
**Fully backwards compatible.**
This adds substantial speed increases by making cache smaller on the associations between models and permissions. Part of the changes focus on the `PermissionRegistar`, specifically `getPermissions()`.

@drbyte @freekmurze , i'm really not that immersed on this package, please check this PR, 
~~I have my doubts if the fields `'id', 'name', 'guard_name'` can change by custom fields, but~~ this idea makes everything faster when you have tons of permissions and roles

Closes #1798
Closes #1626

~~maybe, it is necessary to add `if(is_array($this->permissions))` for transition from `Collection` cache, to `array` cache~~, now auto upgrade to array cache after `expiration_time` or after manually `rebuild cache` 

Also, uses `$collection->first(...)` or `$collection->filter(...)` instead of `$collection->where(...); $collection->where(...);` to avoid creating two or more temporary arrays, now stop as soon as a match is found when uses first. This performance improvement is noticeable when they have many permissions

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -163,12 +163,40 @@ class Permission extends Model implements PermissionContract
     }
 
     /**
-     * Set roles for cache load.
+     * Fill model from array.
      *
-     * @param \Illuminate\Database\Eloquent\Collection $roles
+     * @param array $attributes
      */
-    public function setRolesCollection(Collection $roles)
+    protected function fillModelFromArray(array $attributes)
     {
-        $this->relations['roles'] = $roles;
+        if (isset($attributes['roles'])) {
+            $roleClass = app(PermissionRegistrar::class)->getRoleClass();
+            $this->relations['roles'] = new Collection();
+
+            foreach ($attributes['roles'] as $value) {
+                $this->relations['roles']->push($roleClass::getModelFromArray($value));
+            }
+            unset($attributes['roles']);
+        }
+
+        $this->attributes = $attributes;
+        if (isset($attributes['id'])) {
+            $this->exists = true;
+            $this->original['id'] = $attributes['id'];
+        }
+        return $this;
+    }
+
+    /**
+     * Get model from array.
+     *
+     * @param array $attributes
+     *
+     * @return \Spatie\Permission\Contracts\Permission
+     */
+    public static function getModelFromArray(array $attributes): ?PermissionContract
+    {
+        $permission = new static;
+        return $permission->fillModelFromArray($attributes);
     }
 }
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -36,7 +36,7 @@ class Permission extends Model implements PermissionContract
     {
         $attributes['guard_name'] = $attributes['guard_name'] ?? Guard::getDefaultName(static::class);
 
-        $permission = static::getPermissions(['name' => $attributes['name'], 'guard_name' => $attributes['guard_name']])->first();
+        $permission = static::getPermission(['name' => $attributes['name'], 'guard_name' => $attributes['guard_name']]);
 
         if ($permission) {
             throw PermissionAlreadyExists::create($attributes['name'], $attributes['guard_name']);
@@ -85,7 +85,7 @@ class Permission extends Model implements PermissionContract
     public static function findByName(string $name, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
+        $permission = static::getPermission(['name' => $name, 'guard_name' => $guardName]);
         if (! $permission) {
             throw PermissionDoesNotExist::create($name, $guardName);
         }
@@ -106,7 +106,7 @@ class Permission extends Model implements PermissionContract
     public static function findById(int $id, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-        $permission = static::getPermissions(['id' => $id, 'guard_name' => $guardName])->first();
+        $permission = static::getPermission(['id' => $id, 'guard_name' => $guardName]);
 
         if (! $permission) {
             throw PermissionDoesNotExist::withId($id, $guardName);
@@ -126,7 +126,7 @@ class Permission extends Model implements PermissionContract
     public static function findOrCreate(string $name, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
-        $permission = static::getPermissions(['name' => $name, 'guard_name' => $guardName])->first();
+        $permission = static::getPermission(['name' => $name, 'guard_name' => $guardName]);
 
         if (! $permission) {
             return static::query()->create(['name' => $name, 'guard_name' => $guardName]);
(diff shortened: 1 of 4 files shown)


kalemdzievski:
This issue ([#1798](https://github.com/spatie/laravel-permission/discussions/1798#discussioncomment-1078601)) causes a big performance problem for our platform.

Tested out this MR. Works nice. Cut the size of the cache from 3.8MB to 160kB, which is a huge improvement.

It would be great if this improvement can be merged.

fredsal:
Tested and works like a charm

PaolaRuby:
I need this 👍

alvarorvn:
it's a good solution, nice work!

dj-Andres:
It would improve in a good way because the waiting time would be reduced in this way it would help to improve the performance

kalemdzievski:
I just wanna queue in again. 

It seems like this was the root cause of 10+ outages in our system for the last month or so. 
It was a bit hard to pinpoint the issue, but eventually, we started digging into this library because there were many pointers.

In the last two outages, we first disabled the check for permissions at the begging of a request (not great, not terrible), and this resolved the first of these two last outages. In the second (the last one at the moment, I hope so..), we totally disable the cache for the library, and immediately, everything got back to normal. We gonna roll like this for some time. It's not going to be an issue.

This really put a lot of pressure on our Redis instance.

Anyways, the library is pretty awesome, it was super useful to us and I hope we can get this improvement as soon as possible it will be even better.

Cheers!

fredsal:
@davejamesmiller could it be helpful if you do a benchmark of this?

erikn69:
@drbyte why?
https://github.com/spatie/laravel-permission/blob/771158773ef29db36ae8949153c7ddfa72bae9a4/src/PermissionRegistrar.php#L35-L36
https://github.com/spatie/laravel-permission/blob/771158773ef29db36ae8949153c7ddfa72bae9a4/src/PermissionRegistrar.php#L57
can I delete that property? or is it thought for the future?

drbyte:
> can I delete that property? or is it thought for the future?

I believe the `$cacheModelKey` property is set as a `static` property to avoid extra lookups.
What is your reasoning for wanting to delete it?

> can i change old `cache_expiration_time`(`config('permission.cache_expiration_time')`) to `\DateInterval::createFromDateString('24 hours')` as default?

It's written this way (lookup `.`, fallback to `_`) to accommodate old syntax to avoid breaking changes.
What's your reasoning for considering hard-coding a value?

erikn69:
>I believe the $cacheModelKey property is set as a static property to avoid extra lookups.

I understand that, but it is deprecated because the code where it was used was deleted
Compare [2.35.0](https://github.com/spatie/laravel-permission/blob/2.35.0/src/Traits/HasPermissions.php) and [2.34.0](https://github.com/spatie/laravel-permission/blob/2.35.0/src/Traits/HasPermissions.php)

>What's your reasoning for considering hard-coding a value?

It's been a long time since that version(and it was a  really small time with `cache_expiration_time` on v2), but it would be better to leave it if you think that

brunoatsuperhero:
Hello, Thanks for the fix. I wonder how long it would take to get this one "officially" release? 

Cheers

drbyte:
> @drbyte review, corrections, thoughts?
> 
> also, why?
> 
> https://github.com/spatie/laravel-permission/blob/771158773ef29db36ae8949153c7ddfa72bae9a4/src/PermissionRegistrar.php#L35-L36
> 
> 
> https://github.com/spatie/laravel-permission/blob/771158773ef29db36ae8949153c7ddfa72bae9a4/src/PermissionRegistrar.php#L57
> 
> 
> can I delete that property? or is it thought for the future?
> https://github.com/spatie/laravel-permission/blob/771158773ef29db36ae8949153c7ddfa72bae9a4/src/PermissionRegistrar.php#L54
> 
> 
> can i change old `cache_expiration_time`(`config('permission.cache_expiration_time')`) to `\DateInterval::createFromDateString('24 hours')` as default?

Okay, yes, changing both would be acceptable. Thanks for pointing those out.

src/Models/Permission.php
```diff
@@ -137,11 +137,64 @@ public static function findOrCreate(string $name, $guardName = null): Permission
...
+     *
+     * @param array $attributes
+     *
+     * @return \Spatie\Permission\Contracts\Permission
      */
-    protected static function getPermissions(array $params = []): Collection
+    public static function getModelFromArray(array $attributes): ?PermissionContract
     {
-        return app(PermissionRegistrar::class)
-            ->setPermissionClass(static::class)
```
drbyte, line 144: Is removal of this call to `setPermissionClass()` intentional? What's lost by doing this?

erikn69: intentional, because
https://github.com/spatie/laravel-permission/blob/771158773ef29db36ae8949153c7ddfa72bae9a4/src/PermissionRegistrar.php#L45
but thanks for point that, i gonna revert that change, because i didn't read
>This is only intended to be called by the PermissionServiceProvider on boot, so that long-running instances like Swoole don't keep old data in memory.

So, it is an important fix, I suppose

drbyte:
Thank you @erikn69

src/PermissionRegistrar.php
```diff
@@ -114,27 +111,59 @@ public function clearClassPermissions()
...
+                    foreach ($permission['roles'] ?? [] as $j => $roles) {
+                        unset($permissions[$i]['roles'][$j]['pivot']);
+                    }
+                }
+                return $permissions;
+            });
+            if (is_array($this->permissions)) {
+                $permissions = new Collection();
+                foreach ($this->permissions as $value) {
+                    $permissions->push($this->permissionClass::getModelFromArray($value));
```
incrize, line 135: @erikn69, @drbyte 
Thanks for the improvements. Is there a reason why new methods were not added to the Role and Permission interface?

erikn69: Adding methods to an interface is a BC break, wich methods ? `getModelFromArray`?

drbyte: @erikn69 I'd accept a PR which updates the interface, since I'm leaning towards pushing the currently-open PRs into a v5.0 ... so your PR could be safely merged into that.

incrize: Backward compatibility already broken since 4.3. if used own models for roles and permissions, a fatal error is thrown due to the lack of the getModelFromArray method. It looks like now it makes no sense to wait for the 5th version to change the interface.

erikn69: @drbyte How about I put a condition so that it only uses the cache trick only when extending from spatie models? Just for now

drbyte: Yes, that should solve the interface issue and fix the BC

erikn69: @drbyte fixed on #1824

drbyte: @incrize, do you mind testing the fix? It's now merged into `dev-master`.

erikn69: #1826 @drbyte maybe better and permanent solution

incrize: Checked out changes #1826 locally, all my tests pass.

PR #1804, opened 2021-08-04
Teams/Groups feature

erikn69:
##TEAMS FEATURE
-------
**Usage in [docs](https://github.com/erikn69/laravel-permission/blob/teams_feature/docs/basic-usage/teams-permissions.md)**

A **fully backwards compatible** addition to for teams/groups functionality.  I have added the tests and docs for this PR, and i can help to bug's review around it. I have been using this code overriding the methods with custom models for 2 years without problems, now i want to share with everyone.

Teams table must be already migrated on db, and foreign key can be customizable by config file. `team_id` as default. ~~Also, i can improve the upgrade method with another `migration` if necessary~~ (done, added upgrade migration and command `setup-teams`, roles with teams support on command `create-role`).

**Example Communities #1782**:
```php
'teams' => true,
/*
  * Change this if you want to use teams feature and your related model
  * foreign key are other than `team_id`.
  */
'team_foreign_key' => 'community_id',

```
image
Where green tables(`users`, `community_user`, and `communities`) are handled by custom migrations. Only new pivot `community_id` is added to the package migration file.

Closes #1812
Closes #1782
Closes #1743
Closes #1744
Closes #1654
Closes #1629
Closes #1634
Closes #1499
Closes #1362
Closes #1353
Closes #1348
Closes #1187
Closes #1186
Closes #1135
Closes #1067
Closes #551
Closes #540
Closes #8
And many others

##BUGS FIX
-------
~~On tests `*_before_saving_object_doesnt_interfere_with_other_objects`, sqls from first object are executing on second object and try to sync and load relation again~~(#1819)

And other small fixes that i already forget

PR #1807, opened 2021-08-06
- Added load migration method to service provider boot

PouriyaK7:
Hi, with this you don't need to publish migrations for using it. Laravel will migrate it without any publish needed.

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -28,6 +28,8 @@ class PermissionServiceProvider extends ServiceProvider
         $this->app->singleton(PermissionRegistrar::class, function ($app) use ($permissionLoader) {
             return $permissionLoader;
         });
+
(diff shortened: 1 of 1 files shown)


erikn69:
I added a new optional migration on #1804 https://github.com/spatie/laravel-permission/pull/1804/commits/6dd191a892c1beb52465d3d137e16cb3f6dbdec5#diff-3e1e1d66222aac18141faee38f0fddf9c939c755b5c5275a9b68c15c95764c12 , it is for upgrade teams support fields

Question
will `$this->loadMigrationsFrom([__DIR__ . '/../database/migrations']);` publish that migration too?

drbyte:
Seems to be quite incomplete.
Doesn't load the `.php.stub` file, even after renaming it to just `.php`
PR is missing removal of other publishing statements.
BC concerns.

Issue #276, opened 2017-04-25
RoleDoesNotExist in migrations


djkiko20:
I have migration file in which I like to assign role to existing user in my DB. 
Code fail when try to assign role `viewer` on user.

I have in my database:
```
ID | name   | guard_name
1  | viewer | api
```

This is my code in migration file:

```
foreach (User::all() as $user) {
    $user->assignRole('viewer');
}
```

Can I force some `guard` or maybe there is some other solution about this problem ?

sebastiandedeyne:
Could you try clearing you cache between creating and assigning the role?

If you're using the default guard, you don't need to worry about that.

djkiko20:
@sebastiandedeyne  I try to clear cache but problem still here.
My default guard is `web` but my role `viewer` has `guard_name = api` . Maybe this is the problem but this must be like this because I have users on production and all must have `viewer` role.

sebastiandedeyne:
Since you're dealing with multiple guards, instead of passing a string try to assign an actual object with `assignRole` (`$user->assignRole($viewerRole)`)

djkiko20:
`  The given role or permission should use guard `web` instead of `api`.  `

`viewer` role has `quard_name = api`, but my default guard is set to `web`. I think that there is the problem...

sebastiandedeyne:
You can't "share" roles between different guards. Different guards require different roles. If you need need one for two guards, create a second role with the same name.

wyudong:
@sebastiandedeyne I don't not quite understand `assignRole($user->assignRole($viewerRole))` you said. I have two different guards, `web` and `api`. So how shall I assign a role whose `guard_name = api` to a registered user?

    // register
    $user = User::create($inputList);
    $user->assignRole($user->assignRole('composer', 'api'));    // error

Thanks!

sebastiandedeyne:
Could you share the configuration of your guards in `config/auth.php`?

wyudong:
@sebastiandedeyne Here is my `config/auth.php`

```
<?php

return [
    'defaults' => [
        // 'guard' => 'web',
        'guard' => 'api',
        'passwords' => 'users',
    ],

    'guards' => [
        'web' => [
            'driver' => 'session',
            'provider' => 'users',
        ],

        'api' => [
            // 'driver' => 'token',
            // 'driver' => 'passport',
            'driver' => 'jwt',
            'provider' => 'users',
        ],
    ],

    'providers' => [
        'users' => [
[...]

Issue #274, opened 2017-04-23
check permission given by the plugin and ownership permission


stefanocurnis:
Hey guys!

Imagine a **Blog** where every admin can edit all the **Posts**. Post author can edit his own posts even if he's not an admin.
I store all my roles and permissions via spatie/laravel-permission and a simple CRUD backend.
Now i have to check for the "edit post" permission at the Gate. This plugin **automatically define Gates** but what if i need to add other controls to the gates, like ownership check? (post.user_id==user.id)
How can i define a duplicate gate with the same name "edit-post" in AuthServiceProvider which checks post.user_id==user.id as well as "edit-post"  permission assigned with this plugin?

Thanks a lot for your packages! :)

freekmurze:
I don't get what you mean with duplicate gates. I'd have only one `edit-post` check that first checks if a user is and admin and then if the user is the owner of the post.

PR #1817, opened 2021-08-20
Change default-guard-lookup to prefer current user's guard

### LINKED ISSUE #1515: Multiple guards are independent from the Auth::guard() which is used for login?

### LINKED ISSUE #1682: Inserting wrong default guard name while adding role and/or permission
**Describe the bug**
It's inserting the wrong guard name while creating a new role and/or permission also check against the wrong guard while verifying the user role/permission.

**Versions**
4.0.0

PHP version: 7.4

**To Reproduce**
I think it could be done by fixing the models.

**Expected behavior**
It should add web as the default guard name while creating a new role and/or permission.

**Desktop (please complete the following information):**
 - OS: Linux
 - Version 20.04
 
**In my auth config file my defaults guard is web***

### LINKED ISSUE #1793: Checking for permissions checks for first auth guard in the list, rather than default or one that route
Checking if user `can:permission` through a middleware for some reason checks first known auth guard (which was `web` for me, rather than `sanctum` from `laravel/sanctum` package in my case), rather one that user was checked via `auth:guard` with.

**Versions**
- spatie/laravel-permission package version: 4.2.0
- Laravel framework: 8.47.0

PHP version: 8.0.3
Database version: 8.0.24

**To Reproduce**
* Create a user, a role and a permission, and award role to a user, and permission to a role
* Create a custom guard to auth with
* Create a routegroup that is protected with `auth:custom` middleware
* Create a route within that group that is protected with `can:%permission_name%`

**Expected behavior**
* Guard checked for permissions is one that was used in auth middleware.
* Working out of guard without fiddling with the order of guards in config and/or hardcoding it to the model

**Desktop (please complete the following information):**
 - OS: Linux Mint 18

drbyte:
Historically, since v2.0.0, the "detected guard_name" that was applied during lookups was detected from the "first possible defined guard" for that model, even if the "current" guard wasn't "first in the list".
This has led to some confusion when expecting that the current user's guard would actually be used.

This PR changes things to **use the current user's guard** first, assuming it's in the list of potential guards acceptable for the current User model.
After that it will fallback to the old behavior of selecting the first available matchable guard for the current User model.

**THIS IS A BREAKING CHANGE, so be sure to test all authorization aspects of your application to ensure desired behavior.**
You MAY also be able to remove some old complex monkey-patching that you might have done to work around this issue in prior versions.

Fixes #1682
Fixes #1608
Fixes #1384
Fixes #1515
Fixes #1516
Fixes #1793 

Old refs: #298, #274, https://github.com/spatie/laravel-permission/issues/276#issuecomment-301477726

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -8,14 +8,16 @@ use Illuminate\Support\Collection;
 class Guard
 {
     /**
-     * return collection of (guard_name) property if exist on class or object
-     * otherwise will return collection of guards names that exists in config/auth.php.
+     * Return a collection of guard names suitable for the $model,
+     * as indicated by the presence of a $guard_name property or a guardName() method on the model.
      *
      * @param string|Model $model model class object or name
      * @return Collection
      */
     public static function getNames($model): Collection
     {
+        $class = is_object($model) ? get_class($model) : $model;
+
         if (is_object($model)) {
             if (\method_exists($model, 'guardName')) {
                 $guardName = $model->guardName();
@@ -25,8 +27,6 @@ class Guard
         }
 
         if (! isset($guardName)) {
-            $class = is_object($model) ? get_class($model) : $model;
-
             $guardName = (new \ReflectionClass($class))->getDefaultProperties()['guard_name'] ?? null;
         }
 
@@ -34,6 +34,23 @@ class Guard
             return collect($guardName);
         }
 
+        return self::getConfigAuthGuards($class);
+    }
+
+    /**
+     * Get list of relevant guards for the $class model based on config(auth) settings.
+     *
+     * Lookup flow:
+     * - get names of models for guards defined in auth.guards where a provider is set
+     * - filter for provider models matching the model $class being checked (important for Lumen)
+     * - keys() gives just the names of the matched guards
+     * - return collection of guard names
+     *
+     * @param string $class
+     * @return Collection
+     */
+    protected static function getConfigAuthGuards(string $class): Collection
+    {
         return collect(config('auth.guards'))
             ->map(function ($guard) {
                 if (! isset($guard['provider'])) {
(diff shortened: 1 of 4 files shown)


drbyte:
Feedback welcome. Particularly interested in any BC problems created by this.

/ping @erikn69  @musapinar @thoresuenert

/credit @Daniyal-Javani re # 1384 , @thoresuenert re: # 1515, 1516

erikn69:
I have tested this, and it is not a breaking change for me, it works with and without my middleware patch 👍

kellerjmrtn:
Late to the party, but for anyone seeing issues after updating

We had all of our permissions and roles defined solely for the `web` guard, but used both `web` and `api` guards for logins (username/password vs api key/token). We were therefore implicitly relying on the old incorrect behavior to log us in via the `api` guard, but check permissions and roles via the `web` guard. After updating, this no longer worked and errors were being (correctly) thrown

See https://spatie.be/docs/laravel-permission/v6/basic-usage/multiple-guards#content-forcing-use-of-a-single-guard for the fix. Adding
```
protected function getDefaultGuardName(): string
{
    return 'web';
}
```
worked for us and permissions are now being checked with the `web` guard, despite logging in via the `api` guard

Luke-Shepp:
This really needs to be highlighted in the upgrade notes. It's took a lot of digging through changelogs to figure out why out user->can() has stopped working after an upgrade through 5.x

PR #1825, opened 2021-08-24
custom field database has role

ferdhika31:
i think we need this

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -171,7 +171,7 @@ trait HasRoles
      * @param string|null $guard
      * @return bool
      */
-    public function hasRole($roles, string $guard = null): bool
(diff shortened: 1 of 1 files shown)


drbyte:
> i think we need this

Please explain the business problem you're trying to solve with this, and why this approach was necessary.

erikn69:
@ferdhika31 maybe a configuration is better, and tests are needed, I think if they use a custom field, it breaks this code and maybe others
https://github.com/spatie/laravel-permission/blob/dab7234595d9e4810e4e458570d11c2d9f1f2850/src/PermissionRegistrar.php#L123

src/Traits/HasRoles.php
```diff
@@ -171,16 +171,16 @@ public function syncRoles(...$roles)
      * @param string|null $guard
```
erikn69, line 171: Missing `@param string|null $customField`, here and on the others

PR #1826, opened 2021-08-24
Avoid BC break on cache fix

erikn69:
Last try
Using laravel's `hydrate` for avoid changes on interface, 
i've never used it before but it seems to work
Better solution for #1799 problem

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -119,22 +119,22 @@ class PermissionRegistrar
     {
         if ($this->permissions === null) {
             $this->permissions = $this->cache->remember(self::$cacheKey, self::$cacheExpirationTime, function () {
-                $permissions = $this->getPermissionClass()->select('id', 'name', 'guard_name')
-                    ->with('roles:id,name,guard_name')
-                    ->get();
-
-                if (! method_exists($this->getPermissionClass(), 'getModelFromArray')) {
-                    return $permissions;
-                }
-
                 // make the cache smaller using an array with only required fields
-                return $permissions->map(function ($permission) {
-                    return $permission->only('id', 'name', 'guard_name') + ['roles' => $permission->roles->map->only('id', 'name', 'guard_name')->all()];
-                })->all();
+                return $this->getPermissionClass()->select('id', 'name', 'guard_name')
+                    ->with('roles:id,name,guard_name')->get()
+                    ->map(function ($permission) {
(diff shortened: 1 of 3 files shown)


src/PermissionRegistrar.php
```diff
@@ -123,18 +123,19 @@ private function loadPermissions()
...
             if (is_array($this->permissions)) {
-                $this->permissions = (new Collection($this->permissions))->map(function ($permission) {
-                    return $this->permissionClass::getModelFromArray($permission);
+                $this->permissions = $this->getPermissionClass()::hydrate(
+                    collect($this->permissions)->map(function ($item) {
+                        return collect($item)->only('id', 'name', 'guard_name')->all();
+                    })->all()
+                )
+                ->each(function ($permission, $i) {
+                    $permission->setRelation('roles', $this->getRoleClass()::hydrate($this->permissions[$i]['roles'] ?? []));
```
drbyte, line 138: Does this make it as "small" as before? I think it leaves the `['pivot']` sub-array, doesn't it?

erikn69: I don't think that, `hydrate` makes a Collection from array, a couple lines up i make a array from Collection for cache

```php
return $permission->only('id', 'name', 'guard_name') + 
     ['roles' => $permission->roles->map->only('id', 'name', 'guard_name')->all()];
```
 `$permission->roles->map->only('id', 'name', 'guard_name')` returns array without `['pivot']`

drbyte: I guess if `$this->permissions` is never an array, it won't matter.

erikn69: Yes, if `$this->permissions` is a collection, it avoid that code, only with array makes `hydrate` for transform to a colection, It is to support transition, we can remove that condition on next major version, It works same as before(tested on mi applications, no `['pivot']`, minimal expression)

```php
array:358 [▼
  0 => array:4 [▼
    "id" => 197
    "name" => "view_audit"
    "guard_name" => "web"
    "roles" => array:3 [▼
      0 => array:3 [▼
        "id" => 0
        "name" => "System"
        "guard_name" => "web"
      ]
      1 => array:3 [▶]
      2 => array:3 [▶]
    ]
  ]
```

erikn69:
@drbyte new idea, make cache even smaller using alias for field's names, i'm using ` ?? $item['id'], ?? $item['name']` for avoid breaks on everyone who already has 4.3.0

PR #1834, opened 2021-09-01
[V4] Avoid Roles over-hydration

danilopinotti:
When we using this lib in an application with a lot of permissions and roles, we can observe that it has an over-hydration of "Role" model on loading permissions.

image
RAM usage: image

After this PR, It reduces dramatically:
image
RAM usage: image

The execution time average is reduced too.
From ~330ms to ~270ms

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -32,6 +32,9 @@ class PermissionRegistrar
     /** @var string */
     public static $cacheKey;
 
+    /** @var array */
+    private $cachedRoles = [];
+
     /**
      * PermissionRegistrar constructor.
      *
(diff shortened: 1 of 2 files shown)


erikn69:
Yes, it's better, only a question
why not just  `private $cachedRoles = [];` and `$this->cachedRoles = [];`
example 
```php
private function getHydratedRole(array $item)
  {
      $roleId = $item['i'] ?? $item['id'];

      if (!isset($this->cachedRoles[$roleId])) {
          $roleClass = $this->getRoleClass();
          $roleInstance = new $roleClass;
          $this->cachedRoles[$roleId] = $roleInstance->newFromBuilder([
              'id' => $roleId,
              'name' => $item['n'] ?? $item['name'],
              'guard_name' => $item['g'] ?? $item['guard_name'],
          ]);
      }

      return $this->cachedRoles[$roleId];
  }
```

erikn69:
Excelent thanks

danilopinotti:
@erikn69  Thanks for your review and suggestions.

I applied them.

PR #1837, opened 2021-09-02
[V4] Avoid BC break on extra custom fields (on smaller cache)

### LINKED ISSUE #1836: Columns not available for extended Permissions model
**Describe the bug**
Hi, first of all, I'm sorry if this is not a bug, but I'm having a hard time trying to get my app working with the latest release, so I'm not sure if it's a specific problem from my use case or if it's general.
Basically, recent releases kind of "broke" backwards compatibility with some features that are still documented as available, such as the ability to extend the models.
On my application I've extended both models (Permissions and Roles), mainly to have an extra attribute for the permissions (my app is divided between modules, so I have a `module_id` column on my permissions table that I double check, to make sure users can't use permissions of modules they don't have, build custom menus and so on).
I haven't updated the package for a few months now, and when I tried to update today, I encountered this issue, apparently in order to shrink the cache, the package has now the attributes of the permissions model hard-coded on the PermissionRegistrar, so I can't check for any other columns I have on the DB because they will never be available.
I'm not sure if this was intended, but if so, I guess we should update the docs and possibly document it as a breaking change (?).

**Versions**
- spatie/laravel-permission package version: 5.1.1
- illuminate/framework package version: 8.58.0

PHP version:
8.0.10

Database version:
MySQL 8.0.23

**To Reproduce**
Just add a new column to the permissions table and try to access it anywhere on the app using a method such as the `findByName` from the Permission's model.

**Expected behavior**
I guess it would be nice if we could have the attributes we want to cache on the config or something, so we can still shrink the cache in a way, but also have the flexibility to have extra columns, if needed.

### LINKED ISSUE #2016: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'name' in 'field list'
### Code

$administratorRole = **OBJECT**;
$permissions = **ARRAY OF PERMISSIONS**;
$administratorRole->syncPermissions($permissions);

~~~sql
 SQLSTATE[42S22]: Column not found: 1054 Unknown column 'name' in 'field list' (SQL: select `id`, `id` as `i`, `name` as `n`, `guard_name` as `g` from `permissions`)
~~~

image
image

#### Composer 
"spatie/laravel-permission": "^5.5",

#### PHP
PHP 7.4.27 (cli) (built: Dec 20 2021 21:28:15) ( NTS )

Can any one help on this !Thanks in advance

erikn69:
Closes #1836
Closes #2046
Closes #2016

Set all models fields on cache, still using alias and arrays for smaller cache, added alias for most common field names([#1827](https://github.com/spatie/laravel-permission/discussions/1827#discussioncomment-1234029)).

Example 
```php
array:358 [▼
  0 => array:8 [▼
    "i" => 197
    "n" => "view_audit"
    "g" => "web"
    "l" => "Audit"
    "e" => null
    "y" => "A"
    "r" => array:3 [▼
      0 => array:5 [▼
        "i" => 0
        "t" => null
        "n" => "System"
        "g" => "web"
        "y" => "A"
      ]
```
`1631207353a:358:{i:0;a:8:{s:1:"i";i:197;s:1:"n";s:10:"view_audit";s:1:"g";s:3:"web";s:1:"l";s:10:"Audit";s:1:"e";N;s:1:"y";s:1:"A";s:1:"r";a:3:{i:0;a:5:{s:1:"i";i:0;s:1:"t";N;s:1:"n";s:8:"System";s:1:"g";s:3:"web";s:1:"y";s:1:"A";}i:`

Added test `it_can_use_custom_fields_from_cache`, checked not over hydration

image

## UPDATE:
Inspired by @danilopinotti on #1834, now cache avoid duplicate roles on each permissions, only save ids, and roles are saved on different array, this makes cache even smaller

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -11,6 +11,8 @@ use Spatie\Permission\Contracts\Role;
 
 class PermissionRegistrar
 {
+    use PermissionLoaderTrait;
+
     /** @var \Illuminate\Contracts\Cache\Repository */
     protected $cache;
 
@@ -31,9 +33,6 @@ class PermissionRegistrar
 
     /** @var string */
     public static $cacheKey;
-
-    /** @var array */
-    private $cachedRoles = [];
 
     /**
      * PermissionRegistrar constructor.
@@ -120,37 +119,11 @@ class PermissionRegistrar
      */
     private function loadPermissions()
     {
-        if ($this->permissions !== null) {
+        if ($this->permissions) {
             return;
         }
 
-        $this->permissions = $this->cache->remember(self::$cacheKey, self::$cacheExpirationTime, function () {
-            // make the cache smaller using an array with only required fields
-            return $this->getPermissionClass()->select('id', 'id as i', 'name as n', 'guard_name as g')
-                ->with('roles:id,id as i,name as n,guard_name as g')->get()
-                ->map(function ($permission) {
-                    return $permission->only('i', 'n', 'g') +
-                        ['r' => $permission->roles->map->only('i', 'n', 'g')->all()];
-                })->all();
-        });
-
-        if (is_array($this->permissions)) {
-            $this->permissions = $this->getPermissionClass()::hydrate(
-                collect($this->permissions)->map(function ($item) {
-                    return ['id' => $item['i'] ?? $item['id'], 'name' => $item['n'] ?? $item['name'], 'guard_name' => $item['g'] ?? $item['guard_name']];
-                })->all()
-            )
-            ->each(function ($permission, $i) {
-                $roles = Collection::make($this->permissions[$i]['r'] ?? $this->permissions[$i]['roles'] ?? [])
-                        ->map(function ($item) {
-                            return $this->getHydratedRole($item);
-                        });
-
-                $permission->setRelation('roles', $roles);
-            });
-
-            $this->cachedRoles = [];
-        }
+        $this->loadPermissionsFromCache();
     }
 
     /**
@@ -221,21 +194,13 @@ class PermissionRegistrar
         return $this->cache->getStore();
     }
 
-    private function getHydratedRole(array $item)
+    /**
+     * Get the instance of the Cache Repository.
+     *
+     * @return \Illuminate\Cache\Repository
+     */
+    public function getCacheRepository(): \Illuminate\Cache\Repository
     {
-        $roleId = $item['i'] ?? $item['id'];
-
-        if (isset($this->cachedRoles[$roleId])) {
-            return $this->cachedRoles[$roleId];
-        }
-
-        $roleClass = $this->getRoleClass();
-        $roleInstance = new $roleClass;
-
-        return $this->cachedRoles[$roleId] = $roleInstance->newFromBuilder([
-            'id' => $roleId,
-            'name' => $item['n'] ?? $item['name'],
-            'guard_name' => $item['g'] ?? $item['guard_name'],
-        ]);
+        return $this->cache;
     }
 }
--- a/src/PermissionLoaderTrait.php
+++ b/src/PermissionLoaderTrait.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Spatie\Permission;
+
+use Illuminate\Database\Eloquent\Collection;
+
+trait PermissionLoaderTrait
+{
+    /**
+     * Array for cache alias
+     */
+    private function aliasModelFields(array &$keys, array $newKeys = [], array $except = []): void
+    {
+        $alphas = ! count($keys) ? range('a', 'n') : range('o', 'z');
+
+        foreach ($newKeys as $i => $value) {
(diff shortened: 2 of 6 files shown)


erikn69:
@danilopinotti please check that something does not fail on your side

erikn69:
@danilopinotti please check that something does not fail on your side(`PermissionRegistrar.php`)

erikn69:
@danilopinotti please check that something does not fail on your side(`PermissionRegistrar.php`)

danilopinotti:
@erikn69 Yes! I tried now and it works fine.

But I have a question. There is a reason to set an alias for "descrip", once already has an alias for "description"?

erikn69:
>There is a reason to set an alias for "descrip", once already has an alias for "description"?

Yes, they are two diferent fields, so they have two diferent alias, i made a little research for guess which are the most common extra field names on this package, that will not affect performance

src/PermissionRegistrar.php
```diff
@@ -124,32 +121,50 @@ private function loadPermissions()
...
         }
 
-        $this->permissions = $this->cache->remember(self::$cacheKey, self::$cacheExpirationTime, function () {
-            // make the cache smaller using an array with only required fields
-            return $this->getPermissionClass()->select('id', 'id as i', 'name as n', 'guard_name as g')
-                ->with('roles:id,id as i,name as n,guard_name as g')->get()
-                ->map(function ($permission) {
-                    return $permission->only('i', 'n', 'g') +
-                        ['r' => $permission->roles->map->only('i', 'n', 'g')->all()];
+        $alias = ['id'=>'i', 'name'=>'n', 'guard_name'=>'g', 'label' => 'l', 'type' => 'y', 'description' => 'd', 'descrip' => 'e'];
```
isecchin, line 124: What do you think about possibly defining the `$alias` array in the package config as I suggested on the issue #1836?

Maybe even the columns of the roles as well (we would retrieve the whole model and cache only the attributes defined on the config).

We can use the values you put here as a default on the configuration, but that would allow for users to change them according to their needs, maybe having even smaller caches (e.g. if they don't need to cache a specific column, or defining an alias for custom columns as well).

That way this package doesn't need to worry about merging this values with the custom attributes of the models, having to worry about ignoring others (we could remove the `$except` array, if the attribute is not mapped on the config, we just don't cache it) and we just leave for the end user to edit their configuration according to their needs on each individual project.

We could set a `models` (or some other name, idk) to the `cache` tag on permissions:

```php
[
    'cache' => [
        'models' => [
            'role' => [
                'id' => 'i',
                'name' => 'n',
                'guard_name' => 'g',
            ],
            'permission' => [
                'id' => 'i',
                'name' => 'n',
                'guard_name' => 'g',
                'label' => 'l',
                'type' => 'y',
                'description' => 'd',
                'descrip' => 'e'
            ],
        ],
    ],
]
```

danilopinotti: Or even in the model class as a static property "$attributesAlias" or " $cacheAttributesAlias" by example

erikn69: At first I considered that, but almost everyone only uses the library with normal fields, adding that configuration could complicate things more(the aliases cannot be repeated, it makes unexpected errors), 
with this change make cache even smaller(thanks for the idea @danilopinotti), 
also the cache is not visible to the developer, if @drbyte thinks it would be a good idea, I can add it

isecchin: My idea was not necessarily about customizing the aliases that you've already setup, although it would allow for that as well, but to allow to define shorter tags for custom fields, from what I see, the PR fixes the problem of custom attributes by retrieving all the attributes of the model, ignoring the timestamps (that I personally don't need, but may be useful for someone, idk), but it uses the full column name on it, so say I have 3 columns that I need to add to the cache, it could increase its size drastically, but with the ability to set the aliases, I could use the same idea you used with a single or a couple of letters only.
I get your concern about making it complicated and causing unexpected errors, but at least the problem of repeated aliases could be fixed by just defining the array in a flipped way (with the alias as keys).
With a customized option, I could shrink my cache as well, since I don't really need the `label`, `type`, `description` and `descrip` columns, but need the `module_id`.
I may be wrong, but I guess users don't mess that much with default config variables unless they need to do so, so I don't think people would have problems with this approach (or the one suggested by @danilopinotti)

erikn69: I undestand your idea, but i need @drbyte point of view(he makes the merge)

>just defining the array in a flipped way (with the alias as keys)

That is good idea

> since I don't really need the label, type, description and descrip columns, but need the module_id

Don't worry, if you don't have any of these fields your cache will not change size, because if they don't exist they are skipped, `modulo_id` will be a small weight, initially my cache weighed 3.1mb now with this change it weighs 60kb and I have 3 custom fields, I achieved a 99% decrease

erikn69: > just defining the array in a flipped way (with the alias as keys)

I followed your idea, looks a little faster now, @isecchin @danilopinotti please try

PR #1883, opened 2021-10-05
used phpcsfixer finder and added phpcsfixer as dev dependency for development

bhushan:
## Description

- used phpcsfixer finder in rules file
- added phpcsfixer as dev dependency for development as not every developer will have it installed


bhushan:
Thumbs down for hacktoberfest request or PR?

PR #1886, opened 2021-10-10
Allow Team ID to be of different type than int

bloomlive:
We use UUIDs for teams (tenants), therefore it would allow us to only change migrations.


erikn69:
Duplicate of #1857

PR #1889, opened 2021-10-13
Ability to manually register all permissions on Laravel's gate

ulhaq:
This could be useful in cases where we don't need to retrieve the permissions from the database on every request.
For example, when permissions are stored and retrieved from an access token.


PaolaRuby:
Why reformatting the whole file just for an if?, it makes the change harder to read

PR #1892, opened 2021-10-15
Scope permissions based on id or name

dees040:
In the `HasRoles` trait you have the ability to scope roles based on the id or name. Also, when giving permission(s) to a model you can give those permissions based on the permission's name or id. However, when using the permission scope you can only use the permission name. So to match the scope functionality from the `HasRoles` trait I though it would be nice to also have the ability to scope permissions based on the id or name.

### DIFF
--- a/tests/HasPermissionsTest.php
+++ b/tests/HasPermissionsTest.php
@@ -82,6 +82,24 @@ class HasPermissionsTest extends TestCase
     }
 
     /** @test */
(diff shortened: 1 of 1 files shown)


AchrafBardan:
Sam for me, i am stuck on this issue for 2 days.

PaolaRuby:
@freekmurze @Nielsvanpach is this package abandoned?

PR #1909, opened 2021-11-03
[V5] UUID full support, support for custom key names on Role,Permission model

erikn69:
- UUID full support on all methods
- Support for custom key names on Role,Permission model instead of defaut one.(`'id'`)
- Other small fixes

keeps the same operation as always

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,11 +22,11 @@
         }
     ],
     "require": {
-        "php" : "^7.2.5|^8.0",
-        "illuminate/auth": "^6.0|^7.0|^8.0",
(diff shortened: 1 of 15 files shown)


freekmurze:
Thanks! This PR seems to do a lot, I see a new migration in there, which makes me think some of the changes are breaking. 

Could you please break this PR up in smaller ones that are easier to review?

erikn69:
Other migration its only for test

erikn69:
Custom new migration is inside `test` path
>  tests/CreatePermissionCustomTables.php

erikn69:
I am trying to push change but how it is closed it is not working

PR #1913, opened 2021-11-04
[V5] Support for custom key names on Role,Permission

erikn69:
https://github.com/spatie/laravel-permission/pull/1909#issuecomment-960652059
>Could you please break this PR up in smaller ones that are easier to review?

First part only changes propertie `id` to `getKey()` or `getKeyName()`
- Support for custom key names on Role,Permission model instead of defaut one.


freekmurze:
Thanks!

PR #1926, opened 2021-11-17
Add support for PHP 8.1

freekmurze:


MarcusOhman91:
This also removed support for Laravel 6. Should be mentioned in docs that for Laravel 6 you need to use v5.3.2

PR #1944, opened 2021-11-26
[V5] WherePivot instead of only Where on team relation pivot, better readability

erikn69:
- [IMPORTANT] Fixed missing return on `getPermissionsTeamId() `helper
- Better readability
- Early returns
- WherePivot instead of only Where(fix relations with teams, less code)
- Global helper `getPermissionsTeamId` used
- `assignRole`, `givePermissionTo` fixed(less code, less bucles with `reduce`, `ensureModelSharesGuard` throws earlier exception)

Works the same, no test nedded

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -44,18 +44,19 @@ trait HasPermissions
      */
     public function permissions(): BelongsToMany
     {
-        return $this->morphToMany(
+        $relation = $this->morphToMany(
             config('permission.models.permission'),
             'model',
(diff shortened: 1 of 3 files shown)


erikn69:
@freekmurze if you need help to migrate syntax to PHP 8 i have time

freekmurze:
Thanks!

freekmurze:
@erikn69 I think at some point in the future, I'll rewrite this package for better readability and maintenance. I feel like it has overgrown a bit now. I'll take care of PHP 8 syntax then.

PR #1946, opened 2021-11-29
Add Cache Repository getter on `PermissionRegistrar`

angeljqv:
Make public Cache Repository getter, usefull when creating custom traits and we want to manipulate permission cache repository 
Also acording https://spatie.be/guidelines/laravel-php#docblocks, phpdoc from `getCacheStore` removed

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -5,6 +5,8 @@ namespace Spatie\Permission;
 use Illuminate\Cache\CacheManager;
 use Illuminate\Contracts\Auth\Access\Authorizable;
 use Illuminate\Contracts\Auth\Access\Gate;
+use Illuminate\Contracts\Cache\Repository;
+use Illuminate\Contracts\Cache\Store;
 use Illuminate\Database\Eloquent\Collection;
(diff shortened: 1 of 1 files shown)


angeljqv:
@freekmurze ping

src/PermissionRegistrar.php
```diff
@@ -76,11 +78,14 @@ public function initializeCache()
         self::$pivotRole = config('permission.column_names.role_pivot_key') ?: 'role_id';
         self::$pivotPermission = config('permission.column_names.permission_pivot_key') ?: 'permission_id';
 
-        $this->cache = $this->getCacheStoreFromConfig();
+        $this->cache = $this->getCacheRepository();
     }
 
-    protected function getCacheStoreFromConfig(): \Illuminate\Contracts\Cache\Repository
+    public function getCacheRepository(): Repository
```
freekmurze, line 84: Renaming this function seems like unnecessarily risky. Why not just change `protected` to `public` so this won't break other people's code?

angeljqv: @freekmurze done,
I added a specific getter for that, not breaking changes

freekmurze:
Thanks!

PR #1948, opened 2021-12-02
[V5] Scope Roles/Permissions by user(and team when available)

erikn69:
https://github.com/spatie/laravel-permission/discussions/1947#discussioncomment-1738676
Useful when teams enabled and editing roles/permissions from an user and many teams(Admin Panel)

**Usage:**
```php
//when no teams enabled or wants default global team
Permission::user($USER_ID)->get(); 
Role::user($USER_ID)->get();
//with specific team
Permission::user($USER_ID, $TEAM_ID)->get();
Role::user($USER_ID, $TEAM_ID)->get();
```
Complements of `scopeRole`, `scopePermission`, but for `Role.php`, `Permission.php` models
https://github.com/spatie/laravel-permission/blob/08cc9bdbf89640cb1d64fb01eb9513c1cbb2549d/src/Traits/HasPermissions.php#L69-L91
https://github.com/spatie/laravel-permission/blob/08cc9bdbf89640cb1d64fb01eb9513c1cbb2549d/src/Traits/HasRoles.php#L74-L99

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -2,6 +2,7 @@
 
 namespace Spatie\Permission\Models;
 
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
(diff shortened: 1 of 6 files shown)


freekmurze:
I'm going to pass on this to keep the package as light as it can be. Add this to your own project if needed.

PR #1952, opened 2021-12-10
Adds documentation for `LazilyRefreshDatabase`

lukeraymonddowning:
Howdy!

When using `LazilyRefreshDatabase`, you ideally want to avoid seeding before each and every test, because that makes the trait useless. The approach described in this doc PR shows how to get around this, and also shows the user how to avoid a caveat with cached permissions when lazily refreshing the database.

Thanks for all the hard work!

Kind Regards,
Luke


freekmurze:
Thanks!

Issue #1957, opened 2021-12-16
Teams permission setup migration fails after upgrading from 4.*


saulens22:
**Before creating a new bug report**
Please check if there isn't a similar issue on [the issue tracker](https://github.com/spatie/laravel-permission/issues) or in [the discussions](https://github.com/spatie/laravel-permission/discussions).

**Describe the bug**
After upgrading to 5.* and setting up Teams functionality, migration created by `php artisan permission:setup-teams` fails to rename two primary keys.
On first run of `artisan migrate` I get
```SQLSTATE[HY000]: General error: 1025 Error on rename of './forge/#sql-alter-1-643' to './forge/model_has_permissions' (errno: 150 "Foreign key constraint is incorrectly formed") (SQL: alter table `model_has_permissions` drop primary key)```

On second run it doesn't try to rename same key, as required columns are there, but it fails on another one:
```SQLSTATE[HY000]: General error: 1025 Error on rename of './forge/#sql-alter-1-659' to './forge/model_has_roles' (errno: 150 "Foreign key constraint is incorrectly formed") (SQL: alter table `model_has_roles` drop primary key)```.

On third run migration is successful, but primary keys keys in database are unchanged. This might cause reduced performance as pre-migration primary keys are in the database and they don't contain `team_id` column.

Adding `Schema::disableForeignKeyConstraints();` to migration fixes this issue.

**Versions**
You can use `composer show` to get the version numbers of:
- spatie/laravel-permission package version: 5.4.0
- **laravel**/framework package: v8.76.1

PHP version: 8.1

Database version: MariaDB 10.6.5

**To Reproduce**

- Upgrade from 4.* to 5.*
- Update config to use Teams
- Run `php artisan permission:setup-teams`
- Run `php artisan migrate`

Here is my example code and/or tests showing the problem in my app:
Migration created by `php artisan permission:setup-teams` :
```
<?php

use Illuminate\Support\Facades\Schema;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Database\Migrations\Migration;

class AddTeamsFields extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        $teams = config('permission.teams');
        $tableNames = config('permission.table_names');
        $columnNames = config('permission.column_names');

        if (! $teams) {
            return;
        }
        if (empty($tableNames)) {
            throw new \Exception('Error: config/permission.php not loaded. Run [php artisan config:clear] and try again.');
        }
[...]

PR #1958, opened 2021-12-16
Disable foreign key checks in teams migration

saulens22:
Suggested fix to #1957


systemsolutionweb:
No need to use `Schema::enableForeignKeyConstraints();` afterwards?

freekmurze:
I'm thinking that in most scenario's the current migrations work just fine. Make adjustments in your project if needed.

PR #1965, opened 2021-12-20
[v5] [docs] fix use multiple guards link in basic usage page

Kamona-WD:
Fix use multiple guards link in basic usage page


freekmurze:
Thanks!

PR #1970, opened 2021-12-21
[V5] Fix `Duplicate entry 'roles_name_guard_name_unique'` on migration for teams

### LINKED ISSUE #1966: Integrity constraint violation on creating role with team id
When I try to create role with same name but different team id, it doesn't allow to do it and throws the error.

1062 - Duplicate entry 'admin-web' for key 'roles_name_guard_name_unique'

Can you please help in the same?

Thanks.

erikn69:
Closes #1966


freekmurze:
Thanks!

PR #1980, opened 2022-01-03
Set nullable for team_foreign_key field in model_has_roles table

Dugard:
@erikn69  Along with command permissions, we may have permissions that have nothing to do with commands.

https://github.com/spatie/laravel-permission/discussions/1840#discussioncomment-1277282

### DIFF
--- a/database/migrations/add_teams_fields.php.stub
+++ b/database/migrations/add_teams_fields.php.stub
@@ -60,7 +60,7 @@ class AddTeamsFields extends Migration
 
         if (! Schema::hasColumn($tableNames['model_has_roles'], $columnNames['team_foreign_key'])) {
             Schema::table($tableNames['model_has_roles'], function (Blueprint $table) use ($tableNames, $columnNames) {
-                $table->unsignedBigInteger($columnNames['team_foreign_key'])->default('1');;
+                $table->unsignedBigInteger($columnNames['team_foreign_key'])->nullable();
                 $table->index($columnNames['team_foreign_key'], 'model_has_roles_team_foreign_key_index');
 
                 if (DB::getDriverName() !== 'sqlite') {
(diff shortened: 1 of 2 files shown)


erikn69:
Are you using teams?
Did you run test?
Looks like a breaking change for https://github.com/spatie/laravel-permission/blob/6f4764c0cafe1cf508b7b9801a81b2caa858150a/src/Traits/HasRoles.php#L57-L61

Dugard:
Hi! Yes, I ran `composer install` and `composer test`. But then for some
reason I added `->primary()` and didn't check it 🤦🏻‍♂️
If my solution is helpful, I will be glad.

https://github.com/spatie/laravel-permission/pull/1980/commits/a1424cbc8d652412fd0c9bf9eb498f1d4c511036

composer test: OK (364 tests, 786 assertions)


пн, 3 янв. 2022 г. в 16:46, erikn69 ***@***.***>:

erikn69:
https://github.com/spatie/laravel-permission/blob/6f4764c0cafe1cf508b7b9801a81b2caa858150a/database/migrations/add_teams_fields.php.stub#L61-L83

Dugard:
Ok, then so.
https://github.com/spatie/laravel-permission/pull/1980/commits/5561de0f3330cef6f19cd95a7dd0d67879a8288c

пн, 3 янв. 2022 г. в 18:18, erikn69 ***@***.***>:

erikn69:
maybe add a test where the use case for which this change is necessary is demonstrated, maybe i'm wrong, but i still thinking that those roles never gonna show up https://github.com/spatie/laravel-permission/blob/6f4764c0cafe1cf508b7b9801a81b2caa858150a/src/Traits/HasRoles.php#L57-L61

Dugard:
@erikn69 I am currently working on a project in which clients can create cabinets for their companies. Several persons can work in each companies. Company = team. Plus our team is also working on the whole project. And our team is not part of the client teams, although it has some access to their cabinets. Also, our team has access to functions outside cabinets.
Can create a hidden cabinet and in other cabinets a hidden role. But it somehow looks crooked...

Dugard:
There is also an variant to make team types: team_type && team_id...

erikn69:
I understand but the migrations are customizable, that is the reason of publish command, at the end of the day @freekmurse has the last word I just try that he has the complete information so that he can decide 👍

lukebouch:
@erikn69 @Dugard Maybe I do not understand correctly, but in order to add the ability to assign roles globally, don't you need to also modify the function that checks those roles and not just the migration schema?

That would require extending whatever class that is responsible for checking the role of a user, right?

erikn69:
The software is provided "AS IS", feel free to overwrite everything you need for your purposes

freekmurze:
I've decided to keep things as they are for now, as the current migration seems to work fine for mosts.

You can indeed publish and modify the migration if needed in your own project.

luizmagao (commented):
Comment and line 95 e 96. Ow right!

PR #2011, opened 2022-02-05
Allow revokePermissionTo to accept Permission[]

### LINKED ISSUE #2010: revokePermissionTo incorrect behaviour with Contracts\Permission[] argument
**Description**
When called with an argument of Permission[] rather than string[] or int[] when revoking multiple permissions, revokePermissionsTo does not revoke any permissions.

givePermissionTo does work correctly with an argument of Permission[] though, so this is a bit confusing.

**Versions**
- laravel/framework package version: 8.79.0

PHP version: 8.1

**To Reproduce**
1. Assign permissions to model with HasPermissions trait
`$u->givePermissionTo([Permission::firstWhere('name', 'view articles'), Permission::firstWhere('name', 'edit articles'), Permission::firstWhere('name', 'delete articles')])`
2. Revoke permissions
`$u->revokePermissionTo([Permission::firstWhere('name', 'edit articles'), Permission::firstWhere('name', 'delete articles')])`
3. Check model's permissions. No permissions will have been revoked.

**Expected Behavior**
The specified permissions are removed from the model.

**Environment:**
 - OS: Linux x86_64
 - Version 5.10.76

brausepulver:
resolves #2010

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -384,15 +384,33 @@ trait HasPermissions
     }
 
     /**
-     * Revoke the given permission.
+     * Revoke the given permission(s).
      *
      * @param \Spatie\Permission\Contracts\Permission|\Spatie\Permission\Contracts\Permission[]|string|string[] $permission
      *
      * @return $this
      */
-    public function revokePermissionTo($permission)
+    public function revokePermissionTo(...$permissions)
     {
-        $this->permissions()->detach($this->getStoredPermission($permission));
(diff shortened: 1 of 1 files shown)


erikn69:
this improvement should also be applied to `HasRoles.php` https://github.com/spatie/laravel-permission/blob/6a3ed627cee28a552b5176c172ae0abc5eb30925/src/Traits/HasRoles.php#L156

freekmurze:
Fixed via #2014

PR #2022, opened 2022-02-17
[V5] Improve typing in role's findById and findOrCreate method

itsfaqih:
This PR fixes the docblock of findById and findOrCreate method in Role model to prevent false error reporting in text editor
image


freekmurze:
Thanks!

PR #2039, opened 2022-03-02
Fix Blade Directives incompatibility with renderers

### LINKED ISSUE #2038: @role displaying on page when spatie/markdown in use
**Describe the bug**
When spatie/laravel-markdown and spatie/laravel-permission are installed the @role type directives are displayed as part of the HTML.

**Versions**
- spatie/laravel-markdown: 2.2.1
- illuminate/framework package: v8.83.2

PHP version: 8.0.16

Database version: mysql  v 8.0.28

**To Reproduce**
Have both laravel-markdown and laravel-permission installed and inside a blade view include both <x-markdown> and an @role type directive. When the page is rendered the @role is output as text

`<x-markdown>{{ $markdown }}</x-markdown> @role('admin')<a href='/admin'>Admin Panel</a>@endrole`

**Example Application**
Here is a link to my Github repo containing a minimal Laravel application which shows my problem: https://github.com/kaptk2/permission-issue

**Expected behavior**
I expected the custom blade directives to be honored and not displayed to the end user.

 **Additional context**
This issue also exists with Laravel 9, the example repository above is using Laravel 9.

 - OS: Ubuntu
 - Version 20.04

erikn69:
Closes #2038
>When `spatie/laravel-markdown` and `spatie/laravel-permission` are installed the `@role` type directives are displayed as part of the HTML.

With `spatie/laravel-markdown` all `laravel-permission` blade directives stop working

Based on [spatie/laravel-blade-javascript](https://github.com/spatie/laravel-blade-javascript) which it is working with `spatie/laravel-markdown`
https://github.com/spatie/laravel-blade-javascript/blob/1a0d5daaf91e123799b0250b216f75d30264502a/src/BladeJavaScriptServiceProvider.php#L21

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -84,65 +84,65 @@ class PermissionServiceProvider extends ServiceProvider
 
     protected function registerBladeExtensions()
     {
-        $this->app->afterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {
-            $bladeCompiler->directive('role', function ($arguments) {
-                list($role, $guard) = explode(',', $arguments.',');
+        $bladeCompiler = $this->app['blade.compiler'];
 
-                return "<?php if(auth({$guard})->check() && auth({$guard})->user()->hasRole({$role})): ?>";
-            });
-            $bladeCompiler->directive('elserole', function ($arguments) {
-                list($role, $guard) = explode(',', $arguments.',');
+        $bladeCompiler->directive('role', function ($arguments) {
+            list($role, $guard) = explode(',', $arguments.',');
 
-                return "<?php elseif(auth({$guard})->check() && auth({$guard})->user()->hasRole({$role})): ?>";
-            });
-            $bladeCompiler->directive('endrole', function () {
-                return '<?php endif; ?>';
-            });
+            return "<?php if(auth({$guard})->check() && auth({$guard})->user()->hasRole({$role})): ?>";
+        });
+        $bladeCompiler->directive('elserole', function ($arguments) {
+            list($role, $guard) = explode(',', $arguments.',');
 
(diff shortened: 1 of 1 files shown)


kaptk2:
This PR fixes #2038 on both the example app and my production application

freekmurze:
Thank you @erikn69

pxpm:
@freekmurze , @erikn69 hey guys 👍 Can you give me a bit guidance here ?

We have a project that we were adding directives using the `Blade` Facade directly: `Blade::directive()`, that was not working properly when used with other packages like this one, that used `afterResolving` callback to inject the directives. 

We tougth about a race condition or something, and changed all our code to work with `afterResolving()`, no more problems. 

The change you guys are doing here was done in `laravel-blade-directive` in 2018 https://github.com/spatie/laravel-blade-javascript/commit/6c468f85664275a13c523bca498e1386868082ed

What I think it's strange is that only now this issue poped up, so I guess something internal in higher depencies changed ? Internally in Laravel or something ? Or it should never worked in the `afterResolving` callback ? 

I just want to make sure that applying the same fix you did here to my package will really solve everything. From my understanding, it should work just fine, just want to make sure!

Thanks,
Pedro

erikn69:
Sorry i can't answer that
https://github.com/spatie/laravel-permission/issues/2044#issuecomment-1058409066

pxpm:
Thanks @erikn69 🙏  ! The comment you referred is not the issue in our case. 

@freekmurze , kudos for your amazing work,  Maybe you can shed some ligth here ✌️ ? I can't really find out why it breaks and it drives me crazy, from my understanding using `afterResolving` it's just a callback for the container, not even related with any blade package, it should not break the way it breaks but it does ... why ? 

Thanks for your time guys !! 🙏

PR #2045, opened 2022-03-03
Fix registerBladeExtensions for Lumen

erikn69:
Closes #2044

https://github.com/laravel/lumen-framework/issues/890#issuecomment-502807776

Issue #2044, opened 2022-03-03
Target class [blade.compiler] does not exist.


dencygb:
While using with `"laravel/lumen-framework": "^8.3.1"`, it gives the following error.

![image](https://user-images.githubusercontent.com/6583127/156634177-49a87b14-eb95-40fe-b9ec-86bb6b9765f6.png)

erikn69:
I think that moving `registerBladeExtensions` from register to boot method fix this, give a try and tell me(I'm not using lumen)
https://github.com/spatie/laravel-permission/blob/677903c69460a1867da833979a0dfdffd163767d/src/PermissionServiceProvider.php#L42

tabacitu:
Hey there! PR https://github.com/spatie/laravel-permission/pull/2047 will solve this. This is such a deep rabbit hole... but we can do this! 💪😅

angeljqv:
@freekmurze must be closed after https://github.com/spatie/laravel-permission/pull/2048

PR #2047, opened 2022-03-07
Revert #2039 - register blade directives after resolving blade compiler

tabacitu:
This PR reverts #2039 , because that PR:
- caused bug https://github.com/spatie/laravel-permission/issues/2044
- made `laravel-permission` compatible with `laravel-markdown`, but **incompatible** with A LOT of very popular Laravel packages, including but not limited to:
    - https://github.com/404labfr/laravel-impersonate
    - https://github.com/albertcht/invisible-recaptcha
    - https://github.com/tighten/ziggy
    - https://github.com/laravel-backpack/crud

@erikn69 **can you please take a look and confirm my logic and PR here?** I think your opinion is very important here to confirm this is a good solution and bring this to an end.

---

Here's the "_executive summary_" for @freekmurze , so that he doesn't spend hours to understand why this is needed:

# Premise

It matters _how_ blade directives are loaded, because:
- packages that use `Blade::directive()` will work with packages that use `Blade::directive()`
- packages that use `afterResolving` will work with packages that use `afterResolving`
- if you have just one package that uses the "other kind" of loading in your Laravel app, the other packages will no longer load their directives;

So we package maintainers have to pick one, and all stick to it. And it's been decided by the crowd: **the most popular packages already use `afterResolving`**. Even taken in isolation, I would argue that it's _better_ for packages to use `afterResolving`, because package should NOT force the resolving of that class. It's better to wait for its natural resolving and just add stuff to the end.

# What happened

Here's a summary:
- for years, `laravel-permission` has been using `afterResolving`;
- issue #2038 came up, saying this package is incompatible with `spatie/laravel-markdown`;
- @erikn69 was kind enough to take a look, and fixed it in #2039 (the PR this reverts) ; the problem is:
    - the fix changes `afterResolving` to `Blade:directive()`, which I think was a mistake;
    - the mindset was the problem: it's not that `spatie/laravel-permission` is incompatible with `spatie/laravel-markdown`, but the other way around: markdown (with 25k downloads) is incompatible with permission (with 7M downloads); **the fix should have been made in `spatie/laravel-markdown`, not in `spatie/laravel-permission`**;
- after that PR was merged, bug #2044 was reported, which then @erikn69 promptly fixed in #2045; 
- other bug reports will follow, as people try to install `spatie/laravel-permission` in combination with other popular packages, since most of them use `afterResolving`;

---

By merging this PR and reverting #2039 :
- [ ] https://github.com/spatie/laravel-permission/issues/2044 will be fixed;
- [ ] https://github.com/spatie/laravel-permission/pull/2045 will no longer be needed, and can be closed;
- [ ] this package becomes compatible with the other popular Laravel packages again;

Going forward, to fix the original issue ( #2038 ), I think a PR should be submitted in `laravel-markdown`, for it to use `afterResolving` too. Happy to help with that if you want.

@freekmurze , @erikn69 - let me know if you have any questions. I've went down the BladeCompiler rabbit hole 1 month ago, it's quite difficult to understand. Of course, let me know if I'm entirely wrong, it wouldn't be the first time 😅 Though I believe I'm right on this one, and we should stick to `afterResolving`, since it's been working well for years.

Cheers!

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -39,7 +39,6 @@ class PermissionServiceProvider extends ServiceProvider
             'permission'
         );
 
-        });
         $this->registerBladeExtensions();
     }
 
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -39,6 +39,7 @@ class PermissionServiceProvider extends ServiceProvider
             'permission'
         );
 
+        });
         $this->registerBladeExtensions();
     }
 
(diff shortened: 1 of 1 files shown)


systemsolutionweb:
First make the PR with the fix to [laravel-markdown](https://github.com/spatie/laravel-markdown)

PaolaRuby:
https://laravel.com/docs/9.x/blade#extending-blade
https://laracasts.com/discuss/channels/laravel/registerbladeextensions-afterresolving-not-working
Just read the docs: you have to write your directives in the boot-method, not in the register-method

erikn69:
>- packages that use Blade::directive() will work with packages that use Blade::directive()
>- packages that use afterResolving will work with packages that use afterResolving
>- if you have just one package that uses the "other kind" of loading in your Laravel app, the other packages will no longer load their directives;

All of these are false, you can use `Blade::directive()`, `$this->app['blade.compiler']`, `afterResolving()` at the same time, i am using it that way, i have `Blade::directive()` on my custom directives, and third party packages uses the others without problem

But i think that using `afterResolving` on markdown it's safer for other packages

tabacitu:
Thanks for the feedback, guys! I super-appreciate it 🙏  So fast and comprehensive!

> First make the PR with the fix to [laravel-markdown](https://github.com/spatie/laravel-markdown)

Was about to right now, but I see @erikn69 has beat me to it. He moves FAST! See https://github.com/spatie/laravel-markdown/pull/35 💪

> Just read the docs: you have to write your directives in the boot-method, not in the register-method

Nice find @PaolaRuby . Indeed it would be better to be in `boot()`. But I tried it, still wouldn't solve the problem here. I found no difference in adding the blade directives in `boot()` vs `register()`.

> i have Blade::directive() on my custom directives, and third party packages uses the others without problem

Hmm interesting. It might not conflict with Blade directives registered _by the project_, haven't tried that. But if you use _packages_ with both `Blade::directive()` and `afterResolving`, the first package will make the rest stop working. See:
- 2017 - https://github.com/albertcht/invisible-recaptcha/issues/41
- 2017 - https://github.com/tighten/ziggy/pull/101
- 2017 - https://github.com/404labfr/laravel-impersonate/pull/25

And similar issues have started popping up now:
- 2022 - https://github.com/404labfr/laravel-impersonate/issues/153
- 2022 - https://github.com/DigitallyHappy/assets/pull/12

> But i think that using afterResolving on markdown it's safer for other packages

Awesome! Glad to hear that 🙏 Thank you for confirming.

So @erikn69 do you agree it's a good idea for @freekmurze  to:
- [ ] merge & tag https://github.com/spatie/laravel-markdown/pull/35
- [ ] merge this PR & tag the release
- [ ] close https://github.com/spatie/laravel-permission/pull/2045
- [ ] close https://github.com/spatie/laravel-permission/issues/2044

?

Thanks again!

erikn69:
> But if you use packages with both Blade::directive() and afterResolving, the first package will make the rest stop working.

Also not true, 
https://github.com/caffeinated/bonsai/blob/466603f12e3ee012d5690c0346c39d5048478d06/src/BonsaiServiceProvider.php#L22-L34
That package works great with any  spatie package and it uses `Blade::directive()`

tabacitu:
Again, thanks for taking the time to thoroughly explore this @erikn69 .

> That package works great with any spatie package and it uses Blade::directive(), maybe there is a race condition but it looks like doesn't matters

Eeexactly - there's a race condition. It does matter, it's just difficult to reproduce the scenarios where it matters 😀 And **without the race condition... you're not actually testing the failing scenario**.  I would have done a PR for this, but the important file is _gitignored_ and gets rewritten upon every `composer dump-autoload`. **To replicate the race condition in your test package please:**
- move `spatie/laravel-permission` to the top of the array, in your `bootstrap/cache/packages.php`;
- run `php artisan view:clear`;


The result of the above will be:

- in 5.5.0, everything works just fine:

CleanShot 2022-03-08 at 11 26 25@2x

- in 5.5.1, things start failing, starting with impersonate in my tests:

CleanShot 2022-03-08 at 11 28 31@2x

(note that in order to see this, not an ugly error, your test project needs to [merge this fix for the impersonate link](https://github.com/erikn69/permission-issue/pull/1))

---

This is what I meant initially. That **in most cases, when `laravel-permission` is using `Blade::directive()`, it just _happens_ to work, because `spatie/laravel-permission` is one of the last service providers to get loaded**. But when it's NOT... it can cause problems for the packages after it - some won't get their directives loaded. Maybe I didn't explain that clearly 🤷‍♂️

---

**In any case. I thought you agreed with me that using `afterResolving` is better. Any reason why we're not concluding and merging this?** This package gets 26.000 downloads every day.

Thanks again, cheers!

PS. Nice job on Bonsai, btw. Wish I knew it existed 1 year ago 😅

tabacitu:
In fact, if you agree this is worth doing @erikn69 , I would close this PR in favour of https://github.com/spatie/laravel-permission/pull/2048 - which is a bit cleaner I think.

tabacitu:
Closing in favor of https://github.com/spatie/laravel-permission/pull/2048

PR #2048, opened 2022-03-08
[Fixes BIG bug] register blade directives after resolving blade compiler

### LINKED ISSUE #2038: @role displaying on page when spatie/markdown in use

tabacitu:
This is a cleaner alternative to https://github.com/spatie/laravel-permission/pull/2047

Fixes #2038, #2044, #2045

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -7,6 +7,7 @@ use Illuminate\Routing\Route;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\ServiceProvider;
+use Illuminate\View\Compilers\BladeCompiler;
 use Spatie\Permission\Contracts\Permission as PermissionContract;
 use Spatie\Permission\Contracts\Role as RoleContract;
 
@@ -39,7 +40,9 @@ class PermissionServiceProvider extends ServiceProvider
             'permission'
         );
 
-        $this->registerBladeExtensions();
+        $this->app->afterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {
+            $this->registerBladeExtensions($bladeCompiler);
+        });
     }
 
     protected function offerPublishing()
@@ -81,10 +84,8 @@ class PermissionServiceProvider extends ServiceProvider
         $this->app->bind(RoleContract::class, $config['role']);
     }
 
-    protected function registerBladeExtensions()
+    protected function registerBladeExtensions($bladeCompiler)
     {
-        $bladeCompiler = $this->app['blade.compiler'];
-
         $bladeCompiler->directive('role', function ($arguments) {
             list($role, $guard) = explode(',', $arguments.',');
 
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -40,7 +40,7 @@ class PermissionServiceProvider extends ServiceProvider
             'permission'
         );
 
-        $this->app->afterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {
+        $this->callAfterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {
             $this->registerBladeExtensions($bladeCompiler);
         });
     }


erikn69:
Looks great 👍

erikn69:
@freekmurze after https://github.com/spatie/laravel-markdown/pull/35 this could be merged if it's better for everyone, 
it would revert to 5.5.0 functionality

tabacitu:
Thanks for the double-check @erikn69 🎉

TLDR for @freekmurze : 
- this package was using `afterResolving('blade.compiler')` until 6 days ago;
- PR #2039 changed that; the change made this package the final stop for registering blade directives; when other packages tried to register blade directives after it... they wouldn't work, they'd be output as plain text;
- this PR reverts #2039 , and leaves the campsite a little cleaner than before;

For details, see #2047 . But it's a lot.

Let me know if you have questions @freekmurze .
Cheers!

erikn69:
For the records, seems like `Blade::directive()`/`$this->app['blade.compiler']` force to resolve `BladeCompiler` inmediatly,  after that others `afterResolving` callbacks never gets executed because `BladeCompiler` is already resolved  (race condition)

erikn69:
For the records, seems like `Blade::directive()`/`$this->app['blade.compiler']` force to resolve `BladeCompiler` inmediatly,  after that others `afterResolving` callbacks never gets executed because `BladeCompiler` is already resolved  (race condition)

erikn69:
https://github.com/laravel/framework/pull/41397#issuecomment-1062169244

fredsal:
@freekmurze ping :/

erikn69:
>don't register Blade extensions in the `register` method.

https://github.com/laravel/framework/pull/41397#issuecomment-1062223510

tabacitu:
While I agree that a fix upstream would be nice, I still think this should be merged ASAP. Every day we delay, this package has 26.000 downloads with potentially breaking setups.

Please don't drag this on any longer.

@freekmurze ?

pxpm:
I am a bit tired to do the Pr today. I can do it tomorrow morning. Don't worry, by lunch time everything is solved!

erikn69:
>@tabacitu the update is: use $this->callAfterResolving instead of $this->app->afterResolving

Still this PR needs to be merged

tabacitu:
Agreed. This PR still needs to be merged.

tabacitu:
I've just changed the PR to use the solution Taylor Otwell and Graham Campbell proposed in https://github.com/laravel/framework/pull/41397 - using `callAfterResolving()` instead of `afterResolving()`. This solves the problem for now. And for the future 💪

@erikn69 can you please double-check so that it's easier for @freekmurze to merge?

pxpm:
@tabacitu @erikn69 @freekmurze 

I was the one discussing this with Taylor Otwell and Graham Campbell in https://github.com/laravel/framework/pull/41397 . 

The solution proposed by them (and it does what we want), is using `$this->callAfterResolving()`, that is basically the same as calling `$this->app->afterResolving()` with an aditional parameter to run the callback if the class is already resolved. 

I don't quite agree with their solution for this problem, I was searching the Container api/docs like mad to find something like this, it happens it's not in `Container`, but in `ServiceProvider`. 

Using `$this->callAfterResolving()` 100% solves the problem. This should be the "de-facto" solution here, since this package was using `afterResolving()`, changing to `Blade::directive()` - same as `app('blade.compiler')` may have big implications in packages that rely on `afterResolving()` of this package. 

@freekmurze  please consider merging this PR as a Backward Compatible change. It will fix your package problem once and for all as Taylor suggested, but would not break so many other packages as changing to `Blade::directive()` approach. 

Wish you all guys the best,
Pedro

freekmurze:
Thanks!

tabacitu:
No, thank you! And thanks @erikn69 and everyone else for the double-check. 

Wheew... This was a nasty one 😅 Glad it's over.

PR #2056, opened 2022-03-16
tiny fix for tiny possible confusion (which got me personally)

toramanlis:
I have added comments to the relevant files to better explain the situation.


freekmurze:
Thanks! To me the original text is clearer than the proposed new text.

PR #2058, opened 2022-03-16
Fix markdown code issues in install guide

duncanmcclean:
Right now, in the docs, the commands & code snippets are being formatted oddly. I've tried to fix it but can't confirm as I don't know how the docs work but pretty confident it'll make some kind of improvement.

image


erikn69:
That is problem on https://github.com/spatie/laravel-markdown

On github with your change it lost indent
image

PR #2062, opened 2022-03-26
Added the runningInConsole condition in PermissionServiceProvider::bo…

yuraplohov:
This PR adds app()->runningInConsole() condition to service-provider boot method. This condition allows to run some code only in CLI mode. It is a small optimisation for requests from web-server.


freekmurze:
Thank you for this. I'm going to pass on this for now, as it's a micro optimisation.

drbyte:
FYI - implemented in upcoming v6

PR #2067, opened 2022-04-13
Remove double semicolon from add_teams_fields.php.stub

morganarnel:
Double semicolon in teams_field migration ;;)
Credit: @CraicOverflow89


freekmurze:
Thanks!

PR #2080, opened 2022-05-01

hannah-massey:


erikn69:
You are droping support for laravel 7 without changing comporser.json and also you forget
https://github.com/spatie/laravel-permission/blob/main/database/migrations/add_teams_fields.php.stub

freekmurze:
I'm going to keep things as they are for now.

Issue #1260, opened 2019-10-24
Uuid not working


LucasTeodoro:
Illuminate\Database\QueryException  : SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: uuid = integer
LINE 1: ...les"."role_id" where "model_has_roles"."model_id" in (7) and...
                                                             ^
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts. (SQL: select "roles".*, "model_has_roles"."model_id" as "pivot_model_id", "model_has_roles"."role_id" as "pivot_role_id", "model_has_roles"."model_type" as "pivot_model_type" from "roles" inner join "model_has_roles" on "roles"."id" = "model_has_roles"."role_id" where "model_has_roles"."model_id" in (7) and "model_has_roles"."model_type" = App\Models\System\User)

sergiz:
Did you do appropriate modifications (migrations and configuration) to actually make it working?

Here are some ideas
https://github.com/spatie/laravel-permission/issues/948

LucasTeodoro:
Yes i made.  

I found a solution by declaring 
protected $keyType = "string";

sergiz:
So your issue was not even realted to this package, but to Laravel in general.

PR #2089, opened 2022-05-06
[V6] Full uuid/guid/ulid support

### LINKED ISSUE #489: Using UUID/GUID for primary key on user table
Not sure where to put this, but I am successfully using this using my UUID user IDs, in place of the integer IDs.

I accomplished this by replacing every instance of `$table->morphs('model')` with:
```php
$table->uuid('model_id');
$table->string('model_type');
```

I am using GUIDs because my main authentication is linked with Active Directory (via [Adldap2/Adldap2-Laravel](https://github.com/Adldap2/Adldap2-Laravel)), and I need to use the AD primary keys as my primary keys to ensure integrity.

Not sure if this is the best place to put this, but I wanted to document this somewhere in case anyone else used this setup.

### LINKED ISSUE #793: Using UUid instead of Native ID's
Hi,

I have updated my migrations to accept uuids instead of ids but when giving roles to users or giving permissions to roles, these two pieces occur issue because uuids appear like names:

**Spatie\Permission\Traits\HasRoles**
```
        if (is_string($role)) {
            return app(Role::class)->findByName($role, $this->getDefaultGuardName());
        }
```
and
**Spatie\Permission\Traits\HasPermissions**
```
        if (is_string($permissions)) {
            return app(Permission::class)->findByName($permissions, $this->getDefaultGuardName());
        }
```

So I added a custom trait to my Role model as below to solve this issue:

```
    namespace App\Traits\Acl;

    use App\Models\Role;
    use App\Models\Permission;

    trait UuidInsteadOfId
    {
        protected function getStoredRole($role) : Role
        {
            if (strlen($role) == 36) {
                return app(Role::class)
                    ->where('id', $role)
                    ->where('guard_name', $this->getDefaultGuardName())
                    ->first();
            }

            if (is_numeric($role)) {
                return app(Role::class)->findById($role, $this->getDefaultGuardName());
            }

            if (is_string($role)) {
                return app(Role::class)->findByName($role, $this->getDefaultGuardName());
            }

            return $role;
        }

        protected function getStoredPermission($permissions)
        {
            if (strlen($permissions) == 36) {
                return app(Permission::class)
                    ->where('id', $permissions)
                    ->where('guard_name', $this->getDefaultGuardName())
                    ->first();
            }

            if (is_numeric($permissions)) {
                return app(Permission::class)->findById($permissions, $this->getDefaultGuardName());
            }

            if (is_string($permissions)) {
                return app(Permission::class)->findByName($permissions, $this->getDefaultGuardName());
            }

            if (is_array($permissions)) {
                return app(Permission::class)
                    ->whereIn('name', $permissions)
                    ->whereIn('guard_name', $this->getGuardNames())
                    ->get();
            }

            return $permissions;
        }
    }

```

I am sure ` if (strlen($permissions) == 36) {` is not the best practice, but as a temporary solution it worked for me.

Thanks for this great work.

### LINKED ISSUE #948: How can I use UUID
Hi,

I'm using UUID as primary key for all of my project tables, how can i disable autoincrement and allow roles and permissions primary key ID  to be fillable? i have managed to do it on other tables but i'm stacked with these two.

with regards

### LINKED ISSUE #1366: UUID bug (ex: Testing race condition)
Hey,

the old issue text is below for reference but I've debugged the issue further and it seems to stem from a problem using UUIDs as primary keys.

I have been getting random test failures on tests that involve a user having permissions. The test case (not really relevant anymore, see below) expects to be able to access a page (http status 200) which works most of the time but because of problems sometimes returns 403.

Long story short, after hours of logging everything and anything I could find in our code and the laravel-permissions code I've started logging the queries and found the culprit:

```
    [10] => Array
        (
            [query] => select `permissions`.*, `model_has_permissions`.`model_uuid` as `pivot_model_uuid`, `model_has_permissions`.`permission_id` as `pivot_permission_id`, `model_has_permissions`.`model_type` as `pivot_model_type` from `permissions` inner join `model_has_permissions` on `permissions`.`id` = `model_has_permissions`.`permission_id` where `model_has_permissions`.`model_uuid` in (0) and `model_has_permissions`.`model_type` = ?
            [bindings] => Array
                (
                    [0] => App\User
                )

            [time] => 0.62
        )
```
(The interesting part is the ```where `model_has_permissions`.`model_uuid` in (0)```)
We've had that problem before in our core application: Laravel seems to cast the UUIDs to integer which returns really weird results which funnily enough work _most_ of the time. But sometimes it just breaks. The solution is to set `public $keyType = 'string';` on the models, which is something I can not do on the Permission and Role models.

One solution might be to to set $keyType on the models and do additional changes to the ones described in https://docs.spatie.be/laravel-permission/v3/installation-laravel/ to the migrations to change all keys to UUID. That could solve the problem.

Has anybody had the same issue or knows how to properly deal with this?

~~Hey,~~

~~I've been getting random test failures on tests that involve a user having permissions. The following test case sometimes fails (it should return 200 but gets 403 because somehow the permissions seem to be missing from the user):~~

```
public function test_when_calling_styleguide_as_a_user_with_the_styleguide_permission_then_it_should_show_the_styleguide()
    {
        // arrange
        $faker = Factory::create();

        app()[PermissionRegistrar::class]->forgetCachedPermissions();

        Permission::create(['name' => 'styleguide']);

        $this->app->make(PermissionRegistrar::class)->registerPermissions();

        $user = User::create([
            'email' => $faker->unique()->safeEmail,
            'email_verified_at' => now()->subYear(),
            'tos_accepted_at' => now(),
        ]);
        $user->givePermissionTo('styleguide');

        // act
        $response = $this->actingAs($user)
            ->get(route('styleguide'));

        // assert
        $response->assertStatus(200);
    }
```

The route called is defined as follows:
```
Route::middleware(['auth', 'permission:styleguide'])->group(function () {
    Route::get('styleguide', 'StaticPageController@styleguide')->name('styleguide');
});
```

and the controller method is a simple
```
public function styleguide()
{
    return view('welcome');
}
```

~~Does anybody have an idea what's going wrong here? I've hit a wall.~~

### LINKED ISSUE #1420: unable to use EfficientUuid with this package
hello there:
I'm using EfficientUuid in my application and when I tried to do `$user->assignRole($someRole)` I get the following error:
`SQLSTATE[22001]: String data, right truncated: 1406 Data too long for column 'model_uuid' at row 1 (SQL: insert into `model_has_roles` (`mo
del_type`, `model_uuid`, `role_id`) values (App\User, f86dfee1-2eaf-4656-8d8f-c696c969da26, 4))
`

as it is trying to insert the string representation of the UUID instead of the binary representation.. can you please suggest how to fix this?

thank you,

### LINKED ISSUE #1434: SQLSTATE[42883]: Undefined function: 7 ERROR: operator does not exist: uuid = integer
When I try create a Role:
image

In the Roles and Permissions models, I put to force these attributes but still returns the error mentioned above.
image

image

When creating, failed when looking for role relationship with permission model
image

### LINKED ISSUE #1477: UUIDs and key contraints
I'm using UUID's in my application and have changed the permissions migration to be
<details>
<summary>Migration</summary>

```
<?php

use Illuminate\Support\Facades\Schema;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Database\Migrations\Migration;

class CreatePermissionTables extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        $tableNames = config('permission.table_names');
        $columnNames = config('permission.column_names');

        if (empty($tableNames)) {
            throw new \Exception('Error: config/permission.php not found and defaults could not be merged. Please publish the package configuration before proceeding.');
        }

        Schema::create($tableNames['permissions'], function (Blueprint $table) {
            $table->uuid('id')->primary();
            $table->string('name');
            $table->string('guard_name');
            $table->timestamps();
        });

        Schema::create($tableNames['roles'], function (Blueprint $table) {
            $table->uuid('id')->primary();
            $table->string('name');
            $table->string('guard_name');
            $table->timestamps();
        });

        Schema::create($tableNames['model_has_permissions'], function (Blueprint $table) use ($tableNames, $columnNames) {
            $table->uuid('permission_id');

            $table->string('model_type');
            $table->uuid($columnNames['model_morph_key']);
            $table->index([$columnNames['model_morph_key'], 'model_type'], 'model_has_permissions_model_id_model_type_index');

            $table->foreign('permission_id')
                ->references('id')
                ->on($tableNames['permissions'])
                ->onDelete('cascade');

            $table->primary(['permission_id', $columnNames['model_morph_key'], 'model_type'],
                    'model_has_permissions_permission_model_type_primary');
        });

        Schema::create($tableNames['model_has_roles'], function (Blueprint $table) use ($tableNames, $columnNames) {
            $table->uuid('role_id');

            $table->string('model_type');
            $table->uuid($columnNames['model_morph_key']);
            $table->index([$columnNames['model_morph_key'], 'model_type'], 'model_has_roles_model_id_model_type_index');

            $table->foreign('role_id')
                ->references('id')
                ->on($tableNames['roles'])
                ->onDelete('cascade');

            $table->primary(['role_id', $columnNames['model_morph_key'], 'model_type'],
                    'model_has_roles_role_model_type_primary');
        });

        Schema::create($tableNames['role_has_permissions'], function (Blueprint $table) use ($tableNames) {
            $table->uuid('permission_id');
            $table->uuid('role_id');

            $table->foreign('permission_id')
                ->references('id')
                ->on($tableNames['permissions'])
                ->onDelete('cascade');

            $table->foreign('role_id')
                ->references('id')
                ->on($tableNames['roles'])
                ->onDelete('cascade');

            $table->primary(['permission_id', 'role_id'], 'role_has_permissions_permission_id_role_id_primary');
        });

        app('cache')
            ->store(config('permission.cache.store') != 'default' ? config('permission.cache.store') : null)
            ->forget(config('permission.cache.key'));
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        $tableNames = config('permission.table_names');

        if (empty($tableNames)) {
            throw new \Exception('Error: config/permission.php not found and defaults could not be merged. Please publish the package configuration before proceeding, or drop the tables manually.');
        }

        Schema::drop($tableNames['role_has_permissions']);
        Schema::drop($tableNames['model_has_roles']);
        Schema::drop($tableNames['model_has_permissions']);
        Schema::drop($tableNames['roles']);
        Schema::drop($tableNames['permissions']);
    }
}
```
</details>

Sadly now, when assigning a role to a User, I started getting this SQL error.

``` 
SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`herpaderp`.`model_has_roles`, CONSTRAINT `model_has_roles_role_id_foreign` FOREIGN KEY (`role_id`) REFERENCES `roles` (`id`) ON DELETE CASCADE) (SQL: insert into `model_has_roles` (`model_id`, `model_type`, `role_id`) values (19bb2b17-6298-42a9-9c53-ca186b12393e, App\User, 0))
```
What did I miss?

### LINKED ISSUE #1537: Illegal offset type when assigning Permission into Role using UUID
I have implemented [laravel-permission][1] to custom my Model using UUID primary key. Then I created a seed:

```
$viewemployee = Permission::create(['id' => Str::uuid(), 'name' => 'view employee']);
$A = Role::create(['id' => Str::uuid(), 'name' => 'A']);
```

But, when I'm trying to assign the permission to the role using:
```
$A->givePermissionTo($viewemployee);
```
It shows error like this when I run the seed
>    ErrorException 
>  Illegal offset type
>  at vendor/laravel/framework/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivot>Table.php:139

Then, I tried another method to assign:

```
$viewemployee->assignRole($A);
```
But, it shows another error:

>   Illuminate\Database\QueryException 
>  SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`db`.`role_has_permissions`, CONSTRAINT `role_has_permissions_role_id_foreign` FOREIGN KEY (`role_id`) REFERENCES `roles` (`id`) ON DELETE CASCADE) (SQL: insert into `role_has_permissions` (`permission_id`, `role_id`) values (8f98272b-cbc3-459b-ab23-fc8fa86dd199, 0))

I have followed the documentation about UUID and Extending correctly, yet I still can't figure it out what I miss so that I can't run my seed.

`role` and `permission` data are successfully inserted, it seems like have a trouble when trying to insert to `role_has_permissions` pivot table.

Here's my migration for `permissions`, `roles` and `role_has_permissions` tables:
```
Schema::create($tableNames['permissions'], function (Blueprint $table) {
      $table->uuid('id');
      $table->string('name');
      $table->string('guard_name');
      $table->timestamps();

      $table->primary('id');
});

Schema::create($tableNames['roles'], function (Blueprint $table) {
     $table->uuid('id');
     $table->string('name');
     $table->string('guard_name');
     $table->timestamps();

     $table->primary('id');
});

Schema::create($tableNames['role_has_permissions'], function (Blueprint $table) use ($tableNames) {
    $table->uuid('permission_id');
    $table->uuid('role_id');

    $table->foreign('permission_id')
        ->references('id')
        ->on($tableNames['permissions'])
        ->onDelete('cascade');

    $table->foreign('role_id')
        ->references('id')
        ->on($tableNames['roles'])
        ->onDelete('cascade');

    $table->primary(['permission_id', 'role_id'], 'role_has_permissions_permission_id_role_id_primary');
});
```

  [1]: https://docs.spatie.be/laravel-permission/v3/advanced-usage/uuid/

### LINKED ISSUE #1561: Strict typing on Role Contract prevents looking Role up by UUID
https://github.com/spatie/laravel-permission/blob/master/src/Contracts/Role.php#L38

this is forcing the `$id` to be an int, and when you swap to UUIDs this function is no longer usable.  You can't even replace the implementation bc that would have to implement this contract as well

throws error:
```
Argument 1 passed to Spatie\\Permission\\Models\\Role::findById() must be of the type int, string given
```

### LINKED ISSUE #1567: Docs issue - UuidTrait can't work
The [UUID documentation page](https://spatie.be/docs/laravel-permission/v3/advanced-usage/uuid#models) suggests creating a trait to define common values needed for UUID primary keys.

Unfortunately, the trait shown can't possibly work.  Laravel's default value for the $incrementing property is true, while the trait defines it as false.   Per the [PHP docs on trait property conflict resolution](https://www.php.net/manual/en/language.oop5.traits.php#language.oop5.interfaces.examples.ex1), this behavior simply isn't supported, and running it proves this.  I get the following error output when attempting to seed a user object:
```
  Whoops\Exception\ErrorException

  Illuminate\Database\Eloquent\Model and App\Traits\HasUuidKey define the same property ($incrementing) in the composition of App\Models\User. However, the definition differs and is considered incompatible. Class was composed
```

### LINKED ISSUE #1584: Implementing UUIDs as a Primary Key has an Issue on Assigning User to a Certain Role
Hi,
I've just tried to integrate this package into the Laravel 8 project using UUIDs implementation as a Primary Key. Everything looks fine until I try to assign a user to a certain role, e.g 'Admin'. My code:

`
$user = $request->user();
$assignUserToRole = $user->assignRole('Admin');
`

and I got the following error:
`SQLSTATE[23000]: Integrity constraint violation: 1452 Cannot add or update a child row: a foreign key constraint fails (`laravel8_bootstrapdb`.`model_has_roles`, CONSTRAINT `model_has_roles_role_id_foreign` FOREIGN KEY (`role_id`) REFERENCES `roles` (`id`) ON DELETE CASCADE) (SQL: insert into `model_has_roles` (`model_id`, `model_type`, `role_id`) values (91b57341-5e99-4057-9b94-826a76579d16, App\Models\User, 91))`

My own Role model that extends the Spatie's Role model:
`<?php

namespace App\Models;

use Illuminate\Database\Eloquent\Factories\HasFactory;
use Illuminate\Support\Str;

use App\Models\Concerns\UsesUuid;

class Role extends \Spatie\Permission\Models\Role
{
    use HasFactory, UsesUuid;

    protected $fillable = [
        'id',
        'name',
        'guard_name',
    ];

    protected $primaryKey = 'id';
    public $incrementing = false;

    protected $casts = [
        'id' => 'string'
    ];
}
`

It looks like the String casting doesn't work like expected. Because the role id still an integer value. And when I tried to implement the String casting in the original Spatie's Role model, everything works fine then. But, I wonder why it works fine when assigning permission to a user which is I thought has a similar case.

So, if you guys know how to solve this issue without touching the original package please let me know. Thanks.

### LINKED ISSUE #1623: cant use syncPermissions or syncRoles while using Uuid
a follow up for https://github.com/spatie/laravel-permission/issues/793 which is about 2 years old.

any plans for fixing it ?

### LINKED ISSUE #2083: typeError Illegal offset type
I get an typeError illegal offset type when trying to assignRole to a user in the seeder. I have changed the default migrations up a bit to support UUID's Is this something I have wrote incorrectly? The roles and permissions do have a uuid in them when i seed them.

Error i get when i try to seed
Screenshot_26
Seeder
Screenshot_27
Migrations
Screenshot_28
Screenshot_29
Screenshot_30
User model
Screenshot_31

erikn69:
- UUID/GUID/ULID full support on all methods
- Custom UUID Models on tests

keeps the same operation as always

----
https://github.com/spatie/laravel-permission/pull/1909#issuecomment-960652059
> I see a new migration in there,

Migration removed, and now use the actual migration with a replace string as temporary migration

>Could you please break this PR up in smaller ones that are easier to review?

This is the smaller piece of code

----
Closes #1623
Closes #1584
Closes #1567
Closes #1561
Closes #1537
Closes #1477
Closes #1420
Closes #1366
Closee #1260
Closes #948
Closes #2043
Closes #1854
Closes #2083
Closes #1434
Closes #489
Closes #793

----
**UPDATE:** Rebased and ULIDs support added 

[Simfony/Component/Uid/Uuid::isValid()](https://github.com/symfony/symfony/blob/ea7cc20cae35db70717d06328140e10b884eb49e/src/Symfony/Component/Uid/Uuid.php#L143-L145)
[Simfony/Component/Uid/Ulid::isValid()](https://github.com/symfony/symfony/blob/ea7cc20cae35db70717d06328140e10b884eb49e/src/Symfony/Component/Uid/Ulid.php#L46-L57)

----
**Maybe change data type on `findById` is a breaking change for somebody overwriting that function, 
if that is true, maybe for V6 release**

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -134,7 +134,7 @@ trait HasPermissions
             if ($permission instanceof Permission) {
                 return $permission;
             }
-            $method = is_string($permission) && ! \Str::isUuid($permission) ? 'findByName' : 'findById';
+            $method = is_string($permission) && ! PermissionRegistrar::isUid($permission) ? 'findByName' : 'findById';
 
             return $this->getPermissionClass()->{$method}($permission, $this->getDefaultGuardName());
         }, Arr::wrap($permissions));
@@ -152,7 +152,7 @@ trait HasPermissions
     {
         $permissionClass = $this->getPermissionClass();
 
-        if (is_string($permission) && ! \Str::isUuid($permission)) {
+        if (is_string($permission) && ! PermissionRegistrar::isUid($permission)) {
             $permission = $permissionClass->findByName(
                 $permission,
                 $guardName ?? $this->getDefaultGuardName()
@@ -204,7 +204,7 @@ trait HasPermissions
     {
         $guardName = $guardName ?? $this->getDefaultGuardName();
 
-        if (is_int($permission) || \Str::isUuid($permission)) {
+        if (is_int($permission) || PermissionRegistrar::isUid($permission)) {
             $permission = $this->getPermissionClass()->findById($permission, $guardName);
         }
 
@@ -449,7 +449,7 @@ trait HasPermissions
     {
         $permissionClass = $this->getPermissionClass();
 
-        if (is_numeric($permissions) || \Str::isUuid($permissions)) {
+        if (is_numeric($permissions) || PermissionRegistrar::isUid($permissions)) {
             return $permissionClass->findById($permissions, $this->getDefaultGuardName());
         }
 
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -134,7 +134,7 @@ trait HasPermissions
             if ($permission instanceof Permission) {
                 return $permission;
             }
-            $method = is_string($permission) ? 'findByName' : 'findById';
+            $method = is_string($permission) && ! \Str::isUuid($permission) ? 'findByName' : 'findById';
 
             return $this->getPermissionClass()->{$method}($permission, $this->getDefaultGuardName());
         }, Arr::wrap($permissions));
@@ -152,14 +152,14 @@ trait HasPermissions
     {
         $permissionClass = $this->getPermissionClass();
 
-        if (is_string($permission)) {
+        if (is_string($permission) && ! \Str::isUuid($permission)) {
             $permission = $permissionClass->findByName(
                 $permission,
                 $guardName ?? $this->getDefaultGuardName()
             );
         }
 
-        if (is_int($permission)) {
+        if (is_int($permission) || is_string($permission)) {
             $permission = $permissionClass->findById(
                 $permission,
                 $guardName ?? $this->getDefaultGuardName()
@@ -204,7 +204,7 @@ trait HasPermissions
     {
         $guardName = $guardName ?? $this->getDefaultGuardName();
 
-        if (is_int($permission)) {
+        if (is_int($permission) || \Str::isUuid($permission)) {
             $permission = $this->getPermissionClass()->findById($permission, $guardName);
         }
 
@@ -449,7 +449,7 @@ trait HasPermissions
     {
         $permissionClass = $this->getPermissionClass();
 
-        if (is_numeric($permissions)) {
+        if (is_numeric($permissions) || \Str::isUuid($permissions)) {
             return $permissionClass->findById($permissions, $this->getDefaultGuardName());
         }
 
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -104,13 +104,13 @@ class Permission extends Model implements PermissionContract
     /**
      * Find a permission by its id (and optionally guardName).
      *
-     * @param  int  $id
+     * @param  int|string  $id
      * @param  string|null  $guardName
      * @return \Spatie\Permission\Contracts\Permission
      *
      * @throws \Spatie\Permission\Exceptions\PermissionDoesNotExist
      */
-    public static function findById(int $id, $guardName = null): PermissionContract
+    public static function findById($id, $guardName = null): PermissionContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
         $permission = static::getPermission([(new static())->getKeyName() => $id, 'guard_name' => $guardName]);
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -108,11 +108,11 @@ class Role extends Model implements RoleContract
     /**
      * Find a role by its id (and optionally guardName).
      *
-     * @param  int  $id
+     * @param  int|string  $id
      * @param  string|null  $guardName
      * @return \Spatie\Permission\Contracts\Role|\Spatie\Permission\Models\Role
      */
-    public static function findById(int $id, $guardName = null): RoleContract
+    public static function findById($id, $guardName = null): RoleContract
     {
         $guardName = $guardName ?? Guard::getDefaultName(static::class);
 
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -384,4 +384,25 @@ class PermissionRegistrar
 
         $this->permissions['roles'] = [];
     }
+
+    public static function isUid($value)
+    {
+        if (! is_string($value) || empty(trim($value))) {
+            return false;
+        }
+
+        // check if is UUID/GUID
+        $uid = preg_match('/^[\da-f]{8}-[\da-f]{4}-[\da-f]{4}-[\da-f]{4}-[\da-f]{12}$/iD', $value) > 0;
+        if ($uid) {
+            return true;
+        }
+
+        // check if is ULID
+        $ulid = 26 == strlen($value) && 26 == strspn($value, '0123456789ABCDEFGHJKMNPQRSTVWXYZabcdefghjkmnpqrstvwxyz') && $value[0] <= '7';
+        if ($ulid) {
+            return true;
+        }
+
+        return false;
+    }
 }
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -83,7 +83,7 @@ trait HasRoles
                 return $role;
             }
 
-            $method = is_numeric($role) ? 'findById' : 'findByName';
+            $method = is_numeric($role) || \Str::isUuid($role) ? 'findById' : 'findByName';
 
             return $this->getRoleClass()->{$method}($role, $guard ?: $this->getDefaultGuardName());
         }, Arr::wrap($roles));
@@ -195,13 +195,13 @@ trait HasRoles
             $roles = $this->convertPipeToArray($roles);
         }
 
-        if (is_string($roles)) {
+        if (is_string($roles) && ! \Str::isUuid($roles)) {
             return $guard
                 ? $this->roles->where('guard_name', $guard)->contains('name', $roles)
                 : $this->roles->contains('name', $roles);
         }
 
-        if (is_int($roles)) {
+        if (is_int($roles) || is_string($roles)) {
             $roleClass = $this->getRoleClass();
             $key = (new $roleClass())->getKeyName();
 
@@ -325,7 +325,7 @@ trait HasRoles
     {
         $roleClass = $this->getRoleClass();
 
-        if (is_numeric($role)) {
+        if (is_numeric($role) || \Str::isUuid($role)) {
             return $roleClass->findById($role, $this->getDefaultGuardName());
         }
 
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -83,7 +83,7 @@ trait HasRoles
                 return $role;
             }
 
-            $method = is_numeric($role) || \Str::isUuid($role) ? 'findById' : 'findByName';
+            $method = is_numeric($role) || PermissionRegistrar::isUid($role) ? 'findById' : 'findByName';
 
             return $this->getRoleClass()->{$method}($role, $guard ?: $this->getDefaultGuardName());
         }, Arr::wrap($roles));
@@ -195,7 +195,7 @@ trait HasRoles
             $roles = $this->convertPipeToArray($roles);
         }
 
-        if (is_string($roles) && ! \Str::isUuid($roles)) {
+        if (is_string($roles) && ! PermissionRegistrar::isUid($roles)) {
             return $guard
                 ? $this->roles->where('guard_name', $guard)->contains('name', $roles)
                 : $this->roles->contains('name', $roles);
@@ -325,7 +325,7 @@ trait HasRoles
     {
         $roleClass = $this->getRoleClass();
 
-        if (is_numeric($role) || \Str::isUuid($role)) {
+        if (is_numeric($role) || PermissionRegistrar::isUid($role)) {
             return $roleClass->findById($role, $this->getDefaultGuardName());
         }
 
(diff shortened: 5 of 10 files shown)


src/Traits/HasPermissions.php
```diff
@@ -152,14 +152,14 @@ public function filterPermission($permission, $guardName = null)
...
 
-        if (is_string($permission)) {
+        if (is_string($permission) && ! PermissionRegistrar::isUid($permission)) {
             $permission = $permissionClass->findByName(
                 $permission,
                 $guardName ?? $this->getDefaultGuardName()
             );
         }
 
-        if (is_int($permission)) {
```
drbyte, line 162: Should is_string be isUid here?

src/Traits/HasPermissions.php
```diff
@@ -152,14 +152,14 @@ public function filterPermission($permission, $guardName = null)
...
+        if (is_int($permission) || is_string($permission)) {
```
drbyte, line 162: Should is_string be isUid here?
And in HasRoles.

erikn69: Not because on line 155
```php 
if (is_string($permission) && ! PermissionRegistrar::isUid($permission)) {
```
So, we already check if is string and not uuid on `findByName`, if it still is string means that it is an uuid for `findById`

drbyte:
Thanks!

drbyte:
Merging to v6/dev-master

PR #2091, opened 2022-05-06
Support custom fields on cache

### LINKED ISSUE #1836: Columns not available for extended Permissions model

### LINKED ISSUE #2016: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'name' in 'field list'

erikn69:
**Not breaking change**
Closes #1836
Closes #2046
Closes #2016

Set all models fields on cache, still using alias and arrays for smaller cache, added automatic alias for avoid using `i,n,g`, 
`i,n,g` causes a problem when you add a custom field like `type`, `description`, `label`

**Before:**
https://github.com/spatie/laravel-permission/blob/b71df07f3b5db4886657b4c17244d66ba6796c1a/src/PermissionRegistrar.php#L282-L285
https://github.com/spatie/laravel-permission/blob/b71df07f3b5db4886657b4c17244d66ba6796c1a/src/PermissionRegistrar.php#L341-L345
**After:** Less code, easier to understand
```php
$permissions = $this->getPermissionClass()->select()->with('roles')->get()
///
$role = $roleInstance->newFromBuilder($this->aliasedArray($item));
$this->cachedRoles[$role->getKey()] = $role;
```

------
**Results:** 
```php
array:358 [▼
  0 => array:8 [▼
    "i" => 197
    "n" => "view_audit"
    "g" => "web"
    "l" => "Audit"
    "e" => null
    "y" => "A"
    "r" => array:3 [▼
      0 => array:5 [▼
        "i" => 0
        "t" => null
        "n" => "System"
        "g" => "web"
        "y" => "A"
      ]
```
`1631207353a:358:{i:0;a:8:{s:1:"i";i:197;s:1:"n";s:10:"view_audit";s:1:"g";s:3:"web";s:1:"l";s:10:"Audit";s:1:"e";N;s:1:"y";s:1:"A";s:1:"r";a:3:{i:0;a:5:{s:1:"i";i:0;s:1:"t";N;s:1:"n";s:8:"System";s:1:"g";s:3:"web";s:1:"y";s:1:"A";}i:`

Added test `it_can_use_custom_fields_from_cache`, checked not over hydration

image

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -52,6 +52,12 @@ class PermissionRegistrar
     /** @var array */
     private $cachedRoles = [];
 
+    /** @var array */
+    private $alias = [];
+
+    /** @var array */
+    private $except = [];
+
     /**
      * PermissionRegistrar constructor.
      *
@@ -174,18 +180,20 @@ class PermissionRegistrar
         });
 
         // fallback for old cache method, must be removed on next mayor version
-        if (! isset($this->permissions['permissions'])) {
+        if (! isset($this->permissions['alias'])) {
             $this->forgetCachedPermissions();
             $this->loadPermissions();
 
             return;
         }
 
+        $this->alias = $this->permissions['alias'];
+
         $this->hydrateRolesCache();
 
         $this->permissions = $this->getHydratedPermissionCollection();
 
-        $this->cachedRoles = [];
+        $this->cachedRoles = $this->alias = $this->except = [];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


freekmurze:
Thanks!

PR #2092, opened 2022-05-09
Support custom primary key names on models

erikn69:
Sometimes is needed use other primary key name than `id`, like when somebody uses UUID format, normally use `uuid` as name for primary keys, or when somebody implements "laravel-permission" in a existing project (#1805)


freekmurze:
Thanks!

PR #2094, opened 2022-05-13
Fix UuidTrait on uuid doc page

abhishekpaul:
Boot method having protected it should be public and no need to call parent boot method


freekmurze:
Thanks!

PR #2095, opened 2022-05-16
for more connection database

ihyme:


erikn69:
Maybe it's a breaking change, because not everybody uses mysql as database(i don't), you could publish and customize migration as you wish

Also, if you want to contribute, at least put a fallback for everybody without `'connection'=>'mysql'` on their config
```diff
-config('permission.connection');
+config('permission.connection') ?: config('database.default');
```

ihyme:
thank you for enlightening me

PR #2096, opened 2022-05-16
Custom primary keys tests(Only tests)

erikn69:
**For avoid breaking this again on the future**

- Complement of  #2092

-----
Also using variables for migrations as starting point for future anonymous migrations support

```php
protected static $migration;
protected static $customMigration;
```


freekmurze:
Thanks!

PR #2098, opened 2022-05-17
Avoid calling the config helper in the role/perm model constructor

### LINKED ISSUE #2097: Fatal error: Uncaught ReflectionException: Class "config" does not exist
If I want to adhere to the principle of dependency injection, and add `Spatie\Permission\Models\Role` to the __construct of any class, I will get an error `Fatal error: Uncaught ReflectionException: Class "config" does not exist`

Because the helper `config()` is not initialized yet at the moment.

A simple example:

```php

namespace App\Services\Permission;

class PermissionService
{
    public function __construct(
        private \Spatie\Permission\Models\Role $role,
    ) {
    }
}
```

Now in the terminal, try typing a command, for example `php artisan cache:clear`, and you will get an error.
I have a solution, I will offer it in PR ))

**Versions**
laravel/framework -  v8.83.11
spatie/laravel-permission - 5.5.3 
PHP 8.0.8 
Windows 11

adiafora:
I described the problem in #2097 with initializing the config.
I propose a solution.

Fixes #2097

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -22,8 +22,6 @@ class Permission extends Model implements PermissionContract
 
     public function __construct(array $attributes = [])
     {
-        $attributes['guard_name'] = $attributes['guard_name'] ?? config('auth.defaults.guard');
-
         parent::__construct($attributes);
 
         $this->guarded[] = $this->primaryKey;
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -22,8 +22,6 @@ class Role extends Model implements RoleContract
 
     public function __construct(array $attributes = [])
     {
-        $attributes['guard_name'] = $attributes['guard_name'] ?? config('auth.defaults.guard');
-
         parent::__construct($attributes);
 
         $this->guarded[] = $this->primaryKey;


erikn69:
What happend with **Permission** model ?
https://github.com/spatie/laravel-permission/blob/8a69aaddbf91810b0175b05a1acb9b3410bff68e/src/Models/Permission.php#L25

adiafora:
@erikn69 I think it needs fixing too. I'll do it

erikn69:
> Could you test with facade `\Config::get('auth.defaults.guard')`, ([Config.php](https://laravel.com/api/9.x/Illuminate/Support/Facades/Config.html))

erikn69:
https://github.com/spatie/laravel-permission/issues/2097#issuecomment-1129600432
Ok , i did test your solution and it don't do anything, look at this example
```php
class Kernel extends ConsoleKernel
{
    public function __construct(
        Application $app,
        Dispatcher $events,
        private \Spatie\Permission\Models\Role $role
    ){
        parent::__construct($app,$events);
        var_dump($role->toArray(), (new \Illuminate\Config\Repository())->get('auth'));
    }
}
```
From this i getting
```batch
php artisan cache:clear
array(1) {   
  ["guard_name"]=> NULL          # this is $role class
}
NULL                             # this is Config\Repository()
Application cache cleared!
```
----
It seems that accessing `Config\Repository` before container creation does not get any config values, it could be the same as only you do
```diff
-$attributes['guard_name'] = $attributes['guard_name'] ?? config('auth.defaults.guard');
+$attributes['guard_name'] = $attributes['guard_name'] ?? 
+                            (function_exists('config') ? config('auth.defaults.guard') : null);
```

erikn69:
https://github.com/spatie/laravel-permission/issues/2097#issuecomment-1129600432
Ok , i did test your solution and it don't do anything, look at this example
```php
class Kernel extends ConsoleKernel
{
    public function __construct(
        Application $app,
        Dispatcher $events,
        private \Spatie\Permission\Models\Role $role
    ){
        parent::__construct($app,$events);
        var_dump($role->toArray(), (new \Illuminate\Config\Repository())->get('auth'));
    }
}
```
From that i am getting
```batch
php artisan cache:clear
array(1) {   
  ["guard_name"]=> NULL          # this is $role class
}
NULL                             # this is Config\Repository()
Application cache cleared!
```
It seems that accessing `Config\Repository` before container creation does not get any config values

----
**I think that we can delete that line without problems, it seems to me that those lines are forgotten code from previous versions**

I did run test without those lines and everything works
```batch
PHPUnit 9.5.20

Runtime:       PHP 8.1.2
Configuration: /laravel-permission/phpunit.xml.dist

...............................................................  63 / 414 ( 15%)
............................................................... 126 / 414 ( 30%)
............................................................... 189 / 414 ( 45%)
............................................................... 252 / 414 ( 60%)
............................................................... 315 / 414 ( 76%)
............................................................... 378 / 414 ( 91%)
....................................                            414 / 414 (100%)

Time: 00:07.732, Memory: 48.00 MB

OK ( 414 tests, 928 assertions)
```

adiafora:
@erikn69  I did. Could you take a look

erikn69:
>I think that we can delete that line without problems,
Read again

erikn69:
you forgot to delete this
```diff
-use Illuminate\Config\Repository as Config;
+
```

erikn69:
you forgot to delete this
```diff
-use Illuminate\Config\Repository as Config;
+
```

adiafora:
ok, I did

adiafora:
Did

angeljqv:
@freekmurze ping

JonPurvis:
Is it possible that this change is the cause of https://github.com/spatie/laravel-permission/issues/2218?

I've switched by app from 5.5.9 to 5.5.10 and the latter now has the above issue, switching back to 5.5.9 resolves it.

PR #2111, opened 2022-06-07
On WildcardPermission class use static instead of self for extending

erikn69:
- https://github.com/spatie/laravel-permission/discussions/2110


freekmurze:
Thanks!

PR #2116, opened 2022-06-16
Use `getKey`, `getKeyName` instead of `id`

erikn69:
Same as others, more abstract on custom primary key names
https://github.com/spatie/laravel-permission/blob/96466b07c9a416fcc095a9097e87131e082e60af/src/Traits/HasPermissions.php#L288


freekmurze:
Thanks!

PR #2117, opened 2022-06-19
[PHP 8.2] Fix `${var}` string interpolation deprecation

Ayesh:
PHP 8.2 deprecates `"${var}"` string interpolation pattern.
This fixes the only such occurrence in `spatie/laravel-permission` package.

 - [PHP 8.2: `${var}` string interpolation deprecated](https://php.watch/versions/8.2/${var}-string-interpolation-deprecated)
 - [RFC](https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation)


freekmurze:
Thanks!

PR #2139, opened 2022-07-14
Update role-permission.md

drdan18:
The explanation of the expected results of hasAllDirectPermissions was incorrect.


drbyte:
Thanks

Issue #1997, opened 2022-01-21
ide @hasanyrole() throws error if multiple values exist


berdjabrahamian:
In my blade template im trying to see if any of the roles a user has access too

`@hasanyrole('admin', 'super-admin')`

`super-admin` exists and the user logged in is assigned that role

`admin` also exists in the db but the logged in user is not assigned that role

When loading the page the following exception is thrown

```
InvalidArgumentException
Auth guard [super-admin] is not defined. (View: ....../src/resources/views/index.blade.php) 
```

Decided to try `dd'ing` without a blade directive
```
Auth()->user()->hasAnyRole('admin', 'super-admin')`
return true
```

removed the `dd` returns true and it works without error;

even tested it by adding a role that isnt in the database and still works as expected.

its clear that the blade directive is throwing the error.

```
prefix
<?php if(auth()->check() && auth()->user()->hasAnyRole(

suffix
)); ?>
```

Versions
- spatie/laravel-permission package version: 5.5.0
- laravel/framework package version: 8.80.0 
- PHP: 8.0.14
- Database: PGSQL 14 
- IDE: PhpStorm 2021.3.1

**To Reproduce**
- Create a user -> assign it a super-admin role
- In phpstorm add a new blade directive following the instructions on the site (https://spatie.be/docs/laravel-permission/v5/advanced-usage/phpstorm)
  - Use the @hasanyrole directive
- in your blade template add:
```
@hasanyrole('admin','super-admin')
something
@endhasanyrole
```
- refresh the page

**Expected behavior**
To return either true or false, based on the user having one of the roles assigned to it.
It should work the same way as doing the following in php or in blade

```
@if( auth()->check() && auth()->user()->hasAnyRole('admin', 'super-admin'))
something
@endif
```

**Environment (please complete the following information, because it helps us investigate better):**
Mac OS 12.1 (21C52) Monteray
M1 Pro 14 Inch

PHP 8.0.14 via homebrew using shivammathur/php/php@8.0
Pgqsql 14 via Postgres App
web server via Caddy server

erikn69:
It is not a bug, you are using it wrong
read documentation: https://spatie.be/docs/laravel-permission/v5/basic-usage/blade-directives#roles

berdjabrahamian:
Ahhhhh, multiple values need to be separated by a pipe.

For the record needs to be

@hasanyrole('role1|role2|role3');
@endhasanyrole


Thanks

erikn69:
[...]

Issue #626, opened 2018-01-18
Weird behavoiur with hasanyrole and else


Doublemedia:
Hello,

I have updated to the latest version.
We have the following check in an view:

    @hasanyrole(['mod', 'admin'])
        Hello admin
    @else
        Hello other people
    @endhasanyrole

This results in:

Parse error: syntax error, unexpected 'else' (T_ELSE), expecting end of file

When I look at the cached file I see this result:

    @hasanyrole(['mod', 'admin'])
        Hello admin
    <?php else: ?>
        Hello other people
    @endhasanyrole

When I change it in:

    @hasanyrole(['mod', 'admin'])
        Hello admin
    @endhasanyrole

It works fine.

Any other people that have experience with this?

Erwin

drbyte:
In your example without the `@else`, what is the generated PHP in the cached/compiled view file?
And, does it actually "show nothing" for "other people"? Or does it output the `@hasanyrole...` literally?

Doublemedia:
I can see the @hasanyrole tags in the source code. So it's not parsed by the PermissionServiceProvider blade compiler.

AlexVanderbist:
Hi @Doublemedia 

I just gave this a test and it seems like you can't create the array inside of the directive. It has to be either a variable or a string (optionally using pipes).

In this case the easiest fix for you would be the string notation:
```blade
@hasanyrole('mod|admin')
    Hello admin
@else
    Hello other people
@endhasanyrole
```

Good luck!

Doublemedia:
I have tested it with the pipes, but this results in the same error as before.
The parsed code in the cache is also the same as before.

The else is parsed, but the @hasanyrole isn't.

Doublemedia:
$this->app->afterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {

This is not executed. Debugging before this show the output. But in it doensn't.

Doublemedia:
Solved. Cleaning cached fixed it. I though I did this already ...

PR #2168, opened 2022-09-05
Support static arrays on blade directives

### LINKED ISSUE #309: @role is not accepting an array
He, I'm check on multiple gaurds but is not working. If I just pass through 'admin' it works.

```php
@role('superadmin', 'admin')
# you may edit this stuff
@else # this is error some times
# you may not edit this stuff
@endrole

```
In the readme [here](https://github.com/spatie/laravel-permission#using-blade-directives-with-multiple-guards) it says that it should work. But it does not.

### LINKED ISSUE #2165: Blade directive `@hasanyrole` does not work with array of role names
Using blade templates this works:
```blade
@hasanyrole('admin|manager')
        It works
@endhasanyrole
```

Unfortunately this piece of code raises an error:
```blade
@hasanyrole(['admin', 'manager'])
        This should work, too
@endhasanyrole
```

The error points to the line with `@hasanyrole` and says: `ParseError syntax error, unexpected ']', expecting ')'`

According to documentation at https://spatie.be/docs/laravel-permission/v5/basic-usage/blade-directives we should be able to pass `$collectionOfRoles`

PS. This works as well:
```blade
@if(auth()->user()->hasAnyRole(['admin', 'manager']))
        This works as well
@endif
```

**Versions**
- spatie/laravel-permission package version: 5.5.5
- laravel: 8.79.0
- PHP version: 7.4.19
- Database version: 10.6.4-MariaDB-log

**Expected behavior**
`@hasanyrole` should accept arrays

erikn69:
Closes #2165
Closes https://github.com/spatie/laravel-permission/issues/1997#issuecomment-1019023930
Closes https://github.com/spatie/laravel-permission/issues/626#issuecomment-358911209
Closes #309

This PR adds support for arrays(also expressions) on blade directives, example

```blade
@hasanyrole(['admin', 'manager'])
        
@endhasanyrole

@hasallroles(['admin', 'manager'])
        
@endhasallroles
```

Without this PR only `PIPE` an `VARIABLE's REF` are supported
https://github.com/spatie/laravel-permission/blob/dbc6c18ef4b2d097350a5faf82382564748e6c97/tests/resources/views/guardHasAllRolesPipe.blade.php#L1-L1
https://github.com/spatie/laravel-permission/blob/dbc6c18ef4b2d097350a5faf82382564748e6c97/tests/resources/views/guardHasAllRoles.blade.php#L1-L1

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -84,62 +84,53 @@ class PermissionServiceProvider extends ServiceProvider
         $this->app->bind(RoleContract::class, $config['role']);
     }
 
+    public static function bladeMethodWrapper($method, $role, $guard = null)
+    {
+        return auth($guard)->check() && auth($guard)->user()->{$method}($role);
+    }
+
     protected function registerBladeExtensions($bladeCompiler)
     {
         $bladeCompiler->directive('role', function ($arguments) {
-            list($role, $guard) = explode(',', $arguments.',');
-
-            return "<?php if(auth({$guard})->check() && auth({$guard})->user()->hasRole({$role})): ?>";
+            return "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasRole', {$arguments})): ?>";
         });
         $bladeCompiler->directive('elserole', function ($arguments) {
-            list($role, $guard) = explode(',', $arguments.',');
-
-            return "<?php elseif(auth({$guard})->check() && auth({$guard})->user()->hasRole({$role})): ?>";
+            return "<?php elseif(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasRole', {$arguments})): ?>";
         });
         $bladeCompiler->directive('endrole', function () {
             return '<?php endif; ?>';
         });
 
         $bladeCompiler->directive('hasrole', function ($arguments) {
-            list($role, $guard) = explode(',', $arguments.',');
-
-            return "<?php if(auth({$guard})->check() && auth({$guard})->user()->hasRole({$role})): ?>";
+            return "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasRole', {$arguments})): ?>";
         });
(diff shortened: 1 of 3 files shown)


src/PermissionServiceProvider.php
```diff
@@ -84,62 +84,53 @@ protected function registerModelBindings()
...
         });
         $bladeCompiler->directive('endhasallroles', function () {
             return '<?php endif; ?>';
         });
 
         $bladeCompiler->directive('unlessrole', function ($arguments) {
-            list($role, $guard) = explode(',', $arguments.',');
-
-            return "<?php if(!auth({$guard})->check() || ! auth({$guard})->user()->hasRole({$role})): ?>";
+            return "<?php if(! \\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasRole', $arguments)): ?>";
```
drbyte, line 126: Should `$arguments` here be `{$arguments}` like the rest?

tests/BladeTest.php
```diff
@@ -28,12 +28,12 @@ public function all_blade_directives_will_evaluate_false_when_there_is_nobody_lo
         $roles = [$role];
         $elserole = 'na';
 
-        $this->assertEquals('does not have permission', $this->renderView('can', ['permission' => $permission]));
+        $this->assertEquals('does not have permission', $this->renderView('can', compact('permission')));
```
drbyte, line 31: I 'get' why you replaced this line with `compact()`, but perhaps it can help add clarity if we keep the original line AND add `compact()`. I know sometimes people read tests for examples, and showing both syntaxes would confirm that both can be used.
Or maybe I'm being pedantic?  :)

erikn69: Both can be used, but  I change it for consistence, I'm going to revert the change

src/PermissionServiceProvider.php
erikn69: Works the same, but i'm going to change that

erikn69: Resolved

PR #2177, opened 2022-09-14
Improve permission middleware: foreach->canany

cesarMtorres:
using canany instead foreach
and other improve
is posible this change???


fredsal:
> using canany instead foreach

`canany`  Yes, 
but using route's name as permission, Not
 `$permission = $request->route()->getName();`

.github/workflows/main.yml
```diff
@@ -0,0 +1,36 @@
+# This is a basic workflow to help you get started with Actions
+
+name: CI
+
```
drbyte, line 4: This `.github/workflows/main.yml` file you added doesn't seem relevant. Please remove.

drbyte (changes requested):

drbyte:
@cesarMtorres Thanks for the contribution.

Unfortunately `canany()` only works since Laravel 8.  This package is currently still compatible with Laravel 6 and 7.

We may consider your change for a future version which drops support for L6 and L7.

cesarMtorres:
@drbyte thanks for the feedback

drbyte:
@cesarMtorres FYI your `canany` change has been implemented in the upcoming v6

PR #2179, opened 2022-09-17
Docs: Add note about non-standard User models

androidacy-user:
This may lead to some confusing results for users who didn't delve into the advanced usage section. Put a note in Prerequisites about the caveat.


drbyte:
Thanks

PR #2194, opened 2022-09-30
Fix returning all roles instead of the assigned

### LINKED ISSUE #1829: hasRole returns true on every role when used inside user model but works perfect outside model like in controller or policy
`public function admin_detail()`
    `{`
        `if (request()->user()) {`
            `return request()->user()->hasRole('Admin') ? $this->hasOne(AdminDetail::class)->with('package')->withDefault() : $this->hasOne(AdminDetail::class, 'user_id', 'user_id')->with('package')->withDefault();`
        `} else {`
            `return $this->hasOne(AdminDetail::class)->withDefault();`
        `}`
    `}`

**hasRole returns true for every role even that not assign to the user**
this is code sample from my User model.
also i have tried this code to get roles of current user
`request()->user()->roles()->pluck('name', 'id')`
but it returns all roles
`#items: array:3 [
    2 => "Admin"
    3 => "Guard"
    1 => "SuperAdmin"
  ]`
  **hasRole is working perfect outside model like in controller or policy**

### LINKED ISSUE #2161: Incorrect result of $user->hasRole('admin'); within the laravel model
I have when using Role either in the blade or in the model code.
I am using Role inside the model function to return records depending on user role. For a normal user, I will return only those records that were inserted by him, and for an admin, I am returning all records.
However, when I use $user->hasRole('admin'); it returns true for all users, even if they do not have admin privileges. Any role name returns true when it is in the roles table. I printed the user id on the web page to ensure that the user did not have any database roles.
When this issue happens, blade @hasRole('admin') returns true for the same not admin logged user.

spatie/laravel-permission package version is 5.5.4.
PHP version:  7.3
Database version: 5.x

Use @hasRole('admin') in any model inside the booted() function and it will return true even if the user didn't have this role.

**Example Application**

```php 
protected static function booted()
    {
        if(!empty(Auth::user()))
        {
            parent::boot();
            $user = Auth::user();

            $user_info = User::where('id',$user->id)->with('driverInfo','clientInfo','launderInfo')->first();
   
            // echo $user_info->launderInfo;
            // die();

            if($user->hasRole('admin'))
            {
                return;
            }
            elseif($user->hasRole('launder') && !empty($user_info->launderInfo))
            {
                $launder_id = $user_info->launderInfo->id;
                static::addGlobalScope('launder_id', function (Builder $builder)  use ($launder_id)
                {
                    $builder->where('launder_id', '=', $launder_id);
                });
            }
            elseif($user->hasRole('user')  && !empty($user_info->clientInfo))
            {
                $client_id = $user_info->clientInfo->id;

                static::addGlobalScope('client_id', function (Builder $builder) use ($client_id)
                {
                    $builder->where('client_id', '=', $client_id);
                });
            }
            elseif($user->hasRole('driver') && !empty($user_info->driverInfo))
            {

                $driver_id = $user_info->driverInfo->id;
                static::addGlobalScope('driver_id', function (Builder $builder)  use ($driver_id)
                {
                    $builder->where('driver_id', '=', $driver_id);
                });
            }
            else
            {
                static::addGlobalScope('id', function (Builder $builder)
                {
                    $builder->where('id', '=', 0);
                });
            }

        }// end if user empty
        else
        {
            return true;
        }
    }
```
**Expected behavior**:

The role function should return true only when the current user is assigned to this role. and the blade @role is unaffected.

 **Additional context**
By the way, am I using the best way to filter model results depending on user roles?

### LINKED ISSUE #2190: auth:user()->roles returning every role instead of just the assigned ones
Hi,

Thank you for this great package and all other open-source packages you provide. I would love to do a PR about this myself but I just couldn't figure out the fix.

I really appreciate the time you take to reply every issue.

I have found that under certain conditions where a relationship is retrieved in a controller before loading the roles 'auth:user()->roles' returns every role not just the assigned ones.

Some people have found the same problem but could not be reproduced so far. Related issues: #2161, #1829

**Versions**
- spatie/laravel-permission package version: 5.5.5
- laravel/framework package: 9.32.0

PHP version: 8.1.1

Database version: sqlite

**To Reproduce**
This behavior occurs when a controller requires a model with a relationship before loading the roles and the relationship uses the hasRole function

web.php
```php
Route::get('/test', function(){
    
    $var = MyParent::withCount('limitedView')->get(); //not used but creates the issue
    
    return[ Auth::user()->roles, App\Models\User::find(Auth::id())->roles ];
});
```

MyParent.php
```php
class MyParent extends Model
{
    use HasFactory;

    public function children(): HasMany
    {
        return $this->hasMany(Child::class);
    }

    public function limitedView(){
        $query = $this->children();

        if(!Auth::user()->hasRole('Super-Admin')){
            $query->orWhere('age', '>=', 18);
        }

        return $query;
    }
}
```

**Basic Test**
```php
public function test_example()
    {
        app(DatabaseSeeder::class)->call(PermissionsDemoSeeder::class);

        $user = User::find(1); //Role: writer
        
        $this->be($user);

        list($rolesBefore, $rolesAfter) = $this->get('/test')->json();

        $this->assertCount(1,$rolesBefore);
        $this->assertCount(1,$rolesAfter);
        $this->assertEquals($rolesBefore[0]["name"], "writer");
        $this->assertEquals($rolesAfter[0]["name"], "writer");

    }
```
**Example Application**
A minimal repo with a test implemented can be found at: https://github.com/SIDESO/issue_demo_spatie_permissions

**Expected behavior**
Only the assigned roles should be returned.

 **Additional context**
The problem disappears adding to the user model:

```php
protected $with=['roles'];
```

 - OS: Windows10
 - Version: 21H2

erikn69:
Closes #2190
Closes #2161
Closes #1829
> This behavior occurs when a controller requires a model with a relationship before loading the roles and the relationship uses the hasRole function
> 
> web.php
> 
> ```php
> Route::get('/test', function(){
>     $var = MyParent::withCount('limitedView')->get(); //not used but creates the issue
>     return[ Auth::user()->roles, App\Models\User::find(Auth::id())->roles ];
> });
> ```
> 
> MyParent.php
> 
> ````php
> class MyParent extends Model
> {
>     public function children(): HasMany {
>         return $this->hasMany(Child::class);
>     }
> 
>     public function limitedView() {
>         $query = $this->children();
> 
>         if(!Auth::user()->hasRole('Super-Admin')){
>             $query->orWhere('age', '>=', 18);
>         }
> 
>         return $query;
>     }
> }
> ```

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -189,6 +189,8 @@ trait HasRoles
      */
     public function hasRole($roles, string $guard = null): bool
     {
+        $this->loadMissing('roles');
+
         if (is_string($roles) && false !== strpos($roles, '|')) {
             $roles = $this->convertPipeToArray($roles);
         }
@@ -248,6 +250,8 @@ trait HasRoles
      */
     public function hasAllRoles($roles, string $guard = null): bool
     {
+        $this->loadMissing('roles');
+
         if (is_string($roles) && false !== strpos($roles, '|')) {
             $roles = $this->convertPipeToArray($roles);
         }
@@ -282,6 +286,8 @@ trait HasRoles
      */
     public function hasExactRoles($roles, string $guard = null): bool
     {
+        $this->loadMissing('roles');
+
         if (is_string($roles) && false !== strpos($roles, '|')) {
             $roles = $this->convertPipeToArray($roles);
         }
@@ -311,6 +317,8 @@ trait HasRoles
 
     public function getRoleNames(): Collection
     {
+        $this->loadMissing('roles');
+
         return $this->roles->pluck('name');
     }
 


erikn69:
@drbyte I thought the same thing, but because this problem only happens when it is the first load before this package loads, I did not know how to achieve it, in the tests everything is already preloaded and it is passing, i didn't find a way to replicate this on test, and i didn't know why is happening

drbyte:
Ya, I feel your pain. 
Thinking on it more ...

PR #2204, opened 2022-10-07
Only load permissions for active guard

isaacdew:
In an app that I've been working on, I have permissions for the guard used by my API and the web guard. I noticed that all permissions were loaded on every request - regardless of whether they were registered to the guard in use. This seems like a lot of overhead for no practical value if you have multiple guards with lots of permissions.

Is something like the approach below practical? Are there cache implications or issues it'd cause with something like Laravel Octane that I'm not considering?

Thanks for taking a look!


parallels999:
> Are there cache implications or issues it'd cause with something that I'm not considering?

Yes, it would be a breaking change, normally you could assign roles and permissions of another guard from current guard, with this change it would be impossible

erikn69:
are you trying to break all the caching functionality?

isaacdew:
@parallels999 
> Yes, it would be a breaking change, normally you could assign roles and permissions of another guard from current guard, with this change it would be impossible

Is that to say there's no way to limit database queries to the current guard in cases where that's all that's needed without breaking changes? Is what you describe documented somewhere?

@erikn69 Appreciate the warm welcome 😄 But that's a good call out. I see that many of the tests are failing because of the change to the cache key. I'll see if I can get that fixed.

erikn69:
you can do it as an optional setting so that we are not forced to use it that way
see `https://github.com/spatie/laravel-permission/pull/2203/files`
Also, you need to add some tests for the new functionality.

isaacdew:
Update - my latest commit fixes existing tests. However, I still need to implement tests for this change.

src/PermissionRegistrar.php
```diff
@@ -312,7 +319,8 @@ private function getSerializedPermissionsForCache()
     {
         $this->except = config('permission.cache.column_names_except', ['created_at','updated_at', 'deleted_at']);
 
-        $permissions = $this->getPermissionClass()->select()->with('roles')->get()
+        $permissions = $this->getPermissionClass()->select()->with('roles')
+            ->when($this->activeGuard, fn($query) => $query->where('guard_name', $this->activeGuard))->get()
```
drbyte, line 323: FYI, this package is still compatible with PHP 7.3, but the `fn()` syntax is not. That's why the tests are failing.

PR #2211, opened 2022-10-14
add getRoleName function to get only single role name as string

SAM2904:
* Add new getRoleName function

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -314,17 +314,20 @@ trait HasRoles
         return $this->roles->pluck('name');
     }
 
-    /** Return only single rolename, developer don't need to use an arr key always.
-     * @param integer $key always sets to 0.
-     * @return String $role_name
+    /** Return only single rolename, developer don't need to use an arr index always.
+     * @param integer $index always sets to 0.
+     * @return string $role_name
+     * OR
+     * @return void
      */
-    public function getRoleName($key = 0): String
+    public function getRoleName($index = 0)
     {
-        if (FALSE === is_int($key)) {
+        if (FALSE == is_int($index)) {
             trigger_error('function getRoleName expected Argument 1 to be Integer', E_USER_WARNING);
         }
-        if (is_int($key)) {
-            return isset($this->roles[$key]->name) ? $this->roles[$key]->name : '';
+        if (array_key_exists($index, $this->roles)) {
+            $role = $this->roles[$index];
+            return isset($role->name) ? $role->name : '';
         }
     }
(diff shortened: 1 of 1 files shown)


SAM2904:
This function will helps the developers to fetch the user role name without using  any array index key after function name.
It also helps to not trigger the error when rolename  is not exist at the specified index key (like as getRoleNames()[$key] function)

parallels999:
>Add new getRoleName function 

You have to add tests for `getRoleName` function,
also function name must be more specific, using key means by key or by index?

> Improve code readability

Looks worst to me

fredsal:
horrible

src/Traits/HasRoles.php
```diff
@@ -314,6 +314,20 @@ public function getRoleNames(): Collection
...
+    /** Return only single rolename, developer don't need to use an arr key always.
+     * @param integer $key always sets to 0.
+     * @return String $role_name
+     */
+    public function getRoleName($key = 0): String
+    {
+        if (FALSE === is_int($key)) {
+            trigger_error('function getRoleName expected Argument 1 to be Integer', E_USER_WARNING);
+        }
+        if (is_int($key)) {
```
fredsal, line 326: Why?
unnecessary

src/Traits/HasRoles.php
```diff
@@ -314,6 +314,20 @@ public function getRoleNames(): Collection
         return $this->roles->pluck('name');
     }
 
+    /** Return only single rolename, developer don't need to use an arr key always.
+     * @param integer $key always sets to 0.
+     * @return String $role_name
+     */
+    public function getRoleName($key = 0): String
+    {
+        if (FALSE === is_int($key)) {
```
fredsal, line 323: Always return a boolean, why `===`?

src/Traits/HasRoles.php
```diff
@@ -314,6 +314,20 @@ public function getRoleNames(): Collection
...
+            trigger_error('function getRoleName expected Argument 1 to be Integer', E_USER_WARNING);
+        }
+        if (is_int($key)) {
+            return isset($this->roles[$key]->name) ? $this->roles[$key]->name : '';
```
fredsal, line 327: > isset($this->roles[$key]->name)
Don't seems to be right

SAM2904: can you please review it again @fredsal 
also, can you please guide me @fredsal to improve/fix this function.

drbyte:
I'm not sure this needs to be in the package. It seems better to have this in userland application if one really needs it.

PR #2219, opened 2022-10-19
Fix undefined index guard_name

### LINKED ISSUE #2218: Undefined index: guard_name
**Describe the bug**
This afternoon I upgraded the version of this package we're using to [5.5.10](https://github.com/spatie/laravel-permission/releases/tag/5.5.10) and then started seeing an error to do with guard_name being undefined:

```
ErrorException
Undefined index: guard_name

Illuminate\Foundation\Bootstrap\HandleExceptions::handleError
vendor/spatie/laravel-permission/src/Models/Role.php:73
```

That line consists of 
```
getModelForGuard($this->attributes['guard_name']),
```

I've downgraded back to 5.5.9 for now which no longer causes the error. It would be good if someone could take a look into this though!

**Versions**
- spatie/laravel-permission package version: **5.5.10**
- illuminate/framework package version: **8.83.25**
- PHP version: **7.4**
- Database version: **MySQL 5.7**

**To Reproduce**
```
Role::with('users')->get();
```
causes the error, however
```
Role::with('permissions')->get();
````
works fine

**Expected behavior**
I'd expect roles to be returned with users that have the role

erikn69:
Closes #2218
Bug introduced on #2098
With Tests

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -64,7 +64,7 @@ class Permission extends Model implements PermissionContract
     public function users(): BelongsToMany
     {
         return $this->morphedByMany(
-            getModelForGuard($this->attributes['guard_name']),
+            getModelForGuard($this->attributes['guard_name'] ?? config('auth.defaults.guard')),
             'model',
             config('permission.table_names.model_has_permissions'),
             PermissionRegistrar::$pivotPermission,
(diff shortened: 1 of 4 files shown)


drbyte:
Thanks

PR #2227, opened 2022-10-21
[Bug] Avoid checking permissions-via-roles on `Role` model (ref `Model::preventAccessingMissingAttributes()`)

juliomotol:
This PR aims to fix an issue when trying to check for `Role`'s `hasPermissionTo()` when `Model::preventAccessingMissingAttributes()` is enabled.

The implemented fix overrides the `HasPermission::getAllPermissions()` to not check for `$this->roles` as the `Role` model will never have a relationship to itself.

### DIFF
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -4,7 +4,6 @@ namespace Spatie\Permission\Models;
 
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
-use Illuminate\Support\Collection;
 use Spatie\Permission\Contracts\Role as RoleContract;
 use Spatie\Permission\Exceptions\GuardDoesNotMatch;
 use Spatie\Permission\Exceptions\RoleAlreadyExists;
@@ -195,13 +194,5 @@ class Role extends Model implements RoleContract
         }
 
         return $this->permissions->contains($permission->getKeyName(), $permission->getKey());
-    }
-
-    /**
-     * Return all the permissions the model has, both directly and via roles.
-     */
-    public function getAllPermissions(): Collection
-    {
-        return $this->permissions->sort()->values();
     }
 }
--- a/src/Models/Role.php
+++ b/src/Models/Role.php
@@ -4,6 +4,7 @@ namespace Spatie\Permission\Models;
 
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Support\Collection;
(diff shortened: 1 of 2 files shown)


erikn69:
Why don't change `getAllPermissions()` on `src/Traits/HasPermissions.php#L308-L318`

**Try this:**
```diff
- if ($this->roles) {
+ if (method_exists($this, 'roles')) {
```
```php
public function getAllPermissions(): Collection
{
    /** @var Collection $permissions */
    $permissions = $this->permissions;

    if (method_exists($this, 'roles')) {
        $permissions = $permissions->merge($this->getPermissionsViaRoles());
    }

    return $permissions->sort()->values();
}
```

https://github.com/spatie/laravel-permission/blob/dc9c48cbff08dcc13fbe3b07d81fa6f0be3de6e3/src/Traits/HasPermissions.php#L308-L318

juliomotol:
I'm down to do that. But what about this:

```php
// HasPermissions.php
public function getAllPermissions(): Collection
{
    /** @var Collection $permissions */
    $permissions = $this->permissions;

    if (in_array(HasRoles::class, class_uses($this))) { // Check for the `HasRoles` trait
        $permissions = $permissions->merge($this->getPermissionsViaRoles());
    }

    return $permissions->sort()->values();
}
```

Let me know your thoughts.

parallels999:
> `if (method_exists($this, 'roles')) {`

Seems to be working

drbyte:
I'm inclined to think that `if (method_exists($this, 'getPermissionsViaRoles')) {` is the most semantically relevant. Certainly quite readable. 
I imagine it's equally performant.

erikn69:
>I'm inclined to think that `if (method_exists($this, 'getPermissionsViaRoles')) {` is the most semantically relevant.  

I think the same, it is a better solution

juliomotol:
That would always result to `true` as `getPermissionViaRoles()` is also defined within `HasPermissions`
https://github.com/spatie/laravel-permission/blob/dc9c48cbff08dcc13fbe3b07d81fa6f0be3de6e3/src/Traits/HasPermissions.php#L297-L303

parallels999:
@juliomotol how about `if (method_exists($this, 'roles')) {`??

juliomotol:
PR has been updated

lloricode:
while not approved this, this is temporary fix

```php
class Role extends \Spatie\Permission\Models\Role 
{

    protected bool $roles = false;
```

lloricode (approved):

juliomotol:
@drbyte Any plans on releasing this?

PR #2234, opened 2022-11-05
[Docs] Fix broken Link that link to freek's blog post

chengkangzai:
Hi Spatie Team while i passing by the docs i found out the link (https://murze.be/when-to-use-gateafter-in-laravel) is broken possibly caused by different slug was used when redirect to new domain. 

I am sure if this is the way to solve the problem as freek can just add a redirection url in freek.dev, feel free to close this PR if this solution is not acceptable 😉 

Love the package btw ❤️


drbyte:
Thanks!

PR #2235, opened 2022-11-06
[Docs] Add syncPermissions() in role-permissions.md

xorinzor:
the `syncPermissions` method on the `Role` object was missing from this page


drbyte:
Thanks!

PR #2248, opened 2022-11-17
Don't throw an exception when checking hasAllPermissions()

mtawil:
I believe this change must be the same as the `hasAnyPermission` functionality as it coded [here](https://github.com/spatie/laravel-permission/blob/main/src/Traits/HasPermissions.php#L238).
Also, throwing an error while we expect a boolean value doesn't make sense.


drbyte:
Thanks!

PR #2249, opened 2022-11-22
feat: must needed instruction add

leemonCodeShaper:
Previously not instruction is provided to user the hasRole trait. It was given in prerequisite section.


angeljqv:
Now point 9 and 10 are the same
https://github.com/spatie/laravel-permission/blob/baa6d35a5d64f522e39f4652c00cd53c4219d3fa/docs/installation-laravel.md?plain=1#L48

PR #2259, opened 2022-12-01
Normalize composer.json

patinthehat:
This PR normalizes composer.json using [ergebnis/composer-normalize](https://github.com/ergebnis/composer-normalize) to ensure that the composer.json file is valid and consistent according to the [Composer schema](https://getcomposer.org/schema.json).

_id:composer-normalization/v1_


freekmurze:
Thanks!

PR #2264, opened 2022-12-05
Added PHP 8.2

SuperDJ:


parallels999:
unnecessary, when you declare the '^8.0', include all the 8.x versions, 8.1, 8.2, 8.3 etc

parallels999:
Also is not wrong using single pipe

> We can not ever remove this or it'd break existing versions, we can only tell you `||` is preferred now :)
https://github.com/composer/composer/issues/6755#issuecomment-610363856

PR #2309, opened 2023-01-18
fix: Lazily bind dependencies

olivernybroe:
Currently the `PermissionRegistrar` is instantiated on every load of the application which takes some of the performance off the application as not everything needs the permission system. 

To fix this the calls have been moved into a `callAfterResolving` method instead.

The same was done with config, to also remove the need to load config changes when permission system is not used.

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -13,7 +13,7 @@ use Spatie\Permission\Contracts\Role as RoleContract;
 
 class PermissionServiceProvider extends ServiceProvider
 {
-    public function boot(PermissionRegistrar $permissionLoader)
+    public function boot()
     {
         $this->offerPublishing();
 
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

erikn69:
https://github.com/spatie/laravel-permission/issues/2319#issuecomment-1419666152
> > https://github.com/spatie/laravel-permission/releases/tag/5.9.0 Please look which of that PRs makes the problem?, maybe could be this #2309
> 
> It worked for me after reverting these PR changes locally

drbyte:
@olivernybroe your contribution is welcome, but breaks things. Do you mind revisiting this?

olivernybroe:
@drbyte I have identified the issue. 
It seems to be related to many static values are being set when instantiating `PermissionRegistrar`. However `PermissionRegistrar` is a singleton, so those properties should not be static. 

I'll create a pr fix for that.

PR #2316, opened 2023-02-03
Extract query to `getPermissionsWithRoles` method.

xiCO2k:
This PR extracts the query a different method.

This will allow to extend the `PermissionRegistrar` class and change the query.

### Example

```php

class PermissionRegistrar extends BasePermissionRegistrar
{
    protected function getPermissionsWithRoles(): Collection
    {
            ->withoutGlobalScopes()
            ->select()
            ->with('roles', fn ($query) => $query->withoutGlobalScopes())
            ->get();
    }
}

```

In this case, I can remove any global scope, applied to the `Permission` and `Role` model.

Thanks,
Francisco

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -273,6 +273,11 @@ class PermissionRegistrar
         return $this->cache->getStore();
     }
 
+    protected function getPermissionsWithRoles(): Collection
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks!

PR #2321, opened 2023-02-07
fix: Lazily bind dependencies

olivernybroe:
Here is a new version of the lazy binding solution. The reason why it was causing issues the first time around was because the `registerPermissions` method was called once `PermissionRegistrar` was resolved instead of when `Gate` was resolved.

I tested this out in a fresh installation of Laravel with the `can` directive in a file and my change here worked. but ofc. would love someone else to also validate

ping @drbyte

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -133,9 +133,9 @@ class PermissionRegistrar
      *
      * @return bool
      */
-    public function registerPermissions(): bool
+    public function registerPermissions(Gate $gate): bool
     {
-        app(Gate::class)->before(function (Authorizable $user, string $ability) {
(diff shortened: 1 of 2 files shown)


danilopinotti:
What about implementing a test to cover the last issues related?

drbyte:
@olivernybroe thanks for the update!

What are your thoughts on test suite changes to support it?

olivernybroe:
@drbyte I can't seem to find a great way of doing that, as we currently resolve all the classes in the setup method in the `TestCase` file. 

So when we get to this part to check the lazy binding, it doesn't matter as they are resolved.

drbyte:
Agreed.

PR #2324, opened 2023-02-09
refactor: Change static properties to non-static

olivernybroe:
Make the static properties on `PermissionRegistrar` non-static. 

This will make sure that if any of the properties are accessed, the cache initialization has happened first.

BREAKING CHANGE:
This is a breaking change as it will make all current migration files fail, as they are currently accessing the static properties which are not static anymore.


erikn69:
What about call config helper instead of app??

olivernybroe:
@erikn69 good idea, prob. more stable in long term, but won't solve that it's a breaking change 👍

erikn69:
> won't solve that it's a breaking change

yes, unfortunately

drbyte:
Thanks guys.
Merging so we can test it in `dev-master`

PR #2333, opened 2023-02-17
Add description column to Roles & Permissions table

hossammonir:
This PR adds the ability to define a **Description** for a role or permission. 

This is useful to provide more information about the role or permission and to display it in the UI or help you to understand what the role or permission is for in future .

```php
$role = Role::create(['name' => 'writer', 'description' => 'A writer can write articles']);
$permission = Permission::create(['name' => 'edit articles', 'description' => 'A user can edit articles']);
```


parallels999:
You have no idea what you are doing
Also, you can change whatever you want on migration, and override as you need

drbyte:
> You have no idea what you are doing

@parallels999 in the future please attempt to use less abrasive language. Everyone is at their own stage in understanding things. Your support of this package is greatly appreciated and welcome; let's keep the tone of our responses helpful as well. 😄

PR #2341, opened 2023-03-01
Avoid loss of all permissions/roles pivots on sync error

### LINKED ISSUE #2339: Syncing an incorrect permission results in loss of all permissions
**Before creating a new bug report**
Please check if there isn't a similar issue on [the issue tracker](https://github.com/spatie/laravel-permission/issues) or in [the discussions](https://github.com/spatie/laravel-permission/discussions).

**Describe the bug**
When using the `syncPermissions` function from the `HasPermissions` trait, providing an accidental wrongly named permission will result in loss of all permissions. This is caused by the fact that `syncPermissions` eagerly removes all existing permissions.

**Versions**
- spatie/laravel-permission package version: `main` at commit 490fa00b0c8286659a534d822f7e74c684c86d27
- illuminate/framework package: N/A

PHP version: any

Database version: N/A

**To Reproduce**
1. Create a model that uses the `HasPermissions` trait
2. Assign some valid permissions to the model
3. Use the `syncPermissions` helper, with a combination of any number of valid permissions and at least one invalid permission
Result: all permissions will be lost on the user

**Example Test**
https://github.com/spatie/laravel-permission/compare/main...jnoordsij:spatie-laravel-permission:permission-sync-fail

**Expected behavior**
Result: the operation fails and the permissions remain the same as before 
OR operation does partial work, removing only those permissions which are not synced.
 

**Proposed (possible) solutions**
1. Use the native Laravel `sync` function with detaching; this does not do a full rollback, but it does ensure only pivots that are no longer present are removed, instead of eagerly removing all of them
2. Copy the behavior of the Laravel `sync` function to achieve something similar here
3. Do a "manual backup" of pre-existing permissions and force restore on error
4. Skip over any non-existing permissions and throw Exception afterwards

erikn69:
Closes #2339

>**Describe the bug**
When using the syncPermissions function from the HasPermissions trait, providing an accidental wrongly named permission will result in loss of all permissions. This is caused by the fact that syncPermissions eagerly removes all existing permissions.
>
>**Steps to reproduce the behavior:**
>Create a model that uses the HasPermissions trait
Assign some valid permissions to the model
Use the syncPermissions helper, with a combination of any number of valid permissions and at least one invalid permission
Result: all permissions will be lost on the user
>
>**Example Test**
https://github.com/spatie/laravel-permission/compare/main...jnoordsij:spatie-laravel-permission:permission-sync-fail

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -343,9 +343,8 @@ trait HasPermissions
      * Returns permissions ids as array keys
      *
      * @param  string|int|array|\Spatie\Permission\Contracts\Permission|\Illuminate\Support\Collection  $permissions
-     * @return array
      */
-    public function collectPermissions(...$permissions)
+    private function collectPermissions(...$permissions): array
     {
         return collect($permissions)
             ->flatten()
@@ -376,7 +375,7 @@ trait HasPermissions
      */
     public function givePermissionTo(...$permissions)
     {
-        $permissions = $this->collectPermissions(...$permissions);
+        $permissions = $this->collectPermissions($permissions);
 
         $model = $this->getModel();
 
@@ -412,6 +411,8 @@ trait HasPermissions
      */
     public function syncPermissions(...$permissions)
     {
+        $this->collectPermissions($permissions);
+
         $this->permissions()->detach();
 
         return $this->givePermissionTo($permissions);
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -96,14 +96,13 @@ trait HasRoles
     }
 
     /**
-     * Assign the given role to the model.
+     * Returns roles ids as array keys
      *
-     * @param  array|string|int|\Spatie\Permission\Contracts\Role|\Illuminate\Support\Collection  ...$roles
-     * @return $this
+     * @param  array|string|int|\Spatie\Permission\Contracts\Role|\Illuminate\Support\Collection  $roles
      */
-    public function assignRole(...$roles)
+    private function collectRoles(...$roles): array
     {
-        $roles = collect($roles)
+        return collect($roles)
             ->flatten()
             ->reduce(function ($array, $role) {
                 if (empty($role)) {
@@ -122,6 +121,17 @@ trait HasRoles
 
                 return $array;
             }, []);
+    }
+
+    /**
+     * Assign the given role to the model.
+     *
+     * @param  array|string|int|\Spatie\Permission\Contracts\Role|\Illuminate\Support\Collection  ...$roles
+     * @return $this
+     */
+    public function assignRole(...$roles)
+    {
+        $roles = $this->collectRoles($roles);
 
         $model = $this->getModel();
 
@@ -175,6 +185,8 @@ trait HasRoles
      */
     public function syncRoles(...$roles)
     {
+        $this->collectRoles($roles);
+
         $this->roles()->detach();
 
         return $this->assignRole($roles);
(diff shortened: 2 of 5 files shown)


erikn69:
@drbyte ready for review

src/Traits/HasPermissions.php
```diff
@@ -412,9 +412,11 @@ function ($object) use ($permissions, $model) {
      */
     public function syncPermissions(...$permissions)
     {
+        $this->collectPermissions(...$permissions);
```
drbyte, line 415: Should this be `$permissions = $this->collectPermissions($permission);`?
- assignment of returned value
- and drop the `...` when passing the parameter (also in line 419)

Also in `HasRoles`?

erikn69: I did think of that, but `collectPermissions` returns an array of ids as keys, example
```php
[
   'id_1' => [], 'id_1' => [], 'id_1' => []
]
```
If i do `$permissions = $this->collectPermissions($permission);` i have to add `array_keys`
```php
public function syncPermissions(...$permissions)
    {
        $permissions = $this->collectPermissions(...$permissions);

        $this->permissions()->detach();

        return $this->givePermissionTo(...array_keys($permissions));
    }

```
Still, that doesn't avoid double `collectPermissions` call


@drbyte **should i make the change?**

erikn69: Done

erikn69: > value is actually a collection, then when we assign the result and then have to run array_keys against it we lose that, don't we?

Yes, in that specific case.

>So, should we revert that $permissions =  part?

I am open to change what is needed to make it work in the best way, I do?

drbyte: Yes let's revert that, and the array_keys.
My apologies for the extra work!

erikn69: Don't worry, done

src/Traits/HasRoles.php
```diff
@@ -96,14 +96,13 @@ public function scopeRole(Builder $query, $roles, $guard = null): Builder
...
     /**
-     * Assign the given role to the model.
+     * Returns roles ids as array keys
      *
-     * @param  array|string|int|\Spatie\Permission\Contracts\Role|\Illuminate\Support\Collection  ...$roles
-     * @return $this
+     * @param  array|string|int|\Spatie\Permission\Contracts\Role|\Illuminate\Support\Collection  $roles
      */
-    public function assignRole(...$roles)
+    private function collectRoles(...$roles): array
```
drbyte, line 103: in `HasPermissions` the "collect" method was `public`.
I agree with keeping it `private` whenever possible. Should the other be made private too?

erikn69: if it would be the best

src/Traits/HasPermissions.php
```diff
@@ -343,9 +343,8 @@ public function getAllPermissions(): Collection
      * Returns permissions ids as array keys
      *
      * @param  string|int|array|\Spatie\Permission\Contracts\Permission|\Illuminate\Support\Collection  $permissions
-     * @return array
      */
-    public function collectPermissions(...$permissions)
+    private function collectPermissions(...$permissions): array
```
drbyte, line 347: Note: in #2182 when this `collectPermissions()` method was created, it was created by @angeljqv as `public`, but probably should have been `private`.
It was released in `v5.5.7`.

Since it's a breaking change, albeit highly unlikely to affect anyone, we won't backport this to `v5`.

drbyte:
Thanks @erikn69 !

src/Traits/HasPermissions.php
angeljqv: Ok, i did test it, it works correctly for me, seems the right way

Issue #2332, opened 2023-02-15
hasPermissionTo never checks roles (because HasRoles trait does not work with Collection)


top-master:
https://github.com/spatie/laravel-permission/blob/0c497767227d19eff395789e65262fb95068aec3/src/Traits/HasRoles.php#L217

Above may be a bug, because Laravel's query results to collection, not an array.

I mean, Currently `hasPermissionTo` is just an alias for `hasDirectPermission`,
because `hasPermissionViaRole` always returns `false`.
And that, because `$permission->roles` returns Collection, never raw-array.

parallels999:
I think that you are wrong, if $roles is not an array it continues to collection handler
https://github.com/spatie/laravel-permission/blob/0c497767227d19eff395789e65262fb95068aec3/src/Traits/HasRoles.php#L227
Also there is many tests for hasRoles
https://github.com/spatie/laravel-permission/blob/main/tests/HasRolesTest.php

top-master:
Seems my `$this->roles` is empty,
I expected that to be loaded, because the permission is loaded based on a simple string.

Throwing something like "roles relation not-loaded" seems reasonable, instead of just returning `false`.

Basically, `false` is a lie.

parallels999:
> After debugging for few hours

 Many hours of debugging can save you 5 minutes of reading documentation

parallels999:
> It simply uses $this->roles, even if "roles" relation is never loaded.

roles as always loaded with `$this->loadMissing('roles'); `

https://github.com/spatie/laravel-permission/blob/0c497767227d19eff395789e65262fb95068aec3/src/Traits/HasRoles.php#L190-L192

>Which basically was hidding it from my review, I mean, you're lucky I don't review your code,
I would reject that instantly, because instanceof Collection check is missing.

😄😄😄😄 Just because you can't understand how things work doesn't mean they're badly done.
As you say(before you deleted), this package is so many years old, it is because it is very tested
Also, why do you delete your comments and their changes history, shame on nonsense?

top-master:
>"can't understand ... doesn't mean they're badly done"

### Human-readability is important, you're basically saying:

* readers are forced to understand,
* without any `insteanceof` check,
* and **even without any comment**.

But there are multiple other things badly done in that line, beside **not** being human-readable:
* if we pass to that method a custom class,
* which has a custom `intersect` method,
* said method is called, and maybe drops all tables, or something like that.
* it's a custom method, and it's allowed to do whatever it wants.
[...]

PR #2346, opened 2023-03-05
Fix HasRoles::hasRole to be type-safe

top-master:
I want below to be seen by members, everyone else, ignore this PR because;

I barely mentiond that I should request/report this PR's changes at #2332,
but my time was wasted with many spammers.

## Basically, #2332's related discussion comes down to:
* `HasRoles::hasRole` has doc-comment.
* We don't need to check input type, which has 1 micro second performance penality.
* Whatever happens, our doc-comment means it's your fault, check yourselves each time you call the function/method.

## My answer:
* PHP ignores your doc-comments, and is not type-safe in this case.
* Each such mixed input function/method is responsible to check it's input types.
* Don't make bad excuses, because both PHP and most IDEs silently ignore doc-comments, hence this is an issue worthy of reporting and/or PR.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -224,7 +224,11 @@ trait HasRoles
             return false;
         }
 
-        return $roles->intersect($guard ? $this->roles->where('guard_name', $guard) : $this->roles)->isNotEmpty();
+        if ($roles instanceof Collection) {
+            return $roles->intersect($guard ? $this->roles->where('guard_name', $guard) : $this->roles)->isNotEmpty();
+        }
+
+        throw new \InvalidArgumentException();
     }
 
     /**
--- a/tests/HasRolesTest.php
+++ b/tests/HasRolesTest.php
@@ -134,6 +134,14 @@ class HasRolesTest extends TestCase
     }
 
     /** @test */
+    public function it_throws_an_exception_when_determining_with_invalid_argument()
+    {
+        $this->expectException(\InvalidArgumentException::class);
+
+        $this->testUser->hasRole(new class {});
+    }
+
+    /** @test */
     public function it_throws_an_exception_when_assigning_a_role_that_does_not_exist()
     {
         $this->expectException(RoleDoesNotExist::class);


drbyte:
1. Yes you're right: we haven't been testing/checking that $roles is of type Collection. That's fixable, similar to what you propose. Thank you for your proposal.

2. How exactly did you come across this "problem"?  I mean: thousands of people are using this package but have not run reported this situation. What EXACTLY was your use-case that caused this to come to your attention?  What was your reason for passing invalid data in a way that others haven't (or didn't realize)? 
THE REASON I'm asking is: I want to understand what edge-cases would bring someone to the situation you created, so that we can consider additional tests or edge-cases.  Your candid sharing of your app-specific scenario would be helpful for optimal handling of this matter.

3. Please shorten your doc-block comments. There's no need to say "this takes only microseconds for an IDE....etc".  Let's focus on stating only what's needed for programmer readability.

top-master:
>"1. ... thanks"

No, I thank you for acceptance, many would simply give bad excuses instead.

>"2. ... use-case" and/or background

Many of my company's unit-tests were failing,
all related to same feature:
* All tests were passing in project's previous Git-commit(s).
* In new Git-commit, the only change was limiting said feature to a certain permission, but user's role already had that permission, and while debugging why `HasRoles::hasRole` would return `false`;
* We updated to latest `laravel-permission`, and added even more tests, where we noticed that `HasRoles::hasRole` is not type-safe.

>"3. ... doc-comment"

I didn't change doc-comment, those are only part of PR's description and/or conversation, **not** files changed.

But if you meant inline-comment, those do not contain what you quoted:
```php
// Simply inform users about their mistake,
// as returning `false` may be a lie
// (and because PHP and most IDEs silently ignore doc-comment).
```
Where "`PHP and most IDEs silently ignore doc-comment`" is common knowledge, and maybe I should remove that single-line to save some bytes???
**I mean** with "bytes" that, it should not make PHP any slower, specially if op-cache extension is used.

Wait, I will simply force-push it's removal (because an entire commit is like spamming).

erikn69:
> What was your reason for passing invalid data in a way that others haven't (or didn't realize)?

@drbyte i think the same, send an unexpected datatypes would be an implementation problem, not from this package
https://github.com/spatie/laravel-permission/blob/490fa00b0c8286659a534d822f7e74c684c86d27/src/Traits/HasRoles.php#L186

But if this change is necessary, all the methods should be changed, because this type of action is performed elsewhere, like `getStoredPermission` on `HasPermissions` trait, it returns collection as is, without checks.
And also if you want to check that deep, maybe we should check that each item in the collection is a role/permission contract, seems too much to me, but it would be the only way to validate those who do not want to follow documentation, 
also if you drop PHP7 you could use union type declarations on arguments and this validation would not be necessary

> I mean: thousands of people are using this package but have not run reported this situation

I imagine that everyone abide by what the documentation indicates and avoid sending unsupported data types, also Collection support is almost internal, string names are normally used
https://github.com/spatie/laravel-permission/blob/3115a43f8683f6b45e81e1bebed04955f3701c9c/src/Traits/HasPermissions.php#L296-L299

>PHP and most IDEs silently ignore doc-comment

Yes, sometimes you need extra work, i am using [laravel-ide-helper](https://github.com/barryvdh/laravel-ide-helper) for that

angeljqv:
> // Simply inform users about their mistake,
// as returning `false` may be a lie
// (and because PHP and most IDEs silently ignore doc-comment).

those comments look obvious and unprofessional, InvalidArgumentException already explains what is happening

> the only change was limiting said feature to a certain permission, but user's role already had that permission, and while debugging why HasRoles::hasRole would return false;

That goes back to describing what was already mentioned at the beginning but it doesn't explain how you are getting to that problem. 

The questions is
>What was your reason for passing invalid data in a way that others haven't (or didn't realize)? 

Or why internally invalid data is arriving? that is the real problem to find

maybe someplace needs `forgetCachedPermissions()` to reset your tests cache, maybe look at this on seeder **https://github.com/spatie/laravel-permission/issues/2342#issuecomment-1451453961**
https://github.com/spatie/laravel-permission/blob/490fa00b0c8286659a534d822f7e74c684c86d27/tests/TestCase.php#L188-L191

top-master:
@drbyte You may already know the answer for above comments, but for reference.

>"how you are getting to that **problem**"

Sorry if my previous comment was not clear enough.

The `HasRoles::hasRole` method not being type-safe is the only "**problem**" remaining.

As already mentioned, how I got to noticing this was while adding new unit-tests,
and my project's old unit-tests were already fixed when I updated to latest-version of this package.

>"all the methods should be changed"

Anyone feeling that way is free to make separate PR,
**but** others doing same mistake is never an excuse,
and is not even an argument against fixing and/or merging this case.

>"check that each item in the collection is a role"

That depends on requirements of the function/method logic .

For example, `HasRoles::hasRole` simply calls `->intersect(...)`, then Laravel handles checking entries, hence there is no such requirement in this PR.

Maybe make PR to Laravel if required.

>"I am using Laravel-IDE-helper"
>"abide by what the documentation indicates"

Me too, and yet both PHP and most IDEs silently ignore doc-comment.

erikn69:
> hence there is no such requirement in this PR

Yes, not here, that's why I put the [example](https://github.com/spatie/laravel-permission/blob/490fa00b0c8286659a534d822f7e74c684c86d27/src/Traits/HasPermissions.php#L444-L472) where it would be necessary for @drbyte 

>Anyone feeling that way is free to make separate PR

Yes, I'm asking @drbyte if he thinks it would be necessary to correct other methods after this one, I have not mentioned that this PR should cover the whole package

top-master:
@drbyte Of course, updating to PHP 8.x syntax is another option, like:
```
function myFunc(MyType|MyOtherType $myArg)
{
    // ...
}
```

But this should be merged for now, while waiting for said update to be done;
Unless said update is any time soon???

parallels999:
`master` branch is for V6 changes, no release date, and features still on discussion

src/Traits/HasRoles.php
```diff
@@ -224,7 +224,12 @@ public function hasRole($roles, string $guard = null): bool
             return false;
         }
 
-        return $roles->intersect($guard ? $this->roles->where('guard_name', $guard) : $this->roles)->isNotEmpty();
+        if ($roles instanceof Collection) {
+            return $roles->intersect($guard ? $this->roles->where('guard_name', $guard) : $this->roles)->isNotEmpty();
+        }
+
+        // Since simply returning `false` may be a lie.
```
systemsolutionweb, line 231: Remove the comment

top-master:
>"Remove the comment"

My company follows so-called "Exception free" design-pattern whenever possilbe,
Hence normally reviewers ask the opposite, like:
* Simply return a value which informs that `$myInstance->lastError()` should be checked.
* Or, explain in comment why we are forced to throw an Exception.

In this case for example, throwing does explain package users their mistake, and the comment explains future package maintainers why that's the only option.

---

@systemsolutionweb GitHub seems to **not** show the status whether you're a member or collaborator or contributor only on reviews.
I mean, as I don't know said status, I can't deside the importance of your review, and whether I should do as asked.

src/Traits/HasRoles.php
systemsolutionweb: You must follow the guidelines of the package in which you want to collaborate, 
that's the reason why you don't get a second review
`InvalidArgumentException` already explains why is an exception, because it is not an abstract `Exception`

angeljqv: >the status whether you're a member or collaborator or contributor.
I mean, as I don't know said status, I can't decide the importance of your review

that is really discriminatory and out of place, avoid that type of comment in the future
It seems to me that all collaborations can help, more eyes to review is better

top-master: @drbyte Thanks for your response, I guess, this package has nowhere anything named `lastError()`, hence throwing is always the only option, hence the comment is not really required.

I will push it's removal then.

drbyte:
Thank you for your suggestions. 

That said, we're not merging this at this time.

PR #2364, opened 2023-03-20
Fix nesting tests

erikn69:
In #2280 a new Migration, and Custom Model were added only for one test, also a complex process for changing migration was added, this PR deletes those files and use existing ones, also i did try to make test more simple

Also in #2280 the consistency between the Permissions and Roles models was lost, i did delete `function getTable()` on permission to match roles model


drbyte:
Thanks!

drbyte:
Merged to v6-dev.
Should we backport this to 5.x as well?

erikn69:
> Should we backport this to 5.x as well

I don't think it's necessary, it doesn't fix any bugs, does not change any previous operation

PR #2376, opened 2023-03-27
[V6] Add PHPStan workflow with fixes

erikn69:
**Larastan added to the workflow** [https://github.com/erikn69/laravel-permission/actions/runs/4569204007/jobs/8065175411)


drbyte:
Thanks for this. 

Do you mind splitting this up into separate PRs for each of the issues being fixed? 
Some of these are breaking changes and it's a little harder for people to update their own apps when they can see a canonical PR related to the actual change that's affecting their own code.

PR #2377, opened 2023-03-28
Only offer publishing when running in console

erikn69:
https://github.com/spatie/laravel-package-tools/blob/d7ea4f74ffe1f7863f01bf5ba4b6d64847dacf8b/src/PackageServiceProvider.php#L58-L69

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -56,6 +56,10 @@ class PermissionServiceProvider extends ServiceProvider
             return;
         }
 
+        if (! $this->app->runningInConsole()) {
+            return;
+        }
+
         $this->publishes([
             __DIR__.'/../config/permission.php' => config_path('permission.php'),
         ], 'permission-config');


src/PermissionServiceProvider.php
```diff
@@ -67,13 +71,17 @@ protected function offerPublishing()
 
     protected function registerCommands()
     {
-        $this->commands([
+        if (! $this->app->runningInConsole()) {
+            return;
+        }
```
drbyte, line 76: Will this block the ability of an application (such as via the app's Admin interface) from calling any of these Artisan commands? 

I recall reverting a similar PR quite awhile back for something akin to this sort of change. Not sure if it was exactly this though.

erikn69: > Will this block the ability of an application from calling any of these Artisan commands?

yes
> Not sure if it was exactly this though

https://github.com/spatie/laravel-permission/commit/6b1aedd0f25dfbca19c174b9f03bf2e69d9eeb28, it was because they want to use cache reset from the app in Laravel Nova, but maybe just leave that command, I see it useful in a interface app to create roles, permissions by command

reverted

PR #2384, opened 2023-03-30
Use base_path helper for lumen

erikn69:
This lets people use this package in Lumen if they use it with larasupport or another package for publishing

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -51,17 +51,12 @@ class PermissionServiceProvider extends ServiceProvider
 
     protected function offerPublishing()
     {
-        if (! function_exists('config_path')) {
-            // function not available and 'publish' not relevant in Lumen
-            return;
-        }
-
         if (! $this->app->runningInConsole()) {
             return;
         }
 
         $this->publishes([
-            __DIR__.'/../config/permission.php' => config_path('permission.php'),
+            __DIR__.'/../config/permission.php' => base_path('config/permission.php'),
         ], 'permission-config');
 
         $this->publishes([


src/PermissionServiceProvider.php
```diff
@@ -51,17 +51,12 @@ public function register()
...
-            return;
-        }
-
         if (! $this->app->runningInConsole()) {
             return;
         }
 
         $this->publishes([
-            __DIR__.'/../config/permission.php' => config_path('permission.php'),
+            __DIR__.'/../config/permission.php' => base_path('config/permission.php'),
```
drbyte, line 59: Does the `$this->publishes()` call trigger errors in Lumen?

ie: Was the Lumen check testing for `config_path()` because that's what was triggering errors? Or was that the test that was implemented to be clear whether Lumen was being used?

erikn69: > Does the $this->publishes() call trigger errors in Lumen?

Not, beacuse lumen requires [`illuminate/support`](https://github.com/laravel/lumen-framework/blob/7f1672b8a6d7f14c72029d56d21672c815723c10/composer.json#L39) for service provider, and it has [`publishes(array $paths, $groups = null)`](https://github.com/illuminate/support/blob/646bfb9e8485e5f82921512a19643f26bbb778d3/ServiceProvider.php#L277-L286)

>Was the Lumen check testing for config_path() because that's what was triggering errors?

Yes, 

>Or was that the test that was implemented to be clear whether Lumen was being used?

On lumen you can manually add a `config_path` helper, and pass that validation

but with this change, it is possible to use it in lumen without adding any helper or other package, and in case of using a package like [lumen-larasupport](https://github.com/irazasyed/larasupport), the `vendor:publish` command will be available

erikn69: https://github.com/irazasyed/larasupport/blob/c9b7f4f7b6ca708dc400250039dc3fee4b6ff15f/src/helpers.php#L42-L45

drbyte:
In case this comes up again, do you mind explaining why you closed this?

erikn69:
I might be over-thinking here, if someone manually adds a `config_path`, there is no errors on this
and if someone adds a third party package like **larasupport**, it brings `config_path` and other helpers, so no problem, it publishes correctly

PR #2388, opened 2023-04-05
[V6] Drop PHP 7.3 support

angeljqv:
It's about time
image

### DIFF
--- a/.github/workflows/run-tests-L8.yml
+++ b/.github/workflows/run-tests-L8.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


drbyte:
What problem does this fix?

angeljqv:
Leaving obsolete versions behind gives us the opportunity to use the features/syntax of the newer versions.

drbyte:
Yes, I understand that PHP 7.3 is now considered ancient.

However, since we're still supporting Laravel 8, which is compatible back to PHP 7.3, it felt sensible to retain it for now.

Which "new features/syntax" are you eager to have this package use?

PR #2391, opened 2023-04-11
[v6] Add BackedEnum support

drbyte:
Fixes #1991

Note: Requires PHP 8.1+

/cc @ceejayoz
/cc @edalzell

### DIFF
--- a/docs/basic-usage/enums.md
+++ b/docs/basic-usage/enums.md
@@ -1,0 +1,103 @@
+---
+title: Enums
+weight: 3
+---
+
+# Enum Prerequisites
+
+Must be using PHP 8.1 or higher.
+
+If you are using PHP 8.1+ you can implement Enums as native types. 
+
+Internally, Enums implicitly implement `\BackedEnum`, which is how this package recognizes that you're passing an Enum.
+
+
+# Code Requirements
+
+You can create your Enum object for use with Roles and/or Permissions. You will probably create separate Enums for Roles and for Permissions, although if your application needs are simple you might choose a single Enum for both.
+
+Usually the list of application Roles is much shorter than the list of Permissions, so having separate objects for them can make them easier to manage.
+
+Here is an example Enum for Roles. You would do similarly for Permissions.
+
+```php
+namespace App\Enums;
+
+enum RolesEnum: string
+{
+    // case NAMEINAPP = 'name-in-database';
+
+    case WRITER = 'writer';
+    case EDITOR = 'editor';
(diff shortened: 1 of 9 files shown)


drbyte:
Feedback welcome. I know there are lots of varying ways people like to implement enums. This covers the basics, but if there are some clear use-cases that aren't supported here, this is a good time to make adjustments.

boumanb:
@drbyte to me this looks like a perfect PR example. It's specific and well-documented. 

> Why is the interface needed?

If I may answer; to assure it's a `BackedEnum`. To quote [from the PHP documentation](https://www.php.net/manual/en/class.backedenum.php#backedenum.intro):

> The BackedEnum interface is automatically applied to backed enumerations by the engine. It may not be implemented by user-defined classes. Enumerations may not override its methods, as default implementations are provided by the engine. It is available only for type checks.

The approach of @drbyte follows the example of the comment in the [BackedEnum PHP documentation](https://www.php.net/manual/en/class.backedenum.php#126826). Although the `BackedEnum` interface may not be implemented it CAN BE extended.

edalzell:
> > Why is the interface needed?
> 
> If I may answer; to assure it's a `BackedEnum`. To quote [from the PHP documentation](https://www.php.net/manual/en/class.backedenum.php#backedenum.intro):
> 
> > The BackedEnum interface is automatically applied to backed enumerations by the engine. It may not be implemented by user-defined classes. Enumerations may not override its methods, as default implementations are provided by the engine. It is available only for type checks.
> 
> The approach of @drbyte follows the example of the comment in the [BackedEnum PHP documentation](https://www.php.net/manual/en/class.backedenum.php#126826). Although the `BackedEnum` interface may not be implemented it CAN BE extended.

Sure but you can do this to get a BackedEnum:

CleanShot 2023-04-11 at 11 11 51@2x

...and  that interface is never checked anywhere, only `BackedEnum`. So it seems like it's not used anywhere?

Sorry for my confusion if I'm way off track here, trying to learn!

edalzell:
To be super clear, I'm talking about this:

CleanShot 2023-04-11 at 11 13 41@2x

drbyte:
> I'll update the code in a bit.

Updated. 
Also revised the docs a bit as well.

Question: For those of you who have been using Enums already, how have you handled `can()` and `@can()` situations? And route middleware parameters?

edalzell:
This is very awesome @drbyte and @erikn69, thanks for doing this!

drbyte:
@erikn69 I'd support a PR for that addition.

PR #2397, opened 2023-04-12
Phpstan fixes

erikn69:
Make PHPStan happy
On [phpstan:v1.10.12](https://github.com/phpstan/phpstan/releases/tag/1.10.12), there is new validations

https://github.com/spatie/laravel-permission/actions/runs/4681036769/jobs/8293124388

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -25,7 +25,7 @@ class PermissionRegistrar
     /** @var string */
     protected $roleClass;
 
-    /** @var Collection|null */
+    /** @var Collection|array|null */
     protected $permissions;
 
     /** @var string */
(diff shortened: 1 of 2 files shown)


src/PermissionRegistrar.php
```diff
@@ -25,7 +25,7 @@ class PermissionRegistrar
     /** @var string */
     protected $roleClass;
 
-    /** @var Collection|null */
+    /** @var array|null */
```
drbyte, line 28: Um ... but it's actually still a collection internally ...

erikn69: You are right, I see that the problem starts with the new version of phpstan

erikn69: Fixed

szepeviktor:
> there is new validations

Which validation is it?

Without the var tag PHPStan does not see the contract.

```
Method Spatie\Permission\Models\Permission::getPermission() should return Spatie\Permission\Contracts\Permission|null but returns Illuminate\Database\Eloquent\Model|null.
```

PR #2410, opened 2023-04-26
Improve Performance of Loading Relationships

jonathan-bird:
If a model is already loaded, there shouldn't be any forcing of reloading a relationship through the `load()` method. Using Laravel's `loadMissing()` method, it will only load if it hasn't already been eager-loaded.

Won't interrupt anyone who doesn't eager load this already, and will improve performance of people who eager load permissions & roles on their model.

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -391,7 +391,7 @@ trait HasPermissions
 
         if ($model->exists) {
             $this->permissions()->sync($permissions, false);
-            $model->load('permissions');
+            $model->loadMissing('permissions');
         } else {
             $class = \get_class($model);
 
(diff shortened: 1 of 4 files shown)


jonathan-bird:
Open to any thoughts or criticism.

parallels999:
Did you run tests??
It seems to me that "load" is used to force reloading after changing some pivot relationship

drbyte:
@jonathan-bird wrote:
> If a model is already loaded, there shouldn't be any forcing of reloading a relationship through the `load()` method. Using Laravel's `loadMissing()` method, it will only load if it hasn't already been eager-loaded.

Thanks for your proposal.
Is there a situation in your app that caused you to run into a problem that led you to submit this PR?

While exploring the failing tests, and the proposed alternate solution, I want to be sure to understand **your** situation, whatever caused you to offer this proposal.

drbyte:
Closed by #2412

drbyte:
@jonathan-bird if there's more about this that you think needs addressing, please let us know!

PR #2412, opened 2023-04-26
Remove force loading model relationships

erikn69:
Alternative to #2410

In some scenarios it is not necessary to load the relationships, and they are loaded later when they are required, this would allow less sqls to be executed in the mentioned scenarios
**Example:** batch user creation and assignment

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -391,7 +391,7 @@ trait HasPermissions
 
         if ($model->exists) {
             $this->permissions()->sync($permissions, false);
-            $model->load('permissions');
+            $model->unsetRelation('permissions');
         } else {
             $class = \get_class($model);
 
@@ -401,7 +401,7 @@ trait HasPermissions
                         return;
                     }
                     $model->permissions()->sync($permissions, false);
-                    $model->load('permissions');
+                    $model->unsetRelation('permissions');
                 }
             );
         }
@@ -442,7 +442,7 @@ trait HasPermissions
             $this->forgetCachedPermissions();
         }
 
-        $this->load('permissions');
+        $this->unsetRelation('permissions');
 
         return $this;
     }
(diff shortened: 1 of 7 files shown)


erikn69:
Actually it is not a bug, it is a performance improvement, it is not necessary to load the relation immediately, it is enough to remove it, if there is any check after that the relation will be reloaded, otherwise this sql and memory usage are avoided

Look at the tests, the sql count is less on
`calling_givePermissionTo_before_saving_object_doesnt_interfere_with_other_objects`
`calling_syncPermissions_before_saving_object_doesnt_interfere_with_other_objects`
`calling_syncRoles_before_saving_object_doesnt_interfere_with_other_objects`
`calling_assignRole_before_saving_object_doesnt_interfere_with_other_objects`
This would decrease sql batch tasks too

erikn69:
>  what the issues were when we specifically added the load() calls into the package

Ok, after sync, detach, the relations may be preloaded, if so, the new changes are not accessible, so use `load` to get the updated relation data from db

erikn69:
@drbyte here https://github.com/spatie/laravel-permission/commit/efab130b4a9ed6cebb461f2a92a9dd9fc5a93f6a

drbyte:
> > what the issues were when we specifically added the load() calls into the package
> 
> Ok, after sync, detach, the relations may be preloaded, if so, the new changes are not accessible, so use `load` to get the updated relation data from db

Right. So it seems our existing `loadMissing()` calls take care of the things needed after `unsetRelation` has been called.

drbyte:
```diff
+         $this->testUser->loadMissing('roles.permissions', 'permissions'); // load relations
```

I'm concerned that requiring this in the Cache tests may indicate ripple-effect problems.

erikn69:
> ```diff
> +         $this->testUser->loadMissing('roles.permissions', 'permissions'); // load relations
> ```
> 
> I'm concerned that requiring this in the Cache tests may indicate ripple-effect problems.

Those tests are supposed to check that no unnecessary calls are made to the db after the relations have been loaded.
Do I go back to how it was before?

drbyte:
> Do I go back to how it was before?

No. I'm just letting you know my thoughts as I look at the PR. 
I'm always trying to identify any hidden BC breaks.

I think this should be fine.

PR #2416, opened 2023-04-26
Test alternate cache drivers

### LINKED ISSUE #2413: Add tests for alternate Cache Drivers
Currently the test suite only tests the default in-memory `array` cache.

I sometimes manually run the test suite against alternate Cache Stores:
https://github.com/spatie/laravel-permission/blob/ec9079aadfe20051a9ccf0a15b6c9ed453b7bca6/tests/TestCase.php#L115-L118

It would be ideal to automate this.

erikn69:
Closes #2413
It seems to work correctly, but I don't know if it is the most appropriate option

### DIFF
--- a/.github/workflows/test-cache-drivers.yml
+++ b/.github/workflows/test-cache-drivers.yml
@@ -1,0 +1,67 @@
+name: "Run Tests - Cache Drivers"
+
+on: [push, pull_request]
+
+jobs:
+  cache:
+
+    runs-on: ubuntu-latest
+
+    services:
+      redis:
+        image: redis
+        ports:
+          - 6379/tcp
+        options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+    strategy:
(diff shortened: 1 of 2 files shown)


drbyte:
I see your point. Ya the temporary file is one way. But that makes me contemplate just using the yml to drive it all?

- set the env var for cache driver (to override phpunit.xml)
- load any needed extensions (eg: to also test memcached, redis) into php

If they all take too long in one batch, we could push the memcached and redis to separate test yml's.

erikn69:
Done

drbyte:
Looks great!

Question: do we know for certain that the expected driver is being used in each variation? 👀

drbyte:
> Question: do we know for certain that the expected driver is being used in each variation? 👀

I guess it is. (https://github.com/drbyte/laravel-permission/actions/runs/4814920227/jobs/8573107658)
Just wanted to be sure it was indeed reading the ENV settings correctly.

PR #2417, opened 2023-04-26
Avoid bringing laravel in tests

erikn69:
it is faster


drbyte:
Yes, it is a little faster.
And consistent with `composer.json`

PR #2423, opened 2023-05-02
[V6] fewer sqls in syncRoles, syncPermissions

erikn69:
This avoid `detach` when model does'nt exists, and it set an empty collection on relation for prevent reload relation from sql
this results in fewer sql calls on sync


erikn69:
@Kolyunya There are tests that use array of IDs
https://github.com/spatie/laravel-permission/blob/da5c8bc931692fa0a1a0fbda1a5995c9379e4159/tests/HasRolesWithCustomModelsTest.php#L98

erikn69:
Maybe, are you using array of strings? 

- https://github.com/spatie/laravel-permission/issues/2530#issuecomment-1797924355

PR #2434, opened 2023-05-22
Add description to table roles and permissions

putera:
Add description to the roles and permissions table, so that we know what the roles and permission is about in detail.


parallels999:
you must do that in your own migration file, not all of us need a description

putera:
Ok then.

PR #2437, opened 2023-05-26
[V6] Update PHPDocs for IDE autocompletion

erikn69:
Closes #2436
**Fix:** IDE doesn't autocomplete on Role/Permission when `create` method is called


drbyte:
Thanks!

PR #2438, opened 2023-05-29
Update uuid.md

ChillMouse:
Resolve problems with Laravel 10. Ex version of uuid.md doesn`t correct working. In this manual -- all good

### DIFF
--- a/docs/advanced-usage/uuid.md
+++ b/docs/advanced-usage/uuid.md
@@ -3,65 +3,61 @@ title: UUID
 weight: 7
 ---
 
-If you're using UUIDs or GUIDs for your User models there are a few considerations to note.
+If you're using UUIDs for your User models there are a few considerations to note.
 
 > THIS IS NOT A FULL LESSON ON HOW TO IMPLEMENT UUIDs IN YOUR APP.
 
 Since each UUID implementation approach is different, some of these may or may not benefit you. As always, your implementation may vary.
 
-
 ## Migrations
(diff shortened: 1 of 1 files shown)


docs/advanced-usage/uuid.md
```diff
@@ -86,6 +104,83 @@ If you want all the role/permission objects to have a UUID instead of an integer
...
+namespace App\Models;
+
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Spatie\Permission\Models\Permission as SpatiePermission;
+
+class Permission extends SpatiePermission
+{
+    use HasFactory, HasUuids;
+    protected $primaryKey = 'id';
```
erikn69, line 147: Also primary key `id` field on Role/Permission can be changed, 
Just use `protected $primaryKey = 'uuid';` on models, and update migration file, it is tested
https://github.com/spatie/laravel-permission/blob/9d7ea6da1d4ed1f41970823a3060aa459f8cae3a/tests/TestCase.php#L176-L179
https://github.com/spatie/laravel-permission/blob/9d7ea6da1d4ed1f41970823a3060aa459f8cae3a/tests/TestCase.php#L185-L188

drbyte:
Thanks

drbyte:
Note that I removed some of the properties in the example models because the HasUuids trait already handles them.

Issue #1037, opened 2019-02-20
Question: get users has not role X


kossa:
Hello,

How to get the list of users that has not role "x" and "y" ?

Thank you

drbyte:
The package currently offers a `roles` scope to let you find users which DO have the specified roles:
https://github.com/spatie/laravel-permission/blob/087a53cf8ec0cf79ad326f612b0d171364f0d9b6/src/Traits/HasRoles.php#L51-L88

You could rework your logic to work with that (either to determine a "with" instead of "without" logic flow, or simply grab a collection of all the "other" roles and pass that to the scope), or you could add your own scope to your model to perform the reverse "not" lookup.

kossa:
@drbyte alright thank you, I need this for feature testing.

biholaindrasinh:
Add this to your user model:
```php
    public function scopeNotRole(Builder $query, $roles, $guard = null): Builder 
    { 
         if ($roles instanceof Collection) { 
             $roles = $roles->all(); 
         } 
  
         if (! is_array($roles)) { 
             $roles = [$roles]; 
         } 
  
         $roles = array_map(function ($role) use ($guard) { 
             if ($role instanceof Role) { 
                 return $role; 
             } 
  
             $method = is_numeric($role) ? 'findById' : 'findByName'; 
             $guard = $guard ?: $this->getDefaultGuardName(); 
  
             return $this->getRoleClass()->{$method}($role, $guard); 
         }, $roles); 
  
         return $query->whereHas('roles', function ($query) use ($roles) { 
             $query->where(function ($query) use ($roles) { 
                 foreach ($roles as $role) { 
                     $query->where(config('permission.table_names.roles').'.id', '!=' , $role->id); 
                 } 
             }); 
         }); 
    }
```
You can use it like this:
```php
$users = User::notRole('Admin')->get();
$users = User::notRole(['Admin', 'Super Admin'])->get();
```
This might help you. Let me know if it's not working.

viraljetani:
@biholaindrasinh This should be included in the original permissions package.. Much needed function.

limonazzo:
@biholaindrasinh thank you

lovecoding-git:
@biholaindrasinh  not working.

drbyte:
You can use basic built-in Laravel Eloquent functions such as `doesntHave()` as documented here:
https://docs.spatie.be/laravel-permission/v3/basic-usage/basic-usage/
and in the Laravel documentation.

[...]

PR #2463, opened 2023-07-13
Add withoutRole and withoutPermission scopes

drbyte:
This allows the inverse of the prior scopes to also allow finding users that do-not-have the specified role or permission

Ref: #1037


drbyte:
Credits to @biholaindrasinh in #1037

PR #2464, opened 2023-07-14
Gate check blocked permission

mahdimsr:
check blocked permission in the hasDirect permission method


drbyte:
Hi. Thanks for the submission.

This package has always focused on "granting" permissions, not "blocking" or "taking away" permissions.

You could extend the methods into your own code to add your customized logic, and use a custom middleware to enforce it.

mahdimsr:
Hi.
Ok. I will 👍

PR #2467, opened 2023-07-19
Adding the ability to use package with service-to-service Passport client

SuperDJ:


src/Middlewares/RoleMiddleware.php
```diff
@@ -12,6 +13,23 @@ public function handle($request, Closure $next, $role, $guard = null)
...
+        $bearerToken = $request->bearerToken();
+        if ($bearerToken) {
+            $client = ClientService::getClient($bearerToken);
+            $assignedRoles = ClientService::getClientRoles($client);
+
+            foreach ($roles as $role) {
+                if (in_array($role, $assignedRoles)) {
+                    return $next($request);
+                }
+            }
```
parallels999, line 30: Will it work the same with this change?
```diff
-$assignedRoles = ClientService::getClientRoles($client);
-foreach ($roles as $role) {
-    if (in_array($role, $assignedRoles)) {
-        return $next($request);
-    }
-}
+if ($client->hasAnyRole($roles)) {
+    return $next($request);
+}
```

parallels999:
Question 

https://github.com/laravel/passport/blob/b566e159c9661c49ca07b8e12a91ef86ee1bb94c/src/Client.php#L77-L84
```php
public function user()
{
    $provider = $this->provider ?: config('auth.guards.api.provider');

    return $this->belongsTo(
        config("auth.providers.{$provider}.model")
    );
}
```

So, `config("auth.providers.{$provider}.model")` must be the model which has the `hasRolesTrait`
```php
$client = ClientService::getClient($bearerToken);
$user = $client->user
```

tests/TestModels/Client.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace Spatie\Permission\Tests\TestModels;
+
+use Laravel\Passport\Client as BaseClient;
+use Spatie\Permission\Traits\HasRoles;
+
+class Client extends BaseClient
+{
+    use HasRoles;
```
parallels999, line 10: maybe
```php
use Illuminate\Contracts\Auth\Access\Authorizable as AuthorizableContract;
use Illuminate\Foundation\Auth\Access\Authorizable;

class Client extends BaseClient implements AuthorizableContract
{
    use Authorizable;
    use HasRoles;
```
For `$client->can($permissions)`

SuperDJ: Just updated the code with the suggested changes. But now getting either `403` when `200` is expected or `GuardDoesNotMatch` exception

src/Middlewares/PermissionMiddleware.php
```diff
@@ -12,6 +13,20 @@ public function handle($request, Closure $next, $permission, $guard = null)
...
         $authGuard = Auth::guard($guard);
 
+        $permissions = is_array($permission)
+            ? $permission
+            : explode('|', $permission);
+
+        // For machine-to-machine Passport clients
+        $bearerToken = $request->bearerToken();
+        if ($bearerToken) {
+            $client = ClientService::getClient($bearerToken);
```
parallels999, line 23: What about `client` build in method? [passport/Guards/TokenGuard.php#L149](https://github.com/laravel/passport/blob/b566e159c9661c49ca07b8e12a91ef86ee1bb94c/src/Guards/TokenGuard.php#L149)

I had something like this in mind:
```php
public function handle($request, Closure $next, $permission, $guard = null)
{
    $authGuard = Auth::guard($guard);

    $bearerToken = $request->bearerToken();
    if ($bearerToken && method_exists($authGuard, 'client')) {
        $user = $authGuard->client();
    }

    $user = $user ?? $authGuard->user();

    if (! $user) {
        throw UnauthorizedException::notLoggedIn();
    }

    if (! method_exists($user, 'hasAnyPermission')) {
        throw UnauthorizedException::missingTraitHasRoles($user);
    }

    $permissions = is_array($permission)
        ? $permission
        : explode('|', $permission);

    if (! $user->canAny($permissions)) {
        throw UnauthorizedException::forPermissions($permissions);
    }

    return $next($request);
}
```

src/Middlewares/RoleOrPermissionMiddleware.php
```diff
@@ -4,13 +4,31 @@
...
         $authGuard = Auth::guard($guard);
+
+        $rolesOrPermissions = is_array($roleOrPermission)
+            ? $roleOrPermission
+            : explode('|', $roleOrPermission);
+
+        // For machine-to-machine Passport clients
+        $bearerToken = $request->bearerToken();
+        if ($bearerToken) {
+            $client = ClientService::getClient($bearerToken);
```
parallels999, line 23: https://github.com/spatie/laravel-permission/pull/2467#discussion_r1269482204 can be applied to all middlewares, so the `ClientService` class can be removed
```php
public function handle($request, Closure $next, $roleOrPermission, $guard = null)
{
    $authGuard = Auth::guard($guard);

    $bearerToken = $request->bearerToken();
    if ($bearerToken && method_exists($authGuard, 'client')) {
        $user = $authGuard->client();
    }

    $user = $user ?? $authGuard->user();

    if (! $user) {
        throw UnauthorizedException::notLoggedIn();
    }

    if (! method_exists($user, 'hasAnyRole') || ! method_exists($user, 'hasAnyPermission')) {
        throw UnauthorizedException::missingTraitHasRoles($user);
    }

    $rolesOrPermissions = is_array($roleOrPermission)
        ? $roleOrPermission
        : explode('|', $roleOrPermission);

    if (! $user->canAny($rolesOrPermissions) && ! $user->hasAnyRole($rolesOrPermissions)) {
        throw UnauthorizedException::forRolesOrPermissions($rolesOrPermissions);
    }

    return $next($request);
}
```

src/Middlewares/RoleMiddleware.php
parallels999:
```php
public function handle($request, Closure $next, $role, $guard = null)
{
    $authGuard = Auth::guard($guard);

    $bearerToken = $request->bearerToken();
    if ($bearerToken && method_exists($authGuard, 'client')) {
        $user = $authGuard->client();
    }

    $user = $user ?? $authGuard->user();

    if (! $user) {
        throw UnauthorizedException::notLoggedIn();
    }

    if (! method_exists($user, 'hasAnyRole')) {
        throw UnauthorizedException::missingTraitHasRoles($user);
    }

    $roles = is_array($role)
        ? $role
        : explode('|', $role);

    if (! $user->hasAnyRole($roles)) {
        throw UnauthorizedException::forRoles($roles);
    }

    return $next($request);
}
```

tests/TestModels/Client.php
SuperDJ: Unfortunate the `getDefaultGuardName` method doesn't appear to be working. Adding a guard name variable does appear to be working

tests/RoleOrPermissionMiddlewareTest.php
```diff
@@ -101,12 +136,28 @@ public function a_user_can_not_access_a_route_protected_by_permission_or_role_mi
 
         $this->assertEquals(
             403,
-            $this->runMiddleware($this->roleOrPermissionMiddleware, 'testRole|edit-articles')
+            $this->runMiddleware($this->roleOrPermissionMiddleware, 'testRole|edit-articles', 'api', true)
```
drbyte, line 139: Must this always be `'api'`?

tests/TestModels/Client.php
```diff
@@ -0,0 +1,21 @@
...
+    use Authorizable;
+
+    /**
+     * Required to make clear that the client requires the api guard
+     *
+     * @var string
+     */
+    protected $guard_name = 'api';
```
drbyte, line 20: FYI Another option could be the following. This package inspects for both the method or the property.
```php
    public function guardName()
    {
        return 'api'; // or perhaps a config() value;
    }
```

src/Middlewares/RoleOrPermissionMiddleware.php
```diff
@@ -11,20 +11,27 @@ class RoleOrPermissionMiddleware
...
         }
 
         $rolesOrPermissions = is_array($roleOrPermission)
             ? $roleOrPermission
             : explode('|', $roleOrPermission);
 
+        if (! method_exists($user, 'hasAnyRole') || ! method_exists($user, 'hasAnyPermission')) {
+            throw UnauthorizedException::missingTraitHasRoles($user);
+        }
+
```
drbyte, line 34: While this is a minor nitpick, moving this up a few lines, before the `$roleOrPermissions = is_array()...` block saves a CPU cycle if this exception needs to be thrown.

tests/TestCase.php
```diff
@@ -158,14 +170,21 @@ protected function setUpDatabase($app)
...
         $this->testUserRole = $app[Role::class]->create(['name' => 'testRole']);
-        $app[Role::class]->create(['name' => 'testRole2']);
-        $this->testAdminRole = $app[Role::class]->create(['name' => 'testAdminRole', 'guard_name' => 'admin']);
         $this->testUserPermission = $app[Permission::class]->create(['name' => 'edit-articles']);
+
+        $this->testAdmin = Admin::create(['email' => 'admin@user.com']);
+        $this->testAdminRole = $app[Role::class]->create(['name' => 'testAdminRole', 'guard_name' => 'admin']);
+        $this->testAdminPermission = $app[Permission::class]->create(['name' => 'admin-permission', 'guard_name' => 'admin']);
+
+        $this->testClient = Client::create(['name' => 'Test', 'redirect' => 'https://google.com', 'personal_access_client' => 0, 'password_client' => 0, 'revoked' => 0]);
```
drbyte, line 180: This is the tiniest of suggested changes, but let's use example.com instead of google, to be clear that nothing here is dependent on any google services.

drbyte:
Thanks for all your work on this!

tests/TestModels/Client.php
SuperDJ: It could be a config value. But I don't know if that would be required since this Model is only used for testing. And I guess, in the docs we will describe what is required to use the Passport Client. And so the guard name is up for the end user the decide? I don't think that this package should include the extended Client class (besides the one for testing)?

SuperDJ:
@parallels999 & @drbyte perhaps one of you can help with this. I have been testing this locally but I run into a problem where the (permission)middleware always returns 403. What I noticed is that `Auth::guard($guard)` returns `Illuminate\Auth\SessionGuard` instead of `Laravel\Passport\Guards\TokenGuard`. Could this be because the route is in the `client` middleware as well?

For example:
```php
Route::group(['middleware' => 'client'], function() {
   Route::get(...)->middleware('permission:edit-posts');
});
```
Almost as if the `client` middleware changes it.

drbyte:
Regarding the 403, I'll have to wait til tomorrow to spend some time looking in detail.

Regarding the currently-failing tests, it's because you changed (rearranged) the order in which the test-case roles/permissions were created, so the id numbers are out of sync with the tests. We can make those tests less rigid by querying the ID numbers before testing against them, but don't need to do that inside this PR, because it's unrelated. You could put them back in original order and we can clean them up in another PR.

erikn69:
I'll take a look

parallels999:
> I found that when using Passport v11 and providing Auth::guard('api) did return the expected TokenGuard

Did you set `Route::middleware(['auth:api'])` on your routes?

SuperDJ:
@parallels999 for the [client credentials grant](https://laravel.com/docs/10.x/passport#protecting-routes) it shouldn't be used:

>If you are using the [client credentials grant](https://laravel.com/docs/10.x/passport#client-credentials-grant-tokens), you should use [the client middleware](https://laravel.com/docs/10.x/passport#client-credentials-grant-tokens) to protect your routes instead of the auth:api middleware.

parallels999:
Ok, I understand the problem, in that case it would be better to read the token to get the client, but is there a way to do it by reusing passport?

SuperDJ:
@parallels999 I haven't been able to find a specific function to do so. I only found several issues trying to do the same such as this one https://github.com/laravel/passport/issues/124#issuecomment-1491170276.

angeljqv:
> For example:
>```
>Route::group(['middleware' => 'client'], function() {
>   Route::get(...)->middleware('permission:edit-posts');
>});
>```

What about scopes? [client-credentials-grant-tokens](https://laravel.com/docs/master/passport#client-credentials-grant-tokens)
```php
Route::get('/orders', function (Request $request) {
    ...
})->middleware('client:role-name-1,role-name-2');
```

angeljqv:
>We don't change the working the client middleware.

I have never said that you should change the behavior of the `client` middleware
I say that scopes already exists, when the token is generated, read documentation again

SuperDJ:
@angeljqv I prefer consistency. And I'm already using laravel-permission package for permissions on the web part of my project. So to me it only seems logical to also apply the permissions to the API part. That way I only have to deal with one "system" for permissions.

angeljqv:
ok, i got it

drbyte:
@SuperDJ This is looking good. Do you have any remaining concerns about it?

drbyte:
Also, are you comfortable squashing the commits into one and then force-pushing that commit back to this branch?

erikn69:
@drbyte you can use squash merge and change the commit message

drbyte:
> @drbyte you can use squash merge PR and change the commit message

Agreed.

SuperDJ:
> Do you have any remaining concerns about it?

@drbyte the only thing I'm not sure of, is what I mentioned [here](https://github.com/SuperDJ/laravel-permission/pull/1#issuecomment-1649247186). Are the checks specific enough to only work for the Client Credentials Grant and not the other Passport Grants. Unfortunatly I'm currently unable to check this.

drbyte:
Unfortunately it's not an area where I have much experience. Does your own app need a solution to that? Or can we safely document it as a limitation for which help is welcomed?

SuperDJ:
@drbyte currently my app only needs to use the Client Credentials Grant. And that is working with the changes made in this PR. But we could add something to the docs indeed. 

Or maybe add another check after line 95 of Guard.php perhaps on line 100 of `if(!$client->fistParty()) {return null;}`. That should atleast prevent 2 other Grants from being affected. Perhaps @erikn69 can say something about it?

erikn69:
> Password Grant should just use the User model

if the token uses User model, `$authGuard->user()` gets the model and `getPassportClient`  is never gonna be called
I think it should be enough as it is, but I can't check it, I don't use passport
```php
$user = $authGuard->user();
// For machine-to-machine Passport clients
if (! $user && $request->bearerToken()) {
    $user = Guard::getPassportClient($guard);
}
```

SuperDJ:
Isn't there some sort of beta release so things like this can be tested?

drbyte:
> Isn't there some sort of beta release so things like this can be tested?

We haven't usually released "betas" on this package. 
But you **can** even now simply `composer require spatie/laravel-permission:^6.0-dev` to use it prior to official release. Anything merged into the package's master branch will be incorporated.

SuperDJ:
@drbyte made it an opt-in feature. That way is up to the end user to decide if they want to use it.

parallels999:
>Mostly user is set and not bearer token
>```php
>if (! $user && $request->bearerToken() && config('permission.use_passport_client_credentials'))
>```

tests/TestCase.php
```diff
@@ -114,6 +114,7 @@ protected function getEnvironmentSetUp($app)
         $app['config']->set('permission.column_names.role_pivot_key', 'role_test_id');
         $app['config']->set('permission.column_names.permission_pivot_key', 'permission_test_id');
         $app['config']->set('view.paths', [__DIR__.'/resources/views']);
+        $app['config']->set('permission.use_passport_client_credentials', true);
```
drbyte, line 117: Should this be moved to the `setUpPassport()` function?

drbyte:
@SuperDJ thanks for all your work on this.

I'm going to merge it so it can be used in dev as beta etc. And if it is found that adjustments are required, those can be done via PR.

PR #2480, opened 2023-08-19
Update contracts to allow for UUID

drbyte:
Also updates to declare `?string` as `$guardName` signature

/cc @erikn69

### DIFF
--- a/src/Contracts/Permission.php
+++ b/src/Contracts/Permission.php
@@ -21,7 +21,6 @@ interface Permission
     /**
      * Find a permission by its name.
      *
-     * @param string|null $guardName
      *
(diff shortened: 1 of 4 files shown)


erikn69:
Looks great to me,
it's a breaking change, but it's an major version, so 👍

parallels999:
If you drop Php7.4 you could use [Union Types](https://php.watch/versions/8.0/union-types) on findbyId

drbyte:
> If you drop Php7.4 you could use [Union Types](https://php.watch/versions/8.0/union-types) on findbyId

Ya, I was itching for that too.

parallels999:
image

drbyte:
Alright. Given that I'd like to avoid changing contracts again, if possible, using Union Types now saves that pain.

PR #2490, opened 2023-08-30
[V5] Avoid triggering `eloquent.retrieved` event

erikn69:
when hydrating the models that come from the cache, the `retrieved` event is being triggered, it should not be triggered because they are supposed to be preloaded from cache, this decreases performance
image

[Illuminate/Database/Eloquent/Model.php#L625-L636](https://github.com/laravel/framework/blob/4989e6de076688ade265e2f1970ab6f0c1b60fcb/src/Illuminate/Database/Eloquent/Model.php#L625-L636), here is the source

Same for **V6**


drbyte:
Thanks!

PR #2492, opened 2023-09-01
[Docs] Listen to DatabaseRefreshed rather than MigrationsEnded in TestCase

Okipa:
Listening to `DatabaseRefreshed` event rather than `MigrationsEnded` can avoid issues during testing after `artisan schema:dump` command has been executed.

In fact, `MigrationsEnded` is not called when the migration is based on a `mysql-schema.sql` file and has no migration executed anymore.

`DatabaseRefreshed` is always executed in both cases.


drbyte:
Thanks!

PR #2499, opened 2023-09-13
Rename "Middlewares" namespace to "Middleware"

drbyte:
Ref #2496


drbyte:
@tabacitu will this create any huge complications with Backpack?

KevinEscobarV:
This change caused the middlewares to fail when updating the package.

`Target class [Spatie\Permission\Middleware\RoleMiddleware] does not exist.`

Middleware is still written in the documentation.

image

drbyte:
@KevinEscobarV Apologies: the online docs site is out of sync with the release. They're incorrectly showing v6 docs even though you've selected v5.
It should be sorted out as soon as v6 is released, which will be within a few days.

PR #2500, opened 2023-09-13
[v6] Reuse code on withoutRole and withoutPermission

erikn69:
on #2463, `withoutRole` and `withoutPermission` were added, but I think that duplicating the same code should be avoided.
Also I didn't understand what `$debug = false` is for
https://github.com/spatie/laravel-permission/blob/8f15d1846aaa5039292978e7775bbf0f2ca73cc8/src/Traits/HasPermissions.php#L129

Laravel do the same 

- [Illuminate/Database/Eloquent/Builder.php#orWhere](https://github.com/laravel/framework/blob/bfa85b811cf6545dd5b6be40b44a523d02897da9/src/Illuminate/Database/Eloquent/Builder.php#L335), 
- [Illuminate/Database/Eloquent/Builder.php#where](https://github.com/laravel/framework/blob/bfa85b811cf6545dd5b6be40b44a523d02897da9/src/Illuminate/Database/Eloquent/Builder.php#L294)

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -96,8 +96,9 @@ trait HasPermissions
      * Scope the model query to certain permissions only.
      *
      * @param  string|int|array|Permission|Collection|\BackedEnum  $permissions
+     * @param  bool  $without
      */
-    public function scopePermission(Builder $query, $permissions): Builder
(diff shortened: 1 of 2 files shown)


drbyte:
Great! Thanks.

PR #2513, opened 2023-10-09

adereksisusanto:


drbyte:
Thanks for your submission.

When submitting a PR it is always important to explain what you are trying to achieve with the proposed changes.

It appears that you're wanting to implement half of full uuid support by checking if a config value has been changed.

Can you explain more about how you arrived at this solution?

Issue #1290, opened 2019-11-27
@can blade with guard name not working


AhmedHdeawy:
Hi, thanks for this awesome package,

i have a problem that is:

i have multiple guards [ 'web', 'admin' ]

in Admin Model i have 
`protected $guard_name = 'admin';`

and i have guard name for each permission in
> permissions table

like the image below 
![p](https://user-images.githubusercontent.com/15928500/69720519-78669780-111b-11ea-8a1d-148146662e5f.png)

the problem is when use @can directive like: 
`@can('admin.users.view')`
it doesn't work, not show the code that in it, even if this admin is Super Admin or Admin

also when use 
`@can('admin.users.view', 'admin')`
it doesn't work also

but when use 
`@if(auth('admin')->user()->can('admin.users.view'))`
it works fine

how i can use @can only because is simple for code.

thanks in advance

sameervirus:
+1

drbyte:
You cannot pass the guard to `@can`. You must use the method you described, or maybe `hasPermissionTo`.

malickateeq:
Anyone solved this issue?

drbyte:
Nothing to "solve". It's operating by design.

itsaxay:
@AhmedHdeawy, You can create own blade directive like

```
        Blade::directive('canhas', function ($permission, $guard = "admin") {
            return "<?php if (auth()->user()->hasPermissionTo({$permission} , \"{$guard}\")) { ?>";
        });
        Blade::directive('endcanhas', function () {
            return '<?php } ?>';
        });
```

MD-Khalid-Hosain:
Solved my problem... Thank you bro @AhmedHdeawy AhmedHdeawy

jishujian:
why can't create own blade directive just like @can('','','')

Lukmauu:
@auth('admin')

Works fine if you admin guard set in your config/auth.php

axlwild:
Given this problem, I've created [a PR](https://github.com/spatie/laravel-permission/pull/2515) to be able to use another blade directive in this case, a little bit more confortable than the whole `@if(auth('admin')->user()->can('admin.users.view'))` sentence.

With this change, you can use the following directive
```
@haspermissionto('admin.users.view', 'admin')
   ...
@endhaspermissionto
```

nguyencuongcd44:
The @can does not work with my custom guard so I changed the default guard to my custom guard, and it worked. 
[...]

PR #2515, opened 2023-10-12
Has permission directive - Non Default guard

axlwild:
Has permission directive to use instead can directive when using non-default guard

Creating this PR to be able to use another blade directive when you use another guard distinct from the default one.

This issue is mentioned [here](https://github.com/spatie/laravel-permission/issues/1290).

With this change, you can use the following directive
```
@haspermission('admin.users.view', 'admin')
   ...
@endhaspermission
```
instead of 
```
@can('admin.users.view')
   ...
@can
```

### DIFF
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -123,6 +123,16 @@ class PermissionServiceProvider extends ServiceProvider
         $bladeCompiler->directive('elserole', fn ($args) => "<?php elseif({$bladeMethodWrapper}('hasRole', {$args})): ?>");
         $bladeCompiler->directive('endrole', fn () => '<?php endif; ?>');
 
+        $bladeCompiler->directive('haspermissionto', function ($arguments) {
+            return "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
+        });
+        $bladeCompiler->directive('elsehaspermissionto', function ($arguments) {
+            return "<?php elseif(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
+        });
+        $bladeCompiler->directive('endhaspermissionto', function () {
+            return '<?php endif; ?>';
+        });
+
         $bladeCompiler->directive('hasrole', fn ($args) => "<?php if({$bladeMethodWrapper}('hasRole', {$args})): ?>");
         $bladeCompiler->directive('endhasrole', fn () => '<?php endif; ?>');
 
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -123,15 +123,9 @@ class PermissionServiceProvider extends ServiceProvider
         $bladeCompiler->directive('elserole', fn ($args) => "<?php elseif({$bladeMethodWrapper}('hasRole', {$args})): ?>");
         $bladeCompiler->directive('endrole', fn () => '<?php endif; ?>');
 
-        $bladeCompiler->directive('haspermissionto', function ($arguments) {
-            return "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
-        });
-        $bladeCompiler->directive('elsehaspermissionto', function ($arguments) {
-            return "<?php elseif(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
-        });
-        $bladeCompiler->directive('endhaspermissionto', function () {
-            return '<?php endif; ?>';
-        });
+        $bladeCompiler->directive('haspermissionto', fn ($arguments) => "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>");
+        $bladeCompiler->directive('elsehaspermissionto', fn ($arguments) => "<?php elseif(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>");
+        $bladeCompiler->directive('endhaspermissionto', fn () => '<?php endif; ?>');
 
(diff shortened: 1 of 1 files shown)


src/PermissionServiceProvider.php
```diff
@@ -123,6 +123,16 @@ protected function registerBladeExtensions($bladeCompiler): void
         $bladeCompiler->directive('elserole', fn ($args) => "<?php elseif({$bladeMethodWrapper}('hasRole', {$args})): ?>");
         $bladeCompiler->directive('endrole', fn () => '<?php endif; ?>');
 
+        $bladeCompiler->directive('haspermissionto', function ($arguments) {
+            return "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
```
parallels999, line 127: Why not follow the syntax of the other directives?

axlwild: I tried to follow that syntax. Would you recommend `@haspermission` or another one better?

parallels999: When i say sintax, means PHP Sintax, not naming
```php
$bladeCompiler->directive('haspermissionto', fn ($args) => "<?php if({$bladeMethodWrapper}('hasPermissionTo', {$args})): ?>");
```

parallels999:
@drbyte hi, couldn't this be solved by adding the guard on the gate directive?
https://github.com/spatie/laravel-permission/blob/a10d72d40ec41e3e65fe390864667f4c2dec2fb7/src/PermissionRegistrar.php#L125-L129
```php
$gate->before(function (Authorizable $user, string $ability, array $args = []) { 
     if (method_exists($user, 'checkPermissionTo')) { 
         $guard = is_string($args[0] ?? null) ? $args[0] : null;
         return $user->checkPermissionTo($ability, $guard) ?: null; 
     } 
 }); 
```
Now we could do
```blade
@can('admin.users.view', 'admin')
   ...
@endcan
```

src/PermissionServiceProvider.php
```diff
@@ -123,15 +123,9 @@ protected function registerBladeExtensions($bladeCompiler): void
...
-        $bladeCompiler->directive('haspermissionto', function ($arguments) {
-            return "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
-        });
-        $bladeCompiler->directive('elsehaspermissionto', function ($arguments) {
-            return "<?php elseif(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>";
-        });
-        $bladeCompiler->directive('endhaspermissionto', function () {
-            return '<?php endif; ?>';
-        });
+        $bladeCompiler->directive('haspermissionto', fn ($arguments) => "<?php if(\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper('hasPermissionTo', {$arguments})): ?>");
```
parallels999, line 126: Why `\\Spatie\\Permission\\PermissionServiceProvider::bladeMethodWrapper`, there is `$bladeMethodWrapper}`, it's the same, also `$args` for consistency
```php
$bladeCompiler->directive('haspermissionto', fn ($args) => "<?php if({$bladeMethodWrapper}('hasPermissionTo', {$args})): ?>");
```

src/PermissionServiceProvider.php
parallels999: you forgot `$args` `$bladeMethodWrapper`

src/PermissionServiceProvider.php
```diff
@@ -123,6 +123,10 @@ protected function registerBladeExtensions($bladeCompiler): void
         $bladeCompiler->directive('elserole', fn ($args) => "<?php elseif({$bladeMethodWrapper}('hasRole', {$args})): ?>");
         $bladeCompiler->directive('endrole', fn () => '<?php endif; ?>');
 
+        $bladeCompiler->directive('haspermission', fn ($args) => "<?php if({$bladeMethodWrapper}('hasPermissionTo', {$args})): ?>");
```
erikn69, line 126: `hasPermissionTo` throws an exception, must be `checkPermissionTo`
https://github.com/spatie/laravel-permission/blob/a10d72d40ec41e3e65fe390864667f4c2dec2fb7/src/Traits/HasPermissions.php#L254-L261

drbyte:
> adding the guard on the gate directive

That would be preferable.  Until now we've avoided adding blade helpers for permissions because they basically duplicate what Laravel already provides in the can/canany/canall series.

A few versions ago I explored your suggestion when tinkering with making guards simpler but all kinds of things broke .... but I think the part you mention might have been fine. I rolled back the whole thing, but probably could have kept that part.
It's worth writing some tests for it and giving it a try!

drbyte:
Current status: undecided between merging this (after tests are added for this new directive), or simply adding it to the `can` logic...

parallels999:
I did test it with the `if(is_string($args[0] ?? null) && !class_exists($args[0])) { $guard = array_shift($args); }` and it works, but i don't know if `protected $policies` could be used with custom strings, on that case that it will fail

drbyte:
Ya, we don't want to break the normal Policy stuff.

drbyte:
@parallels999 Maybe I've been at the keyboard too long today, but I can't get a Blade test to pass when passing a guard.

eg: Added this to `BladeTest.php`:
```php
    /** @test */
    public function the_can_directive_can_accept_a_guard_name()
    {
        $this->testAdminRole->givePermissionTo($this->testAdminPermission);
        $this->testAdmin->assignRole($this->testAdminRole);
    
        // log in as the Admin with the permission-via-role
        auth('admin')->setUser($this->testAdmin);

        $permission = 'admin-permission'; // from $this->testAdminPermission
        $guard = 'admin'; // from $this->testAdminPermission
        $this->assertEquals('has permission', $this->renderView('can', compact('permission', 'guard')));
    }
```

drbyte:
Just thinking "aloud" ... Regarding policies ... since this package is only checking for the permission names provided by this package, does it really need to also check against a policy? Policies should be checking the ability (permission name) against the model, inside the policy, so I don't think we need to worry about cascading the arguments list down to policy names like core Laravel does.

drbyte:
@axlwild ... still considering this. But I'd like to add tests. Can you update the PR to "allow edits by maintainers"? (tick the checkbox)

PR #2526, opened 2023-10-24
Add guard parameter to can()

drbyte:
As discussed in comments in: https://github.com/spatie/laravel-permission/pull/2515#issuecomment-1762134150

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -122,9 +122,12 @@ class PermissionRegistrar
      */
     public function registerPermissions(Gate $gate): bool
     {
-        $gate->before(function (Authorizable $user, string $ability) {
+        $gate->before(function (Authorizable $user, string $ability, array &$args = []) {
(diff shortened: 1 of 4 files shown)


drbyte:
NOTE: Undecided whether this is actually safe, or whether it will break when `can()` is used with additional parameters such as Model names or Model instances.

erikn69:
Seems like a good feature, hopefully it doesn't break any custom implementations 👍

parallels999:
Maybe middlewares need an update
https://github.com/spatie/laravel-permission/blob/c66c0de99a95a88288507fa96676b89625f54488/src/Middleware/PermissionMiddleware.php#L35
https://github.com/spatie/laravel-permission/blob/c66c0de99a95a88288507fa96676b89625f54488/src/Middleware/RoleOrPermissionMiddleware.php#L35

```php
if (! $user->canAny($permissions, $guard)) { 
///
if (! $user->canAny($rolesOrPermissions, $guard) && ! $user->hasAnyRole($rolesOrPermissions)) { 
```

PR #2534, opened 2023-10-30
[Docs] Update unsetRelation() example in teams-permissions.md

shdehnavi:
In the teams-permissions doc, `unsetRelation()` method invoked with 2 parameters, but the method signature uses 1 parameter.
See the [Method Reference](https://github.com/illuminate/database/blob/571aa694dc6f588d4334e4e23d6721ff68893e10/Eloquent/Concerns/HasRelationships.php#L863)

I made the changes and we need to call `unsetRelation()` method twice for `models` and `permissions` relations


drbyte:
Thanks!

PR #2539, opened 2023-11-06
[Docs] Update link in direct-permissions.md

sevannerse:
Update missing link.

### DIFF
--- a/docs/basic-usage/direct-permissions.md
+++ b/docs/basic-usage/direct-permissions.md
@@ -7,7 +7,7 @@ weight: 2
 
 It's better to assign permissions to Roles, and then assign Roles to Users.
 
-See https://spatie.be/docs/laravel-permission/best-practices/roles-vs-permissions for a deeper explanation.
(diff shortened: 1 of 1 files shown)


parallels999:
it works for me

drbyte:
Thanks!

parallels999:
@drbyte seriusly?? this breaks spatie docs, try https://spatie.be/docs/laravel-permission/v6/basic-usage/direct-permissions, and click on the link, now it is missing and redirects to index

drbyte:
Interesting.  
I checked before merging: the docs website was displaying the URL as plain text, not clickable.

drbyte:
Will do some digging.
Will check other links too. We don't do a lot of cross-linking in these docs, so there aren't many links to double-check.

parallels999:
> was displaying the URL as plain text, not clickable

Yes, but the url point to `.md`

drbyte:
Updated

PR #2546, opened 2023-11-09
fix empty permissions in PermissionRegistrar.php

### LINKED ISSUE #2543: [v6] There is no permission named '1' for guard 'web'.
**Describe the bug**
I'm using basic spatie settings  for controlling roles&permissions in my laravel 10 project.

The issue is that I keep facing :
```
**Spatie \ Permission \ Exceptions \ PermissionDoesNotExist
PHP 8.2.4 10.30.0
There is no permission named `whatever***` for guard `web`.** 
```

which happened since upgrading to v6-6.0.1  while using standard commands such as 

  1.  $role->syncPermissions($permissions); and  2. $user->assignRole($request->roles);
  
also I'm using default laravel 'web' guard.

When I'm downgrading to V5.11 or 5.8 I get no error and everything works perfectly !

**Versions**
- spatie/laravel-permission package version:  _6 - 6.0.1_ 
- illuminate/framework package:       _v10.30.0 Laravel Framework._

PHP version:
_8.2.4_

Database version:

_MariaDB 10.4.28_

**To Reproduce**

```
 $role = Role::findOrFail($id);
        $permissions = $request->permission;

        if (!empty($permissions)) {
            $role->syncPermissions($permissions);
        }
```
&
```
   if ($request->roles) {
            $user->assignRole($request->roles);
        }
```

**Example Application**

https://github.com/falkonerX72/L.estate

 - OS: windows
 - Version 11

patrickweh:
hey,

bool check on an empty collection returns true so the cache is never populated. I came across this when doing syncPermissions on my user model threw an exception that the permission doesnt exist event when it does.

This fixes #2543 

greetings

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -191,7 +191,7 @@ class PermissionRegistrar
      */
     private function loadPermissions(): void
     {
-        if ($this->permissions) {
+        if (count($this->permissions ?? [])) {
             return;
         }
 


parallels999:
It seems that your problem is different, it is not related to #2543, if `$this->permissions` is a collection, means that the cache was correctly reloaded, and it is empty.
It seems to me that you are making some kind of manual change to the permissions system(_maybe a seeder_), and you are not adding `forgetCachedPermissions` afterwards.

patrickweh:
yes im seeding permissions, thats correct but even if i do a cache:clear this if statement is always returning true.

its "kind of" related to the issue. My first try was to cast all the permissions into from string to int, which resulted in a permission with id 2 not found for guard web.

Anyway, checking a collection with if will always result in true, even when the collection is empty.

parallels999:
`cache:clear` doesn't work on this case, you must add `app()->make(\Spatie\Permission\PermissionRegistrar::class)->forgetCachedPermissions();` afterwards

drbyte:
> That is the expected behavior, it is trying to check that `$this->permissions` is `not null`

Correct. If the `$this->permissions` property is null or undefined then it needs to load/reload the permissions. Otherwise we don't force the reload.

drbyte:
@patrickweh if you still believe there's a bug to be fixed here, please provide code-samples (where the problem is easily replicatable) and ideally a test that demonstrates the issue.

PR #2403, opened 2023-04-13
[V6] Register OctaneReloadPermissions listener for Laravel Octane

erikn69:
i'm not 100% sure on this 
i can't test octane 

this was a test for octane users to try but no one gave me feedback that it works

On https://github.com/spatie/laravel-permission/issues/2322#issuecomment-1493242776 was said that the problem was the **spatie/once** package, and that **spatie/laravel-permission** works correctly on octane

**I recommend not merging it until you do more testing that it's the correct approach.**

PR #2547, opened 2023-11-09
Reset teamId on octane

### LINKED ISSUE #2545: [Octane] Potential security issue by not resetting teamId?
_Sorry, I haven't reproduced this as an issue, but spotted a potential security issue while reading through the code._

In the context of Octane where the `PermissionRegistrar` is a singleton I believe that the `protected $teamId` field should be set to `null` on every Octane tick.

In most cases it wouldn't be an issue as a middleware should set the `teamId` to the correct team, but in certain circumstances this might not be the case and the teamId would persist between requests. This could be a serious security issue as an admin in one team may get admin permission to another team.

As the `OctaneReloadPermissions` listener is optional, I believe a second listener could be created to **always** reset the field with e.g `setPermissionsTeamId(null);`.

erikn69:
Closes #2545


drbyte:
I notice that the Spatie Multitenancy package listens on `\Laravel\Octane\Events\RequestTerminated`.
Is that a better spot, or are we satisfied with `\Laravel\Octane\Events\OperationTerminated`?

erikn69:
`OperationTerminated` includes `RequestTerminated`, `TaskTerminated`, `TickTerminated`, but I don't know the scope of the others two
https://github.com/spatie/laravel-permission/pull/2403#discussion_r1170529682
https://github.com/spatie/laravel-permission/pull/2403#issuecomment-1509301579

drbyte:
Thanks!

PR #2552, opened 2023-11-12
permission revoke based on ids as well

tariqali021:


parallels999:
There is no change

parallels999:
Still, is https://github.com/spatie/laravel-permission/pull/2552/commits/09ebf39819d0f1a83dc008a1f318898e8a450a13 a refactor?

tariqali021:
Yes I found some tweaks & fix it.. my docker is crashing & unable to test.. would be great if you review it carefully..

parallels999:
just do `composer test`

parallels999:
Seems that #2548 does the same, test it

tariqali021:
That's something else.. This change belongs to revoking permission given as int (primary key of permission table). Currently, it is only supporting based on name.

parallels999:
> This change belongs to revoking permission given as int (primary key of permission table). Currently, it is only supporting based on name.

That is exactly what #2548 is adding

tariqali021:
Yes it is also doing the same what I'm trying to achieve but to me it seems not optimized way to do that.. correct me if I'm wrong. That code is finding each individual record from db either based on it or string. but the good thing is that it is revoking based on both primary key & name column

parallels999:
> That code is finding each individual record from db either based on int or string

On `removeRole` yes, on `revokePermissionTo` no, permissions are getting from cache, not from DB

tariqali021:
yes but for very rare and edge case if we don't have cache stored at certain point. but still acceptable because of edge case.

parallels999:
The documentation should also be updated, it says that it only accepts string or Permission object

tariqali021:
yes and also `enum`

Issue #1094, opened 2019-05-07
Enable Eager Loading Roles/Permissions


ragingdave:
In attempting to call getAllPermissions on a model that has already eager loaded the roles and permissions, causes database queries because the getPermissionViaRoles isn't checking if the roles and roles.permissions are already loaded and calling the database regardless of eager loaded state.

a quick fix might be replacing the load call with loadMissing in getPermissionsViaRoles but I'm not sure if that's the right way to go with it. If it is I would gladly write up a PR with that fix, just wanted to see if there was a reason currently that isn't is place.

drbyte:
I'd entertain a PR with tests showing that it solves the problem stated.

NurdTurd:
👍 Just realized for showing comments with user roles as titles, I'm pulling around 15 queries per request. Oof!

drbyte:
Hmmm ... awhile back I was working on a PR to remove some eager loading because it was doing unnecessary queries. Different spot perhaps, but dropping mention of it here for reference https://github.com/spatie/laravel-permission/pull/1056

ragingdave:
In working the fix and tests, I found that it ends up breaking a test in the cache section, which I'm definitely not familiar with yet, so this PR might take a bit more time to complete. Just wanted to give an update. @NurdTurd, temporarily, if you just override the trait method then you can swap the load with loadMissing, but there may be dragons since the the caching could be affected.

NurdTurd:
@ragingdave Thanks for the response, how would I go about changing the method? Do I have to modify vendor stuff? That feels a bit iffy.

NessterGit:
While I understand the need for the eager loading, the loadMissing() behavior bit me in the arse. In case someone lands here from research, and you want to reset the roles and permissions for a new user check or the same user but different team (my use case), you can use 
```
$this->unsetRelation('roles') 
$this->unsetRelation('permissions') 
```
for example in your switchTeam() function before you do the actual team switch. 

Or override the getPermissionsViaRoles() function in your user model (replacing loadMissing() with load()). But this can cause a LOT more queries. Handle with care ;)

drbyte:
@NessterGit what version of this package were you encountering that with?  Do you mind also testing against `v6.0-dev` because we applied some changes in #2412 which addresses relation loading, albeit not inside the getPermissionsViaRoles() method.

NessterGit:
[...]

PR #2553, opened 2023-11-13
Issue assigning team roles when relationship is preloaded

jameshulse:
### Overview

I ran into an issue assigning the same role to a user across multiple teams. The issue I ran into is as follows:

Say I have two teams in the database: "Team1" and "Team2". I want to load users from the database and assign them some role. Let's call it "testRole". They already have "testRole" on Team1, but I want to assign it for Team2.

If the "roles" relationship has been pre-loaded then the line:

    $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();

In the `HasRoles` trait will see the role on Team1, and not add it for Team2.

### How I ran into the issue

Initially I was loading users from the database and assigning roles in a seeder like so:

```php
setPermissionsTeamId(2);

$users = User::where(...)->get();

foreach ($users as $user) {
    $user->assignRole('testRole');
}
```

As we have [lazy loading disabled](https://laravel.com/docs/10.x/eloquent-relationships#preventing-lazy-loading) in our app this causes an issue in the `spatie/laravel-permission` package [here](https://github.com/spatie/laravel-permission/blob/main/src/Traits/HasRoles.php#L155) as it lazy loads the roles relationship:

```php
$currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();
```

So I updated my snippet above to preload the roles:

```php
setPermissionsTeamId(2);

$users = User::with('roles')->where(...)->get();

...assign roles
```

But now I run into the issue that the role collection has a role with the key `testRole`, so when assigning it in `assignRole` the attachment which uses an `array_diff` won't actually attach a new role to the second team.

### Attempted Solution

As this didn't seem quite right, I have added a unit test to the package to try and reproduce the issue. I naively thought that I can just reload the roles relationship eagerly before we diff the roles, but then other tests fail which are confirming that we don't make unnecessary SQL queryies. Of course we are making a new query so I don't know how to proceed.

Should we allow the roles to be re-loaded, and update the number of allowed SQL queries in the test to be +1? Or is there a better solution here? Maybe if the `roles` relationship is pre-loaded then the line that gets the current role keys can use the current team ID to filter to just the current team?

Something like this (pseudo code):

```php
// Trait\HasRoles.php

public function assignRole(...$roles) {
    ...

    if ($model->exists) {
        $currentRoles = $this->roles
            ->filter(fn $role => $role->pivot->{$teamsKey) == getPermissionsTeamId()) // New filter
            ->map(fn ($role) => $role->getKey())
            ->toArray();

        $this->roles()->attach(array_diff($roles, $currentRoles), $teamPivot);
        $model->unsetRelation('roles');
    } else {
        ...
    }

    ...
}
```

Hopefully we can come to some resolution, if this isn't clear or you have any questions then please feel free to reach out.

Thanks!

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -152,6 +152,8 @@ trait HasRoles
             [app(PermissionRegistrar::class)->teamsKey => getPermissionsTeamId()] : [];
 
         if ($model->exists) {
+            $this->load('roles');
+
             $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();
 
             $this->roles()->attach(array_diff($roles, $currentRoles), $teamPivot);
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -16,6 +16,8 @@ trait HasRoles
 
     private ?string $roleClass = null;
 
+    private mixed $rolesLoadedWithTeamId = null;
+
     public static function bootHasRoles()
     {
         static::deleting(function ($model) {
@@ -60,6 +62,8 @@ trait HasRoles
         }
 
         $teamField = config('permission.table_names.roles').'.'.app(PermissionRegistrar::class)->teamsKey;
+
+        $this->rolesLoadedWithTeamId = getPermissionsTeamId();
 
         return $relation->wherePivot(app(PermissionRegistrar::class)->teamsKey, getPermissionsTeamId())
             ->where(fn ($q) => $q->whereNull($teamField)->orWhere($teamField, getPermissionsTeamId()));
@@ -152,7 +156,9 @@ trait HasRoles
             [app(PermissionRegistrar::class)->teamsKey => getPermissionsTeamId()] : [];
 
         if ($model->exists) {
-            $this->load('roles');
+            if (getPermissionsTeamId() !== $this->rolesLoadedWithTeamId) {
+                $this->load('roles');
+            }
 
             $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();
 
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -16,8 +16,6 @@ trait HasRoles
 
     private ?string $roleClass = null;
 
-    private mixed $rolesLoadedWithTeamId = null;
-
     public static function bootHasRoles()
     {
         static::deleting(function ($model) {
@@ -62,8 +60,6 @@ trait HasRoles
         }
 
         $teamField = config('permission.table_names.roles').'.'.app(PermissionRegistrar::class)->teamsKey;
-
-        $this->rolesLoadedWithTeamId = getPermissionsTeamId();
 
         return $relation->wherePivot(app(PermissionRegistrar::class)->teamsKey, getPermissionsTeamId())
             ->where(fn ($q) => $q->whereNull($teamField)->orWhere($teamField, getPermissionsTeamId()));
@@ -156,9 +152,7 @@ trait HasRoles
             [app(PermissionRegistrar::class)->teamsKey => getPermissionsTeamId()] : [];
 
         if ($model->exists) {
-            if (getPermissionsTeamId() !== $this->rolesLoadedWithTeamId) {
-                $this->load('roles');
-            }
+            $this->loadMissing('roles');
 
             $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();
 
--- a/tests/LazyLoadingTest.php
+++ b/tests/LazyLoadingTest.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Spatie\Permission\Tests;
+
+class LazyLoadingTest extends TestCase
+{
+    /** @test */
+    public function test_it_does_not_violate_global_lazy_loading_rule()
+    {
+        $this->testUser->preventsLazyLoading = true;
+        $this->testUser->wasRecentlyCreated = false;
+
+        $this->expectNotToPerformAssertions();
+
+        $this->testUser->assignRole('testRole');
+    }
+}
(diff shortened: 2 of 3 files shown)


parallels999:
That is the expected behavior, there is no support across teams. This change would reload the roles for those who do not use teams, which is the majority

If you review other tests and examples you will see that when changing the team the relationships must always be reloaded.

https://github.com/spatie/laravel-permission/blob/641bc2f10d7d5db62f8463a5fc011eb6ce24d3df/tests/TeamHasRolesTest.php#L72-L73
https://github.com/spatie/laravel-permission/blob/641bc2f10d7d5db62f8463a5fc011eb6ce24d3df/tests/TeamHasPermissionsTest.php#L30-L31

jameshulse:
> This change would reload the roles for those who do not use teams, which is the majority

This makes sense, but I would hope there is a solution that we can reach.

I had an idea that we could somehow hook into when the 'roles' relationship is loaded and save the current team ID to a field on the model, then, we only need to reload the roles during assignment etc if the current team ID is different to the one that was set when the roles relationship was last loaded. Does that make sense?

Something like:

```php
setPermissionsTeamId(1)

$user = User::with('roles')->first(); // Some field inside `HasRoles` is set to the current team ID of 1

$user->assignRole('testRole');

setPermissionsTeamId(2);

$user->assignRole('testRole'); // Inside assignRole we check the current team ID (2) vs the ID when the roles were loaded (1), and refresh roles as it is different
```

drbyte:
Over the years, in this package we've gone back and forth between eager-loading and lazy-loading relations under various conditions.

Presently we unset the relations after certain operations in order to allow them to be reloaded only when needed to perform a certain operation that depends on them. In a couple spots we expressly eager-load, again due to the need within that context.

While I'm not thrilled at the additional complexity introduced, I'd entertain a discussion which could lead to a tested PR about offering a configuration flag (perhaps `use_lazy_loading`) where the developer could tell the package that all lazy-loading has been disabled for the entire application, and in that app we could intentionally eager-load according to that flag.

Unfortunately the `setPermissionsTeamId()` method is user-agnostic, so we can't auto-unset those relations when changing teamId.

drbyte:
> somehow hook into when the 'roles' relationship is loaded and save the current team ID to a field on the model, then, we only need to reload the roles during assignment etc if the current team ID is different

Interesting. Ya, some sort of internally cached value to compare against, even if it's just a temporary property that's not persisted to the db.

Or ... maybe since you know exactly for your app that this situation exists (particularly because of the disabled lazy-loading), perhaps every time you use setPermissionsTeamId you know that you can't loop through an array of users without including a reload of relations on each model, so instead of the generic:

```php
// generic, dependent upon lazy-loading:
setPermissionsTeamId(2);

$users = User::where(...)->get();

foreach ($users as $user) {
    $user->assignRole('testRole');
}
```

becomes:

```php
setPermissionsTeamId(2);

$users = User::where(...)->get();

foreach ($users as $user) {
    // must eager-load the relations:
    $user->load('roles');      <------------------------ added this
    $user->assignRole('testRole');
}
```

drbyte:
> I'd entertain a discussion which could lead to a tested PR about offering a configuration flag (perhaps use_lazy_loading) where the developer could tell the package that all lazy-loading has been disabled for the entire application, and in that app we could intentionally eager-load according to that flag.

Thinking more on this, I guess there are some options to explore, as far as the lazy-loading portion goes:
- We could query the (user) model for `->preventsLazyLoading` directly on the model, and eager-load it in such case.
- We could query `Model::preventsLazyLoading()` app-wide and eager-load in that case.

jameshulse:
Thanks @drbyte for the suggestion. I think that will be the solution for our app if we can't find a solution - but it did take some debugging to work out that there was an issue, so other users may be affected.

I actually just pushed up a commit which tracks the team ID between loads - and all the unit tests pass now! Please take a look, I think it could be promising.

jameshulse:
I would need to try out my fix in my app to see if it still works with lazy loading disabled.

And then we may need to look at your solution of querying for lazy loading - maybe this is easier as a configuration setting.

drbyte:
If the situation in your app is because data is stale due to not lazy-loading, then if we are intentional to eager-load in the right places when that feature is set in the app, it should resolve the stale-data situation. I think this package should be responsible for not having stale data insomuch as it has control over that when people use its interface/methods.
I think I'd prefer to avoid the temporary property I suggested earlier if the problem is solved by knowing "when" to eager-load.

jameshulse:
> if we are intentional to eager-load in the right places when that feature is set in the app, it should resolve the stale-data situation

This is a good point. I am kind of working around the issue that arose from the lazy loading violation.

I can have a go at detecting the right time to use eager loading in the package.

parallels999:
same "team tracking" here:
https://github.com/spatie/laravel-permission/blob/641bc2f10d7d5db62f8463a5fc011eb6ce24d3df/src/Traits/HasPermissions.php#L395-L396

jameshulse:
> I think I'd prefer to avoid the temporary property I suggested earlier if the problem is solved by knowing "when" to eager-load.

I have updated `assignRole` to load roles if it is missing (and removed the temporary field completely):

```php
    public function assignRole(...$roles)
    {
        $roles = $this->collectRoles($roles);

        ...

        if ($model->exists) {
            // This check is new
            $this->loadMissing('roles');

            $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();

            ...
        }
    }
```

And added a test which enforces the lazy loading violation rule.

As expected we no longer get a lazy loading violation **but**, the test I added which assigns a role in one team, changes teams and assigns another without reloading the relationship still fails.

This might show that the initial test I created is a bad test, and users won't be eagerly loading 'roles' and changing teams now that the eager load violation goes away. I will remove the test for now unless you think it is valuable.

I can imagine people may still run into issues changing teams with relationships pre-loaded, but maybe we don't have to solve this issue now (in this PR).

drbyte:
Related #1094 , #2412

PR #2556, opened 2023-11-15
Hopefully fix issue for testing locally

SuperDJ:
While testing locally I noticed that I got an `403` error even though I was using `Passport::actingAsClient()`. That error occurs because there is no `$user` and no `bearerToken()` provided. But I did notice that `auth()?->client()` returned the `Passport::actingAsClient()` `Client` instance.  

So I'm hoping that with this change testing works as expected. It should now check for either the `bearerToken()` or the loggedin Client provided by  `Passport::actingAsClient()`.

### DIFF
--- a/src/Middleware/PermissionMiddleware.php
+++ b/src/Middleware/PermissionMiddleware.php
@@ -16,7 +16,7 @@ class PermissionMiddleware
         $user = $authGuard->user();
 
         // For machine-to-machine Passport clients
-        if (! $user && $request->bearerToken() && config('permission.use_passport_client_credentials')) {
+        if (! $user && ($request->bearerToken() || auth()?->client()) && config('permission.use_passport_client_credentials')) {
             $user = Guard::getPassportClient($guard);
         }
 
--- a/src/Middleware/RoleMiddleware.php
+++ b/src/Middleware/RoleMiddleware.php
@@ -16,7 +16,7 @@ class RoleMiddleware
         $user = $authGuard->user();
 
         // For machine-to-machine Passport clients
-        if (! $user && $request->bearerToken() && config('permission.use_passport_client_credentials')) {
+        if (! $user && ($request->bearerToken() || auth()?->client()) && config('permission.use_passport_client_credentials')) {
             $user = Guard::getPassportClient($guard);
         }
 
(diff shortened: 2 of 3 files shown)


drbyte:
While my experience with Passport is limited, the changes you are proposing here so far make sense.

Are you satisfied that these changes have resolved the testing problem you were encountering?

SuperDJ:
@drbyte I reset the file to the current situation and run the tests, they failed. Than changed it with this change and they passed. 

I did wonder if for the sake of readability the if statement should become a method?

parallels999:
`getPassportClient` already checks `client()`, 
Maybe you are doing something wrong in your tests, it would be great if you published an example, maybe your problem can be fixed without changing the library
https://github.com/spatie/laravel-permission/blob/c66c0de99a95a88288507fa96676b89625f54488/src/Guard.php#L91-L93
https://github.com/spatie/laravel-permission/blob/c66c0de99a95a88288507fa96676b89625f54488/src/Guard.php#L95
https://github.com/spatie/laravel-permission/blob/c66c0de99a95a88288507fa96676b89625f54488/src/Guard.php#L81-L106

parallels999:
It seems that this was expected, they added a dump bearer
https://github.com/spatie/laravel-permission/blob/c66c0de99a95a88288507fa96676b89625f54488/tests/TestCase.php#L275-L277

parallels999:
> `auth()?->client()` returned the `Passport::actingAsClient()` Client instance.

maybe you forget the guard
```php
Passport::actingAsClient($YOUR_CLIENT, [], 'guard_name_here');
```
**Source:** [laravel/passport/Passport.php#L385](https://github.com/laravel/passport/blob/eb78ec6c192a01fbbe76dd427630a5e500e2486a/src/Passport.php#L385)

SuperDJ:
@parallels999 the way I handle `Passport::actingAsClient` is like the following:

```php
// Some test
public function test()
{
   $this->actingAsClient();
}

// Passport actingAsClient wrapper method
protected function actingAsClient( client = null, $permissions = null )
    {
        if( !$client )
        {
            $client = Client::factory()->create();
        }
        Passport::actingAsClient(
            $client,
            [ '*' ]
        );

        if( $permissions )
        {
            $this->assignRole(<some params>);
        }

        return $this;
    }
```

From what I understand is that `Passport::actingAsClient` simulates a client being loggedin. So to me that is assuming that a bearer token is being passed. Unfortunately that appears not to be the case.

parallels999:
That example doesn't help much. 

You could try to make a PR to Passport so that it adds the dump bearer token automatically

PR #2557, opened 2023-11-15
Support guard_name on middleware

erikn69:
I add some tests
When a model uses more than one guard, it seems that the permissions cannot be reached

I don't know if it is an issue or expected behavior

----

Also, on next test, `testRole` has `web` guard_name, but login and `roleMiddleware` uses `api`, 
give me `200`,Shouldn't it be 403?
```php
/** @test */
public function multiple_guard_user_can_access_role_while_login_using_specific_guard2(): void
{
    config()->set('auth.guards.api.driver', 'session');
    Auth::guard('api')->login($this->testUser);

    $this->testUser->assignRole('testRole');

    $this->assertEquals(
        403,
        $this->runMiddleware($this->roleMiddleware, 'testRole', 'api')
    );
}
```

### DIFF
--- a/src/Middleware/PermissionMiddleware.php
+++ b/src/Middleware/PermissionMiddleware.php
@@ -32,7 +32,7 @@ class PermissionMiddleware
             ? $permission
             : explode('|', $permission);
 
-        if (! $user->canAny($permissions)) {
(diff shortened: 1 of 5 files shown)


drbyte:
Right. Good points.

Yes, should add that test for the 403.

PR #2559, opened 2023-11-16
Fixes middleware namespace in docs

3rgo:
The middleware namespace shown in the documentation was incorrect (missing an "s")


parallels999:
It's v6 docs, the "s" is on v5

drbyte:
**parallels999** is correct.  The `v6` api changed slightly from `v5`. Be sure to consult the documentation for the version of the package that you are using.

3rgo:
My bad, didn't even see it was v6, and I was still on v5

PR #2595, opened 2024-01-22
Update middleware.md

farhadmpr:
Using \Middleware\ instead of \Middlewares\ in Laravel 10


parallels999:
V5 vs V6, read again, 
​you got it all wrong

PR #2601, opened 2024-02-06
Added Team-Knowledge to Permission-Model roles relation

renky:
In huge team-environments the Permission-Model eager loads always all roles of all teams independent of a current used team-id. This can lead to much ram usage due to multiple thousands loaded role-models.

This fix finally reduces amount of loaded role-models significantly.

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -57,12 +57,20 @@ class Permission extends Model implements PermissionContract
      */
     public function roles(): BelongsToMany
     {
-        return $this->belongsToMany(
+        $relation = $this->belongsToMany(
             config('permission.models.role'),
             config('permission.table_names.role_has_permissions'),
             app(PermissionRegistrar::class)->pivotPermission,
             app(PermissionRegistrar::class)->pivotRole
         );
+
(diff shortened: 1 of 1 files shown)


parallels999:
**Also you needs to add some tests, to avoid breaking it again in the future, and to show that a problem is fixed**

--- 

I think this breaks cache feature, because all possible permission-role configurations are saved in cache
https://github.com/spatie/laravel-permission/blob/0e593bb9e6808cc55c67f150878b0847f20b9036/src/PermissionRegistrar.php#L284-L287
Maybe this works in your specific use case, you could have your own custom Permission model

---

>In huge team-environments the Permission-Model eager loads always all role

Also `Model-Role` relation already has that code, because Model uses `hasRoles` trait
After cache feature(first load only), role models must be with teams id query on `Model-Role` and cached on `Permission-Role`
If that query is being executed, it is because you implemented something wrong, or the problem is somewhere else.
https://github.com/spatie/laravel-permission/blob/0e593bb9e6808cc55c67f150878b0847f20b9036/src/Traits/HasRoles.php#L48-L66

PR #2606, opened 2024-02-07
Update middleware.md

mrpritchett:
Incorrect path for middleware docuemntation.


drbyte:
Thanks for your contribution. 
You've submitted a PR to the `v5` branch where the docs reflect the namespace that was used in `v5` and older.
However, the current release is `v6`, and the v6 documentation shows the correct new namespace.

Changing the docs according to this PR would make them inconsistent with the actual package code.

PR #2607, opened 2024-02-08
fix(team): Add nullable team_id

Androlax2:
Added `null` to helpers `getPermissionsTeamId` and `setPermissionsTeamId` and in the `PermissionRegistrar`.

This is to fix a PHPStan problem : 

`Parameter #1 $id of function setPermissionsTeamId expects Illuminate\Database\Eloquent\Model|int|string, null given.`

But when you put `''`, my tests are failing because it expects a `null` as per the doc.

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -109,7 +109,7 @@ class PermissionRegistrar
     }
 
     /**
-     * @return int|string
+     * @return int|string|null
      */
     public function getPermissionsTeamId()
     {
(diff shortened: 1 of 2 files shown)


drbyte:
@erikn69 any downsides to this?

erikn69:
It doesn't affect anything
Seems like a fix to me
https://github.com/spatie/laravel-permission/blob/3e4630f3c2160d952ad96023b32b6c186a62ec5f/src/PermissionRegistrar.php#L101-L103

drbyte:
Thanks!

func0der:
Could have added the type to the property, too, since it is initialized with `null`.
Would make the whole thing more coherent.

PR #2616, opened 2024-02-17
feat(Roles): Support for casting role names to enums

### LINKED ISSUE #2609: Enums shouldn't be casted in the model
**Describe the bug**
I am using an enum for my roles. It is called `UserRoleEnum`. I somehow thought that I need to cast the enum when saving in the database. So I added the following to my `Role`-Model
```php
    protected $casts = [
        'name' => UserRoleEnum::class,
    ];
```
I know now that this was not needed/causes the package to not work correctly anymore.

**Versions**
- spatie/laravel-permission package version: 6.3.0
- laravel/framework package: v10.43.0

PHP version: 8.3.1

Database version: MySQL 8.2.0

**To Reproduce**
* Setup Laravel and the permission package
* Use an enum for the Role
* Cast the enum in the Model
* Assign a user any role
* Use `hasRole(UserRoleEnum::<ROLE>)` and it will always return false

image
The first output are the user roles, the second one the name of the first role and the third one the `hasRole`-check
image

**Example Application**
I didn't create one so far. If you prefer to have one I can set one up. 

**Expected behavior**
I am not sure if it is a bug or a missing information in the documentation (I know that it should say not to cast but it would have saved me a few hours). Both would be valid fixes for me.

 - OS: macOS
 - Version 14.2.1

gajosadrian:
fixes #2609

It's more like a new feature because @schnetzi extended base **Role** model and did:
```php
protected $casts = [
    'name' => UserRoleEnum::class,
];
```

I used a custom accesor to dynamically cast it into the corresponding enum. Otherwise, it returns the original `name` attribute value.
```php
/**
 * @return string|\BackedEnum
 */
public function getNameAttribute()
{
    $name = $this->attributes['name'];

    if (str_contains($name, 'casted_enum')) {
        return TestRolePermissionsEnum::from($name);
    }

    return $name;
}
```

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -228,6 +228,16 @@ trait HasRoles
 
         if ($roles instanceof \BackedEnum) {
             $roles = $roles->value;
+
+            return $this->roles
+                ->when($guard, fn ($q) => $q->where('guard_name', $guard))
+                ->contains(function ($role) use ($roles) {
+                    if ($role->name instanceof \BackedEnum) {
+                        return $role->name->value == $roles;
+                    }
+
+                    return $role->name == $roles;
+                });
         }
 
         if (is_int($roles) || PermissionRegistrar::isUid($roles)) {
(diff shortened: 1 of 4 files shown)


gajosadrian:
Probably need to add tests with string parameters too.

drbyte:
Thanks!

PR #2618, opened 2024-02-19
Get permission and role class from method instead of the configuration in traits

illambo:
**Edit**: I push PR #2623 for a more complete solution.

This PR add the (little) ability to override role and permission model class in traits, this can "facilitate" (in a non-exhaustive manner, see discuss #2571) some cases of applications with package used on multiple databases (without the need to modify the configuration at runtime).

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -58,7 +58,7 @@ class Permission extends Model implements PermissionContract
     public function roles(): BelongsToMany
     {
         return $this->belongsToMany(
-            config('permission.models.role'),
+            $this->getRoleClass(),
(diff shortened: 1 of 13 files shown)


parallels999:
Must be the same change for Models??
https://github.com/spatie/laravel-permission/blob/0e593bb9e6808cc55c67f150878b0847f20b9036/src/Models/Permission.php#L61
https://github.com/spatie/laravel-permission/blob/0e593bb9e6808cc55c67f150878b0847f20b9036/src/Models/Role.php#L70

illambo:
Good point, later I check this. Thanks

drbyte:
still reviewing this and the other PR

PR #2641, opened 2024-03-21
Update HasPermissions::collectPermissions() docblock

Plytas:
Implementation was changed in https://github.com/spatie/laravel-permission/pull/2420 so the comment is currently incorrect.


drbyte:
Thanks!

PR #2658, opened 2024-04-19
Fix can't save the same model twice

### LINKED ISSUE #2638: Cannot save the user model twice in the same request
**Describe the bug**
Hi,
Sorry if this has already been asked but I'm using Laravel fortify to register new users, once user is registered I automatically log him in but I have a listener that listens for Login and automatically edits the user IP and last login and saves the user again. When the user is saved again I get an error Duplicate entry '7-153-User' for key 'model_has_roles.PRIMARY'. 
I fixed the issue by saving quietly but is this expected behavior ?
Now everytime I save a user in the same lifetime cycle it just crashes.
It seems to come from the use of attach instead of sync in the HasRoles trait line 168. see #2420 
Did not have this issue before because I migrated from v4 to v6.

**Versions**
- spatie/laravel-permission package version: 6.4.0
- laravel/framework package: 10.48.3

PHP version: 8.2.16

Database version: mysql 8.0.32

**To Reproduce**

- Create a new User
- Sync or Assign role to the user
- Save the User
- Save the User again

My listener:
```php
public function login(Login $event): void
    {
        try {
            /** @var User $user */
            $user = $event->user;
            activity('Auth')->by($user)
                ->withProperty('IP', request()->getClientIp())
                ->log('Login');

            $user->last_login = Carbon::now();
            $user->last_ip = request()->getClientIp();
            $user->saveQuietly();
        } catch (\Exception $exception) {
            report($exception);
        }
    }
```

**Expected behavior**
I would expect the save to not throw an error if I save the User twice.

 - OS: ubuntu
 - Version: 22.04

erikn69:
- Closes #2638

~This doesn't solve the problem completely~ **Another approach seems to work**
 
```php
$user->save();
//$user = $user->fresh();
$user->save();
```

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -401,14 +401,16 @@ trait HasPermissions
             $model->unsetRelation('permissions');
         } else {
             $class = \get_class($model);
+            $saved = false;
 
             $class::saved(
-                function ($object) use ($permissions, $model, $teamPivot) {
-                    if ($model->getKey() != $object->getKey()) {
+                function ($object) use ($permissions, $model, $teamPivot, &$saved) {
+                    if ($saved || $model->getKey() != $object->getKey()) {
                         return;
                     }
                     $model->permissions()->attach($permissions, $teamPivot);
                     $model->unsetRelation('permissions');
+                    $saved = true;
                 }
             );
         }
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -159,14 +159,16 @@ trait HasRoles
             $model->unsetRelation('roles');
         } else {
             $class = \get_class($model);
+            $saved = false;
 
             $class::saved(
-                function ($object) use ($roles, $model, $teamPivot) {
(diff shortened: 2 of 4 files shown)


drbyte:
Right. I'm wondering the same thing.

drbyte:
Would this be safe instead?
```diff
-    $model->roles()->attach($roles, $teamPivot);
+    $currentRoles = $model->roles->map(fn ($role) => $role->getKey())->toArray();
+    $model->roles()->attach(array_diff($roles, $currentRoles), $teamPivot);
```

erikn69:
>set a flag

Seems, to work
I don't know if Laravel has a way to remove the callback after a single execution or if you can locate and remove the callback.
```php
$class = \get_class($model);
$saved = false;

$class::saved(
    function ($object) use ($permissions, $model, $teamPivot, &$saved) {
        if ($saved || $model->getKey() != $object->getKey()) {
            return;
        }
        $model->permissions()->attach($permissions, $teamPivot);
        $model->unsetRelation('permissions');
        $saved = true;
    }
);
```

PR #2676, opened 2024-06-02
[v7] Update Role contract to match implementation and fix PhpStan errors

SanderMuller:
Allow BackedEnum and set default guardName

Fixes the following PhpStan error:
```
 ------ -------------------------------------------------------------------- 
  Line   app/Console/Commands/EnsureRolesAndPermissionsAreSetCommand.php     
 ------ -------------------------------------------------------------------- 
  29     Method Spatie\Permission\Contracts\Role::hasPermissionTo() invoked  
         with 1 parameter, 2 required.                                       
         ✏️  app/Console/Commands/EnsureRolesAndPermissionsAreSetCommand.php  
  29     Parameter #1 $permission of method                                  
         Spatie\Permission\Contracts\Role::hasPermissionTo() expects         
         Spatie\Permission\Contracts\Permission|string,                      
         App\Enums\Permissions\PermissionEnum::LOGIN_AS_ANY given.           
         ✏️  app/Console/Commands/EnsureRolesAndPermissionsAreSetCommand.php  
 ------ -------------------------------------------------------------------- 
```

caused by this line in my application:
`$role->hasPermissionTo(PermissionEnum::LOGIN_AS_ANY)`

I am passing an enum instance which is allowed according to the docs, and works, but the interface Docblock does not support this. The other error says I am required to pass two parameters, while the second param (`$guardName`) is optional according to the implementations, but not according to the interface

### DIFF
--- a/src/Contracts/Role.php
+++ b/src/Contracts/Role.php
@@ -24,7 +24,7 @@ interface Role
      *
      * @throws \Spatie\Permission\Exceptions\RoleDoesNotExist
      */
-    public static function findByName(string $name, ?string $guardName): self;
+    public static function findByName(string $name, ?string $guardName = null): self;
 
     /**
      * Find a role by its id and guard name.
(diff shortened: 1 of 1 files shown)


drbyte:
I agree with your point.
It may be a breaking change for many though. 
So may have to hold off on this for awhile.

SanderMuller:
> I agree with your point. It may be a breaking change for many though. So may have to hold off on this for awhile.

Maybe we can just update the dockblocks as a patch and then the actual types in a major version?

drbyte:
Yes, that'd be fine.

drbyte:
Added `BackedEnum` in https://github.com/spatie/laravel-permission/commit/59b966fbb201c8dfc97dda31547c90a175f75b02

PR #2694, opened 2024-07-16
Check for 'all' or 'any' permissions before specific permissions

ceilidhboy:
Shouldn't the check for `edit all posts` or `delete any post` be done first, before checking if a user can edit or delete their own posts?

The original code checked if the user can edit their own posts and, if so, would return false if they were not the post author, **even though they have the permission to edit any post**.

By performing the `all`/`any` check first, these permissions still work correctly when the user also has permissions to edit or delete their own posts.

### DIFF
--- a/docs/best-practices/using-policies.md
+++ b/docs/best-practices/using-policies.md
@@ -49,23 +49,23 @@ class PostPolicy
 
     public function update(User $user, Post $post)
     {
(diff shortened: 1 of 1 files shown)


drbyte:
Yes. Thanks.
Granted, someone might prefer the other way, but of course they can write their policy however they choose!

PR #2702, opened 2024-08-10
Do not show as error when no cache exists

francoism90:
I refresh the permission cache on each application deploy.

However this sometimes results in a error, e.g. when the cache hasn't been created yet.

I don't think the check is needed here. :)


drbyte:
Thank you for the contribution. It's good, yet decided to use the other implementation.

PR #2706, opened 2024-08-22
Update super-admin.md

Tamas-hi:
As a first time reader of the package documentation (which by the way is really good!), I realized that in the seeding example we use 'super-admin'.  I haven't tried it yet, but I believe these two should be consistent to work, so I propose we change it here to super-admin as well.


drbyte:
Thank you for the suggestion.
I actually prefer the uppercase unhyphenated form.
Will update the seeder.

PR #2719, opened 2024-09-14
[Docs] use more modern syntax for nullable return type

galangaidilakbar:
This pull request updates the return type of the `before` method in the docs from `bool|null` to `?bool`. This change ensures that the method correctly reflects its possible return values, including true, false, and null.


drbyte:
Thanks.

Maybe my eyes are tired, but both mean the same, don't they?

PR #2733, opened 2024-10-10
Update middleware.md - Add L11 example

eugenefvdm:
Remove reference to Laravel 11 manual and add a concrete example.


drbyte:
Thanks for your contribution.

Unfortunately the example you posted is a duplication of the existing example on lines 25-35.

The line you replaced was addressing another matter: that of reordering middleware if one's app requires it.

PR #2742, opened 2024-10-22
4 events for adding and removing roles or permissions

sven-wegner:
First of all: This is my first pull request. If everything is not correct, please give me feedback!

**Why I created this pull request**

In my company we have an API and a UI. If changes have been made to permissions or roles (e.g. because an additional service has been booked), the UI wants to be informed about this. In my opinion, events are perfect for this!

Unfortunately, this package does not currently offer such events, as the `HasRoles::roles()` and `HasPermissions::permissions()` methods do not fire any. There is also no observer, as there is no `Model` class for the pivot table.

This pull request adds 4 events `PermissionAttached`, `PermissionDetached`, `RoleAttached` and `RoleDetached` for adding and removing roles or permission.

You can then have a listener listen to these in the main instance or work directly with sockets.

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -111,6 +111,17 @@ return [
     'register_octane_reset_listener' => false,
 
     /*
+     * Events will fire when a role or permission is assigned/unassigned:
+     * \Spatie\Permission\Events\RoleAttached
+     * \Spatie\Permission\Events\RoleDetached
+     * \Spatie\Permission\Events\PermissionAttached
+     * \Spatie\Permission\Events\PermissionDetached
+     *
+     * To enable, set to true, and then create listeners to watch these events.
+     */
+    'events_enabled' => false,
+
+    /*
      * Teams Feature.
      * When set to true the package implements teams using the 'team_foreign_key'.
      * If you want the migrations to register the 'team_foreign_key', you must
--- a/src/Events/PermissionAttached.php
+++ b/src/Events/PermissionAttached.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Spatie\Permission\Events;
+
+use Illuminate\Broadcasting\InteractsWithSockets;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Foundation\Bus\Dispatchable;
+use Illuminate\Queue\SerializesModels;
+
+class PermissionAttached
+{
+    use Dispatchable;
+    use InteractsWithSockets;
+    use SerializesModels;
+
+    public function __construct(public Model $model)
+    {}
+}
--- a/src/Events/PermissionAttached.php
+++ b/src/Events/PermissionAttached.php
@@ -8,6 +8,8 @@ use Illuminate\Broadcasting\InteractsWithSockets;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Queue\SerializesModels;
+use Illuminate\Support\Collection;
+use Spatie\Permission\Contracts\Permission;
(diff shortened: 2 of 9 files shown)


parallels999:
Wouldn't it be nice to also report in the event what roles/permissions were attached/detached?

votkapower:
This will work only when giving programaticly roles/permissions, but when used in tool like Laravel Nova or some other tool where a record is added to the pivot table, events will not be fired. 

I used a approach of PivotTable model and emit event against its created & deleted hooks, which doesnt care how you add or sync the permissions, its looks the pivot events and then fires evets.


I'm writing this so others can try it if they need it:

Overide the permisison relation on  the HasPermissions Trait:
```php
 public function permissions(): BelongsToMany
    {
      // ... 
        return $relation = $this->morphToMany(
            config('permission.models.permission'),
            'model',
            config('permission.table_names.model_has_permissions'),
            config('permission.column_names.model_morph_key'),
            PermissionRegistrar::$pivotPermission
        )->using(\App\Models\Pivots\ModelHasPermissions::class); // Use custom pivot class here
    }
```

`ModelHasPermissions` Extends the Pivot table :

```php
use Illuminate\Database\Eloquent\Relations\MorphPivot;
class ModelHasPermissions extends MorphPivot
{
 // ..
}

```

And create observer `ModelHasPermissionsObserver`:

```php
class ModelHasPermissionsObserver
{
    public function created(ModelHasPermissions $pivot)
    { 
        event(new UserPermissionAttached($pivot->model, $pivot->permission));
    }

    public function deleted(ModelHasPermissions $pivot)
    { 
        event(new PermissionRemoved($model, $pivot->permission));
    }
}
```

Observe for changes in AppServiceProvider:

```php
 ModelHasPermissions::observe(ModelHasPermissionsObserver::class);
```

You can do the same for the roles.
Maybe you can also incorporate this in this PR, so we have both?
 I hope it helps.

drbyte:
> Maybe you can also incorporate this in this PR, so we have both?

Wouldn't `both` be redundant, and result in firing events twice?

drbyte:
Looking further, `MorphbyMany(...)->using(...)` allows `->using(null)` to be passed for default operation.
So, we could probably introduce a config like `permission.models.pivots.model_has_permissions` which defaults to null, but could let someone declare a specific pivot model.

The downside to specifying a pivot model is that it introduces an extra db query. But the upside is that you can register observers to it.

drbyte:
> Looking further, `MorphbyMany(...)->using(...)` allows `->using(null)` to be passed for default operation. So, we could probably introduce a config like `permission.models.pivots.model_has_permissions` which defaults to null, but could let someone declare a specific pivot model.
> 
> The downside to specifying a pivot model is that it introduces an extra db query. But the upside is that you can register observers to it.

@sven-wegner I'd like to avoid directly firing events in the traits, and instead use an Observer.

eg:
- `config('permission.models.pivots.model_has_permissions')` defaults to null
- But we provide a Pivot model that could be specified
- And we provide an observer that could be registered in the user's own app, to watch that Pivot model.

That way if one's app needs the events, they can specify the pivot model and register the observer, and then respond to the events as desired.  And if their app doesn't need these events, they can avoid the extra DB Query that the pivot introduces.

And, if they wanted to Observe against fewer events, they could extend and then override the provided Observer, adjusting whatever they need.

drbyte:
I've updated this PR to just use events directly from the trait, and be able to configure whether they fire or not.

The Observer approach can be added via a separate PR. I've staged working code for it, but haven't figured out the tests for it, so don't want to clutter this PR with those fragments. Easier to do separately.

sven-wegner:
@drbyte thank you!

PR #2744, opened 2024-10-23
Allow LDAP model lookup from Auth provider

crossplatformconsulting:
allows for ldap authentication where the model differs from the database model - ldap still uses the App\Models\User, just indirectly.

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -39,6 +39,25 @@ class Guard
     }
 
     /**
+     * Get the model class associated with a given provider.
+     *
+     * @param  string  $provider
+     * @return string|null
+     */
+    protected static function getProviderModel(string $provider): ?string
+    {
+        // Get the provider configuration
+        $providerConfig = config("auth.providers.{$provider}");
+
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks.

It seems we're doing similar lookups in a couple of the `Guard` class methods as well. 
I wonder if we should take this logic there, and simply have this function call it?

crossplatformconsulting:
I see.. I could split out the helper into a new function and use that in the guard, as well as the in the helper function I just modified?

By: Jason: Page
All Rights Reserved Without Prejudice

drbyte:
I was thinking the other way around: Update the `Guard` class (can an existing method be used here?), and then the helper can call the method in the Guard class, kinda like a proxy.

crossplatformconsulting:
Yeah.. I saw that too after I started looking more into it..

Let me dig a little more 😊

By: Jason: Page
All Rights Reserved Without Prejudice

crossplatformconsulting:
better?

PR #2748, opened 2024-10-30
Fix GuardDoesNotMatch should accept collection

### LINKED ISSUE #2745: GuardDoesNotMatch should accept collection but string is passed in Role::hasPermissionTo method
In https://github.com/spatie/laravel-permission/blob/main/src/Models/Role.php#L188

it passes $guardName and it is a nullable string whereas GuardDoesNotMatch exception accepts Collection as a second argument.

erikn69:
- Closes #2745


drbyte:
Thanks!

PR #2749, opened 2024-10-31
Improve performance for hydrated collections

inserve-paul:
After profiling our application it appeared that a lot of time of a request was wasted in the PermissionRegistrar. When creating a collection of Permission objects the array_map method is used to create these objects from the items in cache. However everytime the newInstance method is called on the permission instance.

In this PR I've replaced this call by using the native clone method, this performs about 3 times faster!

For e.g. in our application we have about 400 permissions; which takes +/- 18ms to create the hydrated permission collection. By using the native clone method it only takes 6ms. (Executed on a dev environment, without xDebug and xhProf enabled)


drbyte:
Thanks!

PR #2750, opened 2024-10-31
LDAP model lookup from Auth Provider

crossplatformconsulting:
include ldap check to get the correct model - when using ldap, check for the database model.
(ldap still uses the App\Models\User, just indirectly)

Replaces #2744

### DIFF
--- a/src/Guard.php
+++ b/src/Guard.php
@@ -74,7 +74,6 @@ class Guard
                     return null;
                 }
 
-                // Fetch the model for this guard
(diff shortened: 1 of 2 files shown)


crossplatformconsulting:
apologies.. i accidentally deleted my fork before the PR was completed. re-initiated.

drbyte:
Thanks

lantis1008:
Just ran into this integrating laravel-permission and ldaprecord. Thanks for the fix!

PR #2753, opened 2024-11-04
Fix typo

machacekmartin:
Fixed a very serious issue ❤️


parallels999:
It seems to me that that entire block should have been removed in this version

drbyte:
> Fixed a very serious issue ❤️

Yes. Thanks for helping fix this `major` issue!

PR #1912, opened 2021-11-04
[V5] Cache loader improvements

erikn69:
- Deleted old V4 fallbacks
- Early return used
- Avoid use same bucle twice on permisions (`collect($this->permissions)->map()` and later `->each()`)
- `array_map` instead of `collect($this->permissions)->map()->all()` and other `collect()` again with the first collect

PR #2754, opened 2024-11-04
Remove v5 cache fallback

drbyte:
Ref #1912

/cc @erikn69


erikn69:
Yes, at this point it should be safe to remove that code 👍

PR #2759, opened 2024-11-08
Fix: #2749 bug "Can no longer delete permissions"

### LINKED ISSUE #2757: permission->assignRole after Permission::findOrCreate is not correctly assigning the roles
### Description

When I run the database seeder for roles and permissions, permissons' ``assignRole`` is not correctly adding new roles for the same permission that is queried from ``Permission::findOrCreate``. 

With the example application I provided below, only the first role is assigned.

I noticed this in version 6.10. When I revert to version 6.9, the roles are assigned as expected.

### Steps To Reproduce

1. Create at least two roles (``$role1`` and ``$role2``);
2. Create the permission using ``Permission::findOrCreate('Permission 1')``; 
3. Assign the first role to the permission using ``$permission->assignRole($role1)``;
4. Get the created permission using ``Permission::findOrCreate('Permission 1')``; 
5. Then assign the second role to the same permission also using ``$permission->assignRole($role2)``

### Example Application

```
$role1 = Role::create(['name' => 'Role 1']);
$role2 = Role::create(['name' => 'Role 2']);

$permission = Permission::findOrCreate('Permission 1');
$permission->assignRole($role1);

$permission = Permission::findOrCreate('Permission 1');
$permission->assignRole($role2);
```

6.10

11

### PHP version:

8.2

### Database engine and version:

SQLite

### OS: Windows/Mac/Linux version:

Windows 11 and MacOS 15

### LINKED ISSUE #2758: Can no longer delete permissions returned by "findByName"
### Description

Before when doing:
$permission = Permission::findByName($name, $guardName);

And then calling:
$permission->delete()

The permission would be correct deleted.

Now this no longer happens because the "$exists" property on the permission model being false instead of true.

This behavior has changed since version 6.10.0, worked fine in 6.9.0.

### Steps To Reproduce

1. $permission = Permission::findByName($name, $guardName);
2. $permission->delete()
3. Notice the permission is not deleted.

### Example Application

_No response_

6.10.0

11.30.0

### PHP version:

8.3

### Database engine and version:

singlestore

### OS: Windows/Mac/Linux version:

Ubuntu

erikn69:
[Illuminate/Database/Eloquent/Model.php#L615-L619](https://github.com/laravel/framework/blob/37f96f1d6b515f55ea7dfef144fd41519cca38b6/src/Illuminate/Database/Eloquent/Model.php#L615-L619)
```php
public function newInstance($attributes = [], $exists = false)
{
// This method just provides a convenient way for us to generate fresh model
// instances of this current model. It is particularly useful during the
// hydration of new objects via the Eloquent query builder instances.
```

- `newInstance` was removed on #2749
- Closes #2758

### DIFF
--- a/src/PermissionRegistrar.php
+++ b/src/PermissionRegistrar.php
@@ -349,7 +349,7 @@ class PermissionRegistrar
 
     private function getHydratedPermissionCollection(): Collection
     {
-        $permissionInstance = new ($this->getPermissionClass())();
+        $permissionInstance = (new ($this->getPermissionClass())())->newInstance([], true);
 
         return Collection::make(array_map(
             fn ($item) => (clone $permissionInstance)
@@ -368,7 +368,7 @@ class PermissionRegistrar
 
     private function hydrateRolesCache(): void
     {
-        $roleInstance = new ($this->getRoleClass())();
+        $roleInstance = (new ($this->getRoleClass())())->newInstance([], true);
 
         array_map(function ($item) use ($roleInstance) {
             $role = (clone $roleInstance)
--- a/tests/PermissionTest.php
+++ b/tests/PermissionTest.php
@@ -67,4 +67,15 @@ class PermissionTest extends TestCase
 
         $this->assertEquals($this->testUserPermission->id, $permission_by_id->id);
     }
+
+    /** @test */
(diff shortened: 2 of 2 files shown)


drbyte:
Thanks!

PR #2776, opened 2024-12-11
Explicitly call `loadMissing('permissions')` when the relation is needed, and test with `Model::preventLazyLoading()`

erikn69:
@drbyte hi,
I did add `Model::preventLazyLoading()` to the tests, but it doesn't give me any errors, did I put it in the wrong place(`getEnvironmentSetUp`)?
It would be a good idea to leave it, so that mistake is not made unintentionally.

The other changes are PHPSTAN bugs, they came out when updating

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -10,9 +10,11 @@ jobs:
       fail-fast: false
       matrix:
         php: [8.4, 8.3, 8.2, 8.1, 8.0]
-        laravel: ["^11.0", "^10.0", "^9.0", "^8.12"]
+        laravel: ["^12.0", "^11.0", "^10.0", "^9.0", "^8.12"]
         dependency-version: [prefer-lowest, prefer-stable]
         include:
+          - laravel: "^12.0"
+            testbench: 10.*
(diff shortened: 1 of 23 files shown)


drbyte:
> did I put it in the wrong place(getEnvironmentSetUp)?

Looks like that's a suitable place. So is earlier, like `SetUp()`. But I like it in `getEnvironmentSetUp()`, as you've done.

Confirmation example test, run in someplace like `PermissionTest.php`:
```php
    public function global_prevents_lazy_loading_flag_is_enabled()
    {
        $this->assertTrue(Model::preventsLazyLoading());
    }
```

drbyte:
I think there's value in increasing the size of the testing dataset to be sure we're not missing something by virtue of the fact that Lazy Loading isn't even being required.
This is probably why some of the past issues with loading relations have caused unexpected outcomes. And BC frustrations.

drbyte:
Thanks.

Expanding the tests dataset can be another task/project/PR. The idea being to ensure that the lazy-loading criteria isn't skipped/bypassed due to Laravel determining that N+1 wouldn't happen since too few records are present.

PR #2780, opened 2024-12-26
Replace php-cs-fixer with Laravel Pint

bobbrodie:
## Overview

I've started working with this package, and wanted to take a deeper look at how it's built, so I was reviewing the source. One small thing I came upon was that there was a Composer script that would call `php-cs-fixer fix --allow-risky=yes`. There were two issues:

1. The package did not include the friendsofphp/php-cs-fixer development dependency, so there was an expectation that it would be installed globally
2. No path was set on the format command

I originally made a PR to address this, but then I realized that the GitHub Actions are set up to use Pint, so I made a change that I think could make contributing easier in the future.

### Add dependency

I added the the `laravel/pint` dependency:

```json
"laravel/pint": "^1.0"
```

> NOTE: I constrained to version `^1.0` so that it will install a version old enough to support PHP 8.0 to match this package. Installing on PHP >= 8.1 will install the latest version, `v1.18.3` as of submitting this.

### Configure Laravel Pint

By default, Pint enforces snake_case method names, which will change:

```php
it_throws_an_exception_when_calling_hasPermissionTo_with_an_invalid_type()
```
to
```php
it_throws_an_exception_when_calling_has_permission_to_with_an_invalid_type()
```

The actual name of the method being tested is `hasPermissionTo`, so I added a `pint.json` with the following:

```json
{
    "preset": "laravel",
    "rules": {
        "php_unit_method_casing": false
    }
}
```

### Update Composer script

The `format` script is now defined as:

```json
"format": "pint",
```

### Run Laravel Pint

Finally, I ran `composer format` and let it run through the project.

## Results

The resulting changes are minimal, and all tests are passing. Running `composer format` should now allow contributors to format their code prior to committing and have the same results as the GitHub Action.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "require-dev": {
         "larastan/larastan": "^1.0|^2.0",
         "laravel/passport": "^11.0|^12.0",
-        "laravel/pint": "^1.18",
+        "laravel/pint": "^1.0",
         "orchestra/testbench": "^6.23|^7.0|^8.0|^9.0",
         "phpunit/phpunit": "^9.4|^10.1"
     },
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
     "require-dev": {
         "larastan/larastan": "^1.0|^2.0",
         "laravel/passport": "^11.0|^12.0",
+        "laravel/pint": "^1.18",
         "orchestra/testbench": "^6.23|^7.0|^8.0|^9.0",
         "phpunit/phpunit": "^9.4|^10.1"
     },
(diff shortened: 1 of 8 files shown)


drbyte:
Thanks!

bobbrodie:
Oh hey that was quick! Sure thing -- thanks for merging. I know it's small but it was a nice way to settle in :)

parallels999:
https://github.com/spatie/laravel-permission/blob/main/.github/workflows/fix-php-code-style-issues.yml

Already pint auto commit action

PR #2785, opened 2025-01-05
[Docs] Correct username in new-app.md

trippodi:
The user created is test**er**@example.com; not test@example.com


drbyte:
Thanks!

PR #2786, opened 2025-01-06
Update installation-laravel.md

fcontreras:
Updated the instructions #4 for services providers registration on laravel 11 new projects


drbyte:
I don't understand why you deleted point `#5` and renumbered everything afterward.

fcontreras:
Yeah that was incorrect, I'll send another pull request later. Thanks for checking it, and apologize for the bad commit diff.

PR #2790, opened 2025-01-19
Add configurable team resolver for permission team id

adrenallen:
This adds a configurable team resolver to the permission registrar class.

This allows users to implement their own storage for the team id. This is useful if you are setting a `current_team_id` on the user table [similar to how jetstream works.](https://jetstream.laravel.com/features/teams.html#the-current-team).

By implementing a resolver in this way, the middleware approach is optional as the resolver will always go through your logic via CLI, HTTP, or otherwise. Which means less explicit calls to `setPermissionsTeamId`.

We are defaulting to the `DefaultTeamResolver` so that we are backwards compatible with anyone that already has the config published.

Let me know if more tests are desired, I felt that the current tests did a good job covering the setting and getting of team permission id.

### Example of a custom resolver:

`App\Permissions\CustomerTeamResolver.php`
```php
class CustomTeamResolver implements PermissionsTeamResolver
{
    public function setPermissionsTeamId($id): void
    {
        if ($id instanceof \Illuminate\Database\Eloquent\Model) {
            $id = $id->getKey();
        }

        if (auth()->user()) {
            auth()->user()->current_team_id= $id;
            auth()->user()->save();
        }
        
    }

    public function getPermissionsTeamId() : int|string|null
    {
        return auth()->user()?->current_team_id;
    }
}
```

`config/permission.php`
```php
...
'team_resolver' => \App\Permissions\CustomTeamResolver::class,
...
```


adrenallen:
For context, this is how I accomplished this same functionality prior to this PR:

`AppServiceProvider.php`
```php
class AppServiceProvider extends ServiceProvider
{
    /**
     * Register any application services.
     */
    public function register(): void
    {
        $this->app->beforeResolving(\Spatie\Permission\PermissionRegistrar::class, function () {
            $this->app->singleton(\Spatie\Permission\PermissionRegistrar::class, PermissionRegistrar::class);
        });
    }
```
    
`PermissionRegistrar.php`
```php
<?php

namespace App\Providers;

class PermissionRegistrar extends \Spatie\Permission\PermissionRegistrar
{
    /**
     * Set the team id for teams/groups support, this id is used when querying permissions/roles
     *
     * @param  int|string|\Illuminate\Database\Eloquent\Model|null  $id
     */
    public function setPermissionsTeamId($id): void
    {
        auth()->user()->current_organization_id = $id;
        auth()->user()->save();
    }

    /**
     * @return int|string|null
     */
    public function getPermissionsTeamId()
    {
        return current_organization_id();
    }
}

```

sorrell:
Bump - extremely helpful for jetstream

PR #2796, opened 2025-01-29
Update installation-laravel.md to fix providers.php location.

curiousteam:
4. Optional: The **`Spatie\Permission\PermissionServiceProvider::class`** service provider will automatically get registered. Or you may manually add the service provider to the array in your `bootstrap/providers.php` (or `config/app.php` in Laravel 10 or older) file.


drbyte:
Thanks!

PR #2803, opened 2025-02-04
Fixed bug of loading user roles of different teams to current team

### LINKED ISSUE #2802: Bug when assigning role to user that has same role in a different team
### Bug Description :
This happens when the user has role for ex: `admin` in team `A` and want to assign it same role but in team `B`.
What happens is that the user doesnt get assigned that role `admin` in that team `B` cuz in the treat that the method `assignRole()` sits in, the the `$this->roles` relation property is called to get user's roles and run `array_diff` on those currentRoles and the new Roles to assign (in this case **admin** role), the problem is that if the roles are already loaded on the user object before the teamsId was changed, then the `$this->roles` would return the role `admin` cuz the user has it in the other team not this one, thus the array_diff would return an empty array thus no roles would be assigned, 

#### HasRoles trait, assignRole method :

https://github.com/spatie/laravel-permission/blob/9d4fa2987f9b745854a13e1b36ba68e0b6dd4bdb/src/Traits/HasRoles.php#L154-L159

### Fix Suggestion :
We should instead of calling `$this->roles` which could load the already loaded roles, we should use `$this->roles()` which reload the roles with taking into consideration the new team_id that's set on the session thus no roles interference would happen.

mohamedds-12:
Fixes : 
- #2802

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -151,11 +151,12 @@ trait HasRoles
         $teamPivot = app(PermissionRegistrar::class)->teams && ! is_a($this, Permission::class) ?
             [app(PermissionRegistrar::class)->teamsKey => getPermissionsTeamId()] : [];
 
-        if (app(PermissionRegistrar::class)->teams) {
-            $this->load('roles');
-        }
-
         if ($model->exists) {
+            if (app(PermissionRegistrar::class)->teams) {
+                // explicit reload in case team has been changed since last load
+                $this->load('roles');
+            }
+
             $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();
 
             $this->roles()->attach(array_diff($roles, $currentRoles), $teamPivot);
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -152,7 +152,7 @@ trait HasRoles
             [app(PermissionRegistrar::class)->teamsKey => getPermissionsTeamId()] : [];
 
         if ($model->exists) {
-            $currentRoles = $this->roles->map(fn ($role) => $role->getKey())->toArray();
+            $currentRoles = $this->roles()->get()->map(fn ($role) => $role->getKey())->toArray();
 
             $this->roles()->attach(array_diff($roles, $currentRoles), $teamPivot);
             $model->unsetRelation('roles');
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -151,8 +151,12 @@ trait HasRoles
         $teamPivot = app(PermissionRegistrar::class)->teams && ! is_a($this, Permission::class) ?
             [app(PermissionRegistrar::class)->teamsKey => getPermissionsTeamId()] : [];
 
+        if (app(PermissionRegistrar::class)->teams) {
+            $this->load('roles');
+        }
+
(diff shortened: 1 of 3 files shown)


parallels999:
You are adding extra queries, 
That was developed to take advantage of the preloaded relationships when teams are not enabled, this would affect the performance of all of us who do not use teams

There are many notes in the documentation about using `unsetRelation` in teams, changing teams in the same session is not something that should happen, it is designed to maintain a single team

cesarreyes3:
You could just do `attach` directly
```php
$user->roles()->attach($roleIdsArray, $theTeamId);
```

mohamedds-12:
I have changed the fix to **Reloading user roles before role assigning  `only if teams feature is enabled`** to avoid running extra queries when not using teams, And I have changed the Tests accordingly.

I think it's a logical thing to expect that the package would take into consideration teamSessionId  getting  changed on the fly as the functionality is supported therefore it  should assert to reload  user  roles on each role-related action executed (of course only when teams are enabled) as  for there  is no reason for reloading when not using teams.

I think we should  add this functionality as it is not a breaking change and it's been mentioned earlier
 and it's an excepted behavior
- #1094 
- #2553

drbyte:
I still think this needs some specific tests regarding the problem that's being addressed.

It's probable that the `it_can_assign_same_and_different_roles_on_same_user_different_teams` test doesn't cover this scenario, so another test that does cover the present problem would be beneficial.

mohamedds-12:
> I still think this needs some specific tests regarding the problem that's being addressed.
> 
> It's probable that the `it_can_assign_same_and_different_roles_on_same_user_different_teams` test doesn't cover this scenario, so another test that does cover the present problem would be beneficial.

@drbyte Just updated that test to cover this scenario.

drbyte:
No, what you did was force the test to always pass, which means that the test isn't actually testing the problem.
I meant (adding) a test that would replicate all the steps of the problem you presented, and where the test would fail if the reloading of the relationship was not present.

mohamedds-12:
> No, what you did was force the test to always pass, which means that the test isn't actually testing the problem. I meant (adding) a test that would replicate all the steps of the problem you presented, and where the test would fail if the reloading of the relationship was not present.

@drbyte 
The problem that I presented was being caused when 

1. user's roles property is loaded and has `testRole` for ex.
2. Then team was changed.
3. Then assigned that `testRole` to the user in this new team.

After this the role wouldn't get assigned cuz the `assignRole` considers the role is already assigned to the user cuz it preserves that role that's pre-loaded from the old team.

And this case exactly was handled in the test method `it_can_assign_same_and_different_roles_on_same_user_different_teams ` but what was missing is that the 'roles' relation wasn't pre-loaded to the user, that's why I added loading it to assert that the test doesn't fail even if the relation was pre-loaded before switching teams :
https://github.com/spatie/laravel-permission/blob/ca9acc149fd8e998fb2c9a1cb7638f6509420a74/tests/TeamHasRolesTest.php#L52-L70


You can try and add loading the `roles` relation in that test in the main branch out of this PR's branch then run the test and see, 
it fails when the relation is pre-loaded before team switch, do the same in this branch and it'll pass cuz of the change that was added to the `HasRoles` trait.

drbyte:
Thanks.

mohamedds-12:
> Thanks.

Any time 🙌

PR #2813, opened 2025-02-26
Middleware: support enums in role/permission middleware

marklawntalk:


drbyte:
I think we should also test the `handle()` method of both middleware, as well. Which means also implementing the enum check there.
I've added a potential test (could be refined) as a starting point.

marklawntalk:
Thanks! I've updated the code to handle the handle method.

tests/PermissionMiddlewareTest.php
```diff
@@ -467,6 +467,9 @@ public function the_middleware_can_handle_enum_based_permissions_with_static_usi
     #[Test]
     public function the_middleware_can_handle_enum_based_permissions_with_handle_method()
     {
+        app(Permission::class)->create(['name' => TestModels\TestRolePermissionsEnum::VIEWARTICLES]);
+        app(Permission::class)->create(['name' => TestModels\TestRolePermissionsEnum::EDITARTICLES]);
```
drbyte, line 471: Did these really need to be created for this test? Why not also in the test above?

drbyte (changes requested):
Should also add the tests for the `handle` method on Roles.

tests/PermissionMiddlewareTest.php
marklawntalk: The create is needed in this test because the `handle` method performs db check for string permission, while the using method only returns the parsed string permission.

marklawntalk: I believe the error is due to the model's create method handling the enum, which isn't part of this PR. Should I change the enum to a string in the creation process? I didn't encounter this error in my local tests.

drbyte: Presently, yes. Probably easiest just to add `->value`

marklawntalk: yep, done.

drbyte:
Thanks

Issue #2816, opened 2025-03-04
Refactor exception throwing in migration file to use throw_if


alisalehi1380:
### Description

https://github.com/spatie/laravel-permission/blob/02687346fe01242038a5ef9346696a42eda81250/database/migrations/create_permission_tables.php.stub#L20-L25

Can I refactor to:
```php
throw_if(empty($tableNames), new Exception('Error: config/permission.php not loaded. Run [php artisan config:clear] and try again.'));
throw_if($teams && empty($columnNames['team_foreign_key'] ?? null), new Exception('Error: team_foreign_key on config/permission.php not loaded. Run [php artisan config:clear] and try again.'));
```

### Steps To Reproduce

.

### Version of spatie/laravel-permission package:

6.10

### Version of laravel/framework package:

11

### PHP version:

8.2

PR #2819, opened 2025-03-07
refactor exception throwing in migration file to use throw_if()

ccaioadriano:


Danielopes7:
resolves #2816 ?

drbyte:
While this doesn't appear to solve any bug or problem, I'm not opposed to the refactor.

PR #2823, opened 2025-03-15
Route macro functions add backed enum support

Yi-pixel:
It might be simpler to use the following approaches, but they will break the previous tests (because the middleware name returned is the full class name, not the alias)

```php
\Spatie\Permission\Middleware\PermissionMiddleware::using();
\Spatie\Permission\Middleware\RoleMiddleware::using()
```


drbyte:
I like this. 

And your comment about using FQDN::using() is good too. Perhaps we can incorporate both? (Add them, vs changing prior ones)

drbyte:
Will merge this for now. The other adjustments for `using()` can be explored separately.

PR #2826, opened 2025-04-02
Added backed enum as possible role and permission name when finding or creating them

EranNL:
When finding or creating roles and permissions, you can't use an enum like you can when assigning roles or permissions.
This PR adds support for backed enums when you find or create them. 

Like so:
```php
$permission = Permission::findOrCreate(PermissionEnum::VIEW_USERS);
$role = Role::findOrCreate(RoleEnum::ADMIN);
```

### DIFF
--- a/src/Contracts/Permission.php
+++ b/src/Contracts/Permission.php
@@ -26,7 +26,7 @@ interface Permission
      *
      * @throws \Spatie\Permission\Exceptions\PermissionDoesNotExist
      */
-    public static function findByName(string $name, ?string $guardName): self;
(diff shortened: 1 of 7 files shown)


drbyte:
Hi. Still considering this for inclusion in next major release. A few things enum-related will be breaking changes, so queuing them for then. 
Re-opening.

drbyte:
Note to self, during future review: `BackedEnum` is not supported before PHP 8.1 ... which is why the PHP 8.0 tests are failing.
Hence another reason for queuing for a major release bump.

lloricode (approved):

PR #2834, opened 2025-04-21
Add assign-role command

sediqzada94:
# 🆕 Add Artisan Command to Assign Roles to Users

## 📌 Description

This PR adds a new Artisan command `permission:assign-role` that simplifies the process of assigning roles to users via the CLI.

---

## 🛠️ Command Signature

```bash
php artisan permission:assign-role {name} {userId} {guard?} {userModelNamespace=App\Models\User}

### DIFF
--- a/src/Commands/AssignRole.php
+++ b/src/Commands/AssignRole.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Spatie\Permission\Commands;
+
+use Illuminate\Console\Command;
+use Spatie\Permission\Contracts\Role as RoleContract;
+
+class AssignRole extends Command
+{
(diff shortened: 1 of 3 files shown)


parallels999:
Also not taking into account the "Teams feature"

https://github.com/spatie/laravel-permission/blob/a1851df15da5f9bcb7b47691ba03cd8f320fc4a0/src/Commands/CreateRole.php#L16

drbyte:
While I'm not a fan of adding (and maintaining) a lot of artisan commands for this package, I'll accept this one. There hasn't been much discussion/demand for it, so that means people are using the create-role command or using tinker, or migrations, etc.
A future iteration of this command could add team support as well.

PR #2837, opened 2025-04-23
Fix guard_name not used to set default attribute in Role and Permission model

Ken-vdE:
The Role and Permission model constructors don't use the $guard_name property to set $attributes['guard_name'] (through `Guard::getDefaultName()` as in all other methods).

Also, I believe the `getDefaultGuardName()` method and the `$guard_name` property should also be set in Role and Permission model, not just the User model as per the docs: https://spatie.be/docs/laravel-permission/v6/basic-usage/multiple-guards#content-forcing-use-of-a-single-guard

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -26,7 +26,7 @@ class Permission extends Model implements PermissionContract
 
     public function __construct(array $attributes = [])
     {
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks!

PR #2840, opened 2025-05-09
Remove `collectPermissions` that is not being assigned

JHWelch:
This is relatively straightforward. 

Because `syncPermissions` is calling `givePermissionTo` with the same passed permissions array directly, it does not need to call `collectPermissions`. 

In fact the call is not being assigned to any variable, so it is not really doing anything except wasting some time collecting the permissions, unless there is something I do not understand.

It does not seem to affect any tests so I think this should be fine!


drbyte:
Thanks!

PR #2844, opened 2025-05-13
Fix commented config key typo

### LINKED ISSUE #2843: Default config is wrong for custom WildcardPermission class
### Description

In the default `config/permission.php` file there is this section about overriding or extending the `Spatie\Permission\WildcardPermission` class with a custom one:

```
    /*
     * The class to use for interpreting wildcard permissions.
     * If you need to modify delimiters, override the class and specify its name here.
     */
    // 'permission.wildcard_permission' => Spatie\Permission\WildcardPermission::class,
```

However, the correct key for this should just be `"wildcard_permisson"` without the `permission.` in front of it. If you uncomment this line and change to something custom (e.g. `App\Models\CustomWildcardPermission::class`) as recommended in the comment, it won't work. That's because the `getWildcardClass` in the `HasPermissions.php` trait fille is looking for: 
```
$this->wildcardClass = config('permission.wildcard_permission', WildcardPermission::class);
``` 
Where `permission.wildcard_permission` refers to the permission config file with the top-level `wildcard_permission` key.

So, either the trait should be modified to look at the correct config value, or the config should be modified to reflect what the trait is looking for. 👍

### Steps To Reproduce

1. Make a custom WildcardPermission class that extends Spatie\Permission\WildcardPermission
2. Change the config/permission.php, uncommenting the line `permission.wildcard_permission`
3. Change the class name to your custom WildcardPermission class
4. Add `HasPermissions` trait to your user model
5. In a tinker session, run 
```
$u = User::first()
$u->getWildcardClass()   // outputs: "Spatie\Permission\WildcardPermission"
```

6.17

12.0

### PHP version:

8.4

erikn69:
Closes #2843

>the correct key for this should just be "wildcard_permisson" without the permission. in front of it. If you uncomment this line and change to something custom (e.g. App\Models\CustomWildcardPermission::class) as recommended in the comment, it won't work

### DIFF
--- a/config/permission.php
+++ b/config/permission.php
@@ -172,7 +172,7 @@ return [
      * The class to use for interpreting wildcard permissions.
      * If you need to modify delimiters, override the class and specify its name here.
      */
-    // 'permission.wildcard_permission' => Spatie\Permission\WildcardPermission::class,
+    // 'wildcard_permission' => Spatie\Permission\WildcardPermission::class,
 
     /* Cache-specific settings */
 


coreytukios (approved):

drbyte:
Thanks!

PR #2846, opened 2025-05-21
Add support for configurable role identifier key (e.g., slug)

philiplamb:
I needed to use a slug (explicitly not use "name" column), and to check against it, but this package forces you to use the name in your checkRole methods. I also saw many issues asking for slug support, this doesn't add slug so I made this configurable through the `config/permission.php`.

#1846 

You can now update this line to use your preferred column:
`'role_identifier' => 'name', // e.g., change to 'slug'`

This PR:

- Updates role matching methods (hasRole, hasAllRoles, hasExactRoles, etc.) to respect the configured identifier.
- Adds a getRoleKey() helper method to the HasRoles trait.
- Introduces findByKey() in the Role model to support lookup by the configured identifier.
- Tests name and slug usage. I had to add slug for testing as nullable.
- Has backward compatibility. The default behavior (name) remains unchanged.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -91,7 +91,17 @@ trait HasRoles
                 $role = $role->value;
             }
 
-            $method = is_int($role) || PermissionRegistrar::isUid($role) ? 'findById' : 'findByName';
+            $roleIdentifier = $this->getRoleIdentifier();
+
+            $method = '';
+
+            if (is_int($role) || PermissionRegistrar::isUid($role)) {
+                $method = 'findById';
+            } else if ($roleIdentifier !== 'name') {
+                $method = 'findByIdentifier';
+            } else {
+                $method = 'findByName';
+            }
 
             return $this->getRoleClass()::{$method}($role, $guard ?: $this->getDefaultGuardName());
         }, Arr::wrap($roles));
@@ -238,7 +248,7 @@ trait HasRoles
      */
     public function hasRole($roles, ?string $guard = null): bool
     {
-        $roleKey = $this->getRoleKey();
+        $roleKey = $this->getRoleIdentifier();
 
         $this->loadMissing('roles');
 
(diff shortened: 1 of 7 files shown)


parallels999:
> Has backward compatibility.
>Introduces findByKey() in the Role

Not, it breaks it,  [spatie/laravel-permission/Contracts/Role.php](https://github.com/spatie/laravel-permission/blob/main/src/Contracts/Role.php)
It also breaks other things, but why go into details, hehe
`getRoleKey` seems like a key, really ambigous

philiplamb:
Hi @parallels999 , thanks for your feedback, i fixed the naming and some mistakes. I'd appreciate if you'd go into more details, or i need to do something else. This is my first PR. Thank you.

src/Traits/HasRoles.php
```diff
@@ -91,7 +91,17 @@ public function scopeRole(Builder $query, $roles, $guard = null, $without = fals
...
 
-            $method = is_int($role) || PermissionRegistrar::isUid($role) ? 'findById' : 'findByName';
+            $roleIdentifier = $this->getRoleIdentifier();
+
+            $method = '';
+
+            if (is_int($role) || PermissionRegistrar::isUid($role)) {
+                $method = 'findById';
+            } else if ($roleIdentifier !== 'name') {
+                $method = 'findByIdentifier';
```
parallels999, line 101: > Not, it breaks it, [spatie/laravel-permission/Contracts/Role.php](https://github.com/spatie/laravel-permission/blob/main/src/Contracts/Role.php)

`findByIdentifier` was added in the model and is used internally, 
It means that those who used the contract (interface) without extending the Role model will not have that method and would get an Exception(method not found), you would have to add it to the contract, but either way it would be a breaking change.

I don't see much use for this change just to change the field name, 
but to avoid the breaking change you should only modify `findByName` and not add another method.

drbyte:
> I needed to use a slug (explicitly not use "name" column), and to check against it

In cases like this I typically use the slug to lookup the model ID, and then explicitly pass the model to all the corresponding functions. No performance hit because lookups of role-related details are user-initiated, not pageload-initiated.

PR #2848, opened 2025-05-22
fix roles

m-jahanzeb6:


src/Traits/HasRoles.php
```diff
@@ -235,9 +235,14 @@ public function syncRoles(...$roles)
      * Determine if the model has (one of) the given role(s).
      *
      * @param  string|int|array|Role|Collection|\BackedEnum  $roles
+     * @throws \TypeError if called with multiple arguments instead of an array
      */
     public function hasRole($roles, ?string $guard = null): bool
     {
+        // Detect incorrect usage: more than one argument passed (e.g. hasRole('admin', 'super'))
+        if (func_num_args() > 2) {
+            throw new \TypeError('hasRole() expects a single argument (string, array, Role, Collection, or BackedEnum). For multiple roles, pass an array.');
```
parallels999, line 244: It wouldn't work very well either, because it's not just one argument, it's two, 
and if you send just two, and they're both role names, it would fail silently.

PR #2850, opened 2025-05-23
Remove `collectRoles` that is not being assigned

pkuiper:
`syncRoles` does not need to call `collectRoles`.

The call is not being assigned to any variable, same as pull request https://github.com/spatie/laravel-permission/pull/2840 does but for roles.

It does not seem to affect any tests

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -223,7 +223,6 @@ trait HasRoles
     public function syncRoles(...$roles)
     {
         if ($this->getModel()->exists) {
-            $this->collectRoles($roles);
             $this->roles()->detach();
             $this->setRelation('roles', collect());
         }


erikn69:
@drbyte hi, I think #2840 should be reverted, and look for another type of optimization
if I remember correctly, this line was there to avoid detaching in an error state.

I explain the problem better
If you do `syncPermissions('permission doesnt exist')`, it will do a detach, then a `collectPermissions('permission doesnt exist')`, this step will generate `PermissionDoesNotExist` exception, so the detach is already permanent, the old permissions were lost, and the new ones were not added.

It was a quick solution that didn't add more queries to the process. Tests should have been added to prevent this problem from happening again. I'll add the tests.

drbyte:
Agreed.

PR #2852, opened 2025-06-03
Add translations support for exception messages

nAa6666:


drbyte:
Thanks.
While I still support Alex's point that these exceptions should only occur developer-side because exceptions should be caught before they reach the user, I support this `__()` approach, which leaves all translations up to the developer, and doesn't invite a library of language files into this package.

PR #2857, opened 2025-07-01
Update permission.php

tassawar-ascend:


config/permission.php
```diff
@@ -2,7 +2,7 @@
 
 return [
 
-    'models' => [
+    'model' => [
```
drbyte, line 5: No, this change would break the entire package.

https://github.com/spatie/laravel-permission/blob/a1851df15da5f9bcb7b47691ba03cd8f320fc4a0/src/PermissionRegistrar.php#L57-L58

PR #2869, opened 2025-08-17
Dispatch RoleDetached on syncRoles when events are enabled

josedaian:
This PR enhances the `syncRoles()` method to dispatch `RoleDetached` events when roles are removed, but **only if events are enabled** via configuration.  
When events are disabled, the method still uses the optimized `detach()` path to avoid unnecessary queries and maintain performance.

### Changes

- Updated `syncRoles()` to:
  - Use `removeRole($currentRoles)` when events are enabled, triggering `RoleDetached`.  
  - Keep the original `detach()` implementation when events are disabled.  
- Added tests to confirm that:
  - `RoleDetached` is dispatched on `syncRoles()` when events are enabled.  
  - Query count remains minimal when events are disabled.  

### Why

- Improves consistency: `syncRoles()` now behaves like `removeRole()` in terms of event dispatching.  
- Allows consumers to listen for role detachment events in both `removeRole()` and `syncRoles()` scenarios.  
- Preserves backward compatibility and performance when events are disabled.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -225,8 +225,15 @@ trait HasRoles
     {
         if ($this->getModel()->exists) {
             $this->collectRoles($roles);
-            $this->roles()->detach();
-            $this->setRelation('roles', collect());
+            if (config('permission.events_enabled')) {
+                $currentRoles = $this->roles()->get();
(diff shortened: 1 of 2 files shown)


drbyte:
Thanks!

PR #2890, opened 2025-11-08
Add BackedEnum Support to RoleOrPermissionMiddleware

imhayatunnabi:
I was working on a multi-tenant inventory which has multiple roles and permissions where we decided to modernize our codebase by migrating from string-based permissions to PHP 8.1+ BackedEnums for better type safety and IDE autocompletion. We created a comprehensive enum structure:

```php
enum Permission: string
{
    case VIEW_DASHBOARD = 'view dashboard';
    case EDIT_ARTICLES = 'edit articles';
    case DELETE_ARTICLES = 'delete articles';
    case MANAGE_USERS = 'manage users';
    case PUBLISH_CONTENT = 'publish content';
}

enum Role: string
{
    case ADMIN = 'admin';
    case EDITOR = 'editor';
    case WRITER = 'writer';
}
```

Everything was going smoothly - we updated our routes to use enums with `PermissionMiddleware` and `RoleMiddleware`, and they worked perfectly:

```php
// This worked great
Route::middleware(['permission:'.Permission::EDIT_ARTICLES->value])
    ->group(function () {
        // ...
    });

// This also worked
Route::middleware([PermissionMiddleware::using(Permission::EDIT_ARTICLES)])
    ->group(function () {
        // ...
    });
```

**Then the problem hit**: We had several routes that needed to check for **either** a role **or** a permission (e.g., "allow access if user is an admin OR has edit permission"). We tried to use `RoleOrPermissionMiddleware` with our enums:

```php
// This failed silently - the middleware didn't recognize the enum
Route::middleware([
    RoleOrPermissionMiddleware::using(Permission::EDIT_ARTICLES)
])->group(function () {
    // Users were getting 403 errors even when they had the permission!
});
```

Here I have added complete BackedEnum which is a PHP built-in interface in PHP 8.1+ versions to support to `RoleOrPermissionMiddleware`, making it consistent with the other middleware classes in the package.

### DIFF
--- a/src/Middleware/RoleOrPermissionMiddleware.php
+++ b/src/Middleware/RoleOrPermissionMiddleware.php
@@ -28,9 +28,7 @@ class RoleOrPermissionMiddleware
             throw UnauthorizedException::missingTraitHasRoles($user);
         }
 
-        $rolesOrPermissions = is_array($roleOrPermission)
-            ? $roleOrPermission
-            : explode('|', $roleOrPermission);
+        $rolesOrPermissions = explode('|', self::parseRoleOrPermissionToString($roleOrPermission));
 
         if (! $user->canAny($rolesOrPermissions) && ! $user->hasAnyRole($rolesOrPermissions)) {
             throw UnauthorizedException::forRolesOrPermissions($rolesOrPermissions);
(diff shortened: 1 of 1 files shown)


drbyte:
Thanks. Yes, this middleware has been out of sync with the others.

Issue #2888, opened 2025-11-01
team_foreign_key column inconsistently nullable


coolamit:
### Description

The `team_foreign_key` column is nullable in `roles` table but not in `model_has_permissions` & `model_has_roles`.

So if an app uses some global roles (available to all teams) then it would set the `team_foreign_key` as NULL for those and assign Team ID to the roles specific to a team. But when a role is added where `team_foreign_key` is set to NULL and then assigned to a user, an error is thrown because `team_foreign_key` columns in `model_has_permissions` & `model_has_roles` are not nullable.

I can raise a PR to fix this in the migration stub if that would be acceptable.

### Steps To Reproduce

1. Enable teams and set the `team_foreign_key` in the `permission.php` config.
2. Create a role but do not set the `team_foreign_key` to it.
3. Assign this role to a user.

### Version of spatie/laravel-permission package:

6.17.0

### Version of laravel/framework package:

11.44.2

### PHP version:

8.3

### Database engine and version:

PostgreSQL

### OS: Windows/Mac/Linux version:

Ubuntu

drbyte:
> This can be fixed easily by adding `nullable()` in Lines 62 & 86 of the migration stub. Primary index creation on Lines 65 & 89 will have to be changed to unique index as well since nullable column cannot be part of primary key.

That's already done in the teams migration file:
https://github.com/spatie/laravel-permission/blob/ea7d29cf733705eba94b002462e42c6ff5585f4c/database/migrations/add_teams_fields.php.stub#L31-L77

coolamit:
> > This can be fixed easily by adding `nullable()` in Lines 62 & 86 of the migration stub. Primary index creation on Lines 65 & 89 will have to be changed to unique index as well since nullable column cannot be part of primary key.
> 
> That's already done in the teams migration file:
> 
> [laravel-permission/database/migrations/add_teams_fields.php.stub](https://github.com/spatie/laravel-permission/blob/ea7d29cf733705eba94b002462e42c6ff5585f4c/database/migrations/add_teams_fields.php.stub#L31-L77)
> 
> Lines 31 to 77 in [ea7d29c](/spatie/laravel-permission/commit/ea7d29cf733705eba94b002462e42c6ff5585f4c)

Its there only for `roles` table but not in `model_has_permissions` & `model_has_roles`. So if a role has been added with team ID set to `null` & it is then assigned to a user, app encounters an error because `model_has_permissions` & `model_has_roles` tables don't accept `null` for team ID.

PR #2894, opened 2025-11-10
Fix: Make `team_foreign_key` nullable in pivot tables (#2888)

imhayatunnabi:
Github Issue: https://github.com/spatie/laravel-permission/issues/2888
The `team_foreign_key` column was nullable in the `roles` table but not in `model_has_permissions` & `model_has_roles` pivot tables. This caused errors when assigning global roles (with `team_foreign_key = null`) to users, as the pivot tables rejected NULL values.

```
Verified that migration files already correctly set `team_foreign_key` as nullable in both pivot tables:
- `add_teams_fields.php.stub` (lines 43 & 64)
- `create_permission_tables.php.stub` (lines 65 & 92)
```
Added test coverage to verify the fix and prevent regressions:

- `it_can_assign_global_role_with_null_team_foreign_key_to_user()` - Core scenario from issue
- `it_can_assign_global_role_to_multiple_users_across_different_teams()` - Unique constraint with NULL
- `it_can_query_global_roles_correctly_across_different_teams()` - Query logic with NULL
- `it_allows_direct_database_insertion_with_null_team_foreign_key()` - Direct schema test
- `it_handles_mixed_team_and_global_roles_correctly()` - Edge cases
- `it_allows_direct_database_insertion_with_null_team_foreign_key_for_permissions()` - Permissions table

### DIFF
--- a/database/migrations/add_teams_fields.php.stub
+++ b/database/migrations/add_teams_fields.php.stub
@@ -51,7 +51,7 @@ return new class extends Migration
                 $table->primary([$pivotPermission, $columnNames['model_morph_key'], 'model_type'],
                     'model_has_permissions_permission_model_type_primary');
                 $table->unique([$columnNames['team_foreign_key'], $pivotPermission, $columnNames['model_morph_key'], 'model_type'],
-                    'model_has_permissions_team_foreign_key_permission_model_type_unique');
+                    'model_has_permissions_team_fk_permission_model_type_unique');
                 if (DB::getDriverName() !== 'sqlite') {
                     $table->foreign($pivotPermission)
                         ->references('id')->on($tableNames['permissions'])->onDelete('cascade');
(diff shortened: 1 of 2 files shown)


runofthemill (commented):
thanks @imhayatunnabi for this fix! I ran into an issue, noted a possible fix below

parallels999:
`teamsKey` is a pivot here, It's not designed to handle `null` and a `team_id` at the same time;

the tests you added aren't testing anything, `it_allows_direct_database_insertion_with_null_team_foreign_key`/`it_allows_direct_database_insertion_with_null_team_foreign_key_for_permissions` only prove that you can register with null, but there's no test of how those registers are used, and according to the sources, it's not possible.

https://github.com/spatie/laravel-permission/blob/c9a58b20c656ea6e22b25e6e1b4d3bab0e0ca044/src/Traits/HasRoles.php#L64-L68
https://github.com/spatie/laravel-permission/blob/c9a58b20c656ea6e22b25e6e1b4d3bab0e0ca044/src/Traits/HasPermissions.php#L94-L97

So my question is, why do you need it to be nullable?
If you have overrides and customizations to handle those null values, you can just modify your local migration file.

PR #2896, opened 2025-11-12
Add case-insensitive duplicate checks for permissions and roles

imhayatunnabi:
While working with the Laravel Permission package, I could create both "Admin" and "admin" as separate roles, which caused confusion and permission management issues.

So, Added case-insensitive duplicate validation to both `Permission::create()` and `Role::create()` methods.

Now it will work like this below:
```php
// Should throw RoleAlreadyExists
Role::create(['name' => 'Admin']);
Role::create(['name' => 'admin']); // Exception thrown

// Should throw PermissionAlreadyExists  
Permission::create(['name' => 'Edit Articles']);
Permission::create(['name' => 'edit articles']); // Exception thrown

// Should still work (different guards)
Role::create(['name' => 'admin', 'guard_name' => 'web']);
Role::create(['name' => 'admin', 'guard_name' => 'api']); // OK
```

### DIFF
--- a/src/Models/Permission.php
+++ b/src/Models/Permission.php
@@ -49,6 +49,19 @@ class Permission extends Model implements PermissionContract
             throw PermissionAlreadyExists::create($attributes['name'], $attributes['guard_name']);
         }
 
+        // Check for case-insensitive duplicate - simple approach
+        if (config('permission.check_case_insensitive_duplicates', true)) {
+            $lowerName = strtolower($attributes['name']);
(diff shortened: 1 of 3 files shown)


imhayatunnabi:
Screenshot 2025-11-12 at 11 34 43 PM

The PHPStan CI check is failing, but this is **not related to the changes in this PR**. 

The error is a fatal error in the `orchestra/canvas` dependency:

```
Orchestra\Canvas\Console\ControllerMakeCommand::possibleModels() has #[\Override] attribute, but no matching parent method exists
```

PR #2922, opened 2026-02-21
Add syncModels method to Role model with HasModels trait

Briareos17:


Briareos17:
> Please revert the change in composer.json. Too rigid.

Done

freekmurze:
Thanks!

PR #2926, opened 2026-02-22
Solve recursive permission inheritance using polymorphic helpers

muhammed-rashid:
This PR addresses the infinite recursion issue in the 
Permission
 model and simplifies the trait architecture by replacing hardcoded instanceof checks with a cleaner, polymorphic approach.


freekmurze:
I'm going to pass on this one for now.

PR #2928, opened 2026-02-23
Teams accessors in HasRolesTrait

AnjeyNov:


freekmurze:
Hi @AnjeyNov, thanks for the PR! I pushed three commits on top of yours to address a few things. Happy to revert any of these if you'd rather take a different direction.

**1. `Support\Config` class with typed accessors** (d91b358)

We tend to centralize config lookups behind a small `Config` class (the same pattern we use in `laravel-url-ai-transformer`). I added `Spatie\Permission\Support\Config` with `teamModel()`, `modelHasRolesTable()`, `morphKey()`, `teamForeignKey()`, `teamsEnabled()`, and `ensureTeamsEnabled()`. `teamModel()` is the single point that throws `TeamsNotEnabled` and `TeamModelNotConfigured`, so callers don't have to repeat both checks.

**2. Use Config in HasRoles, drop the no-op container bind** (3f06b11)

`teams()`, `scopeTeam()`, and `scopeWithoutTeam()` now route through `Config::*` instead of repeating `config('permission.models.team')`, `config('permission.table_names.model_has_roles')`, etc. across both methods.

In the same commit I dropped this line from `setTeamClass`:

```php
if (\$teamClass) {
    app()->bind(config('permission.models.team'), \$teamClass);
}
```

For role/permission this binds the contract (`Role::class`, `Permission::class`) to the concrete class. Since there's no `Team` contract, the line was binding `\$teamClass => \$teamClass` (e.g. `Team::class => Team::class`), which has no effect. Also initialized `protected ?string \$teamClass = null;` on the property so it isn't uninitialized before the constructor runs.

**3. Real Team test model + positive coverage for `teams()`** (e686d22)

The previous test setup aliased `User::class` as the team model so `instanceof` would pass. I added a dedicated `tests/TestSupport/TestModels/Team.php`, configure it via `setTeamClass(Team::class)` in `beforeEach`, and create/drop a `teams` table per test. That removed the apologetic comment and let me add two new tests:

- `it('returns the teams a user belongs to via the teams relation')`
- `it('can scope users by team using a model instance')`

I also moved the duplicated `User::all()->each(fn (\$u) => \$u->delete())` into the shared `beforeEach` (`User::query()->delete()`).

Locally `vendor/bin/pest --filter TeamScopeTest` is green (12 passed, 28 assertions). The 61 unrelated failures elsewhere in the suite (Passport `oauth_clients.redirect_uris`) reproduce on the unmodified PR HEAD too, so they're pre-existing.

freekmurze:
Thanks!

ju5t:
@freekmurze @AnjeyNov FYI:

This causes issues on Jetstream installs as the `User` model already has `teams()` implemented there. This caused a bit of a headache as it just crashed our tests and somehow the crash pointed to unrelated code.

The solution was:

```php
use HasRoles, HasTeams {
    HasTeams::teams insteadof HasRoles;
}
```

@freekmurze I think it would help if this was documented. I assume a lot of people use Jetstream and your permission package so chances are they'll have this issue too.

parallels999:
- Related #2950

func0der:
I might be a little late to the party, but was this not considered to be a breaking change? 

On paper it is just adding a new method, should be fine in most libraries as a "feature release.
 But since Laravel works on models like they are arrays on outputting them via http responses, this could break existing code for people having teams enabled. `teams` would be serialized to responses and since the config key is not configured, because nobody expects having to set new, required config in a feature release, everybody's code would just blow up.

Is this just an oversight or I miss something on my side?

PR #2944, opened 2026-04-12
Fix create-role handling when teams are disabled

dhrupo:
This PR fixes `permission:create-role` when `--team-id` is passed while teams are disabled.

## Problem
The command warned that `--team-id` had no effect, but returned early before creating the role. It also left the in-process team context changed because the temporary team id was set before the early return.

## Changes
- keep warning when `--team-id` is used while teams are disabled
- continue creating the role instead of returning early
- restore the previous team context with a `try/finally`
- add a regression test covering role creation and team id restoration

## Tests
- `/opt/homebrew/opt/php@8.3/bin/php vendor/bin/pest tests/Commands/CommandTest.php`

This is a focused bug fix with a matching test, following the contribution guide requirement to include tests.

### DIFF
--- a/src/Commands/CreateRoleCommand.php
+++ b/src/Commands/CreateRoleCommand.php
@@ -23,16 +23,20 @@ class CreateRoleCommand extends Command
         $roleClass = app(RoleContract::class);
 
         $teamIdAux = getPermissionsTeamId();
-        setPermissionsTeamId($this->option('team-id') ?: null);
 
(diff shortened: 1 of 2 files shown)


drbyte:
Given that `--team-id` must be intentionally passed, I don't think it should continue to create the role if the teams feature is disabled. The user should be informed of the problem so they can correct the configuration first. Otherwise the role that was created will be incorrectly linked.

That's why the early-return was implemented.

freekmurze:
We're going to pass on this one for now.

PR #2951, opened 2026-05-21
Add configured permission generation command and added elements to pe…

dkiptugen:
…rmission migration


dkiptugen (commented):
Created a command to be used in generating roles and permissions from the config file

dkiptugen (commented):
Marked the fields as nullable

database/migrations/create_permission_tables.php.stub
```diff
@@ -25,6 +25,8 @@ return new class extends Migration
          */
         Schema::create($tableNames['permissions'], static function (Blueprint $table) {
             $table->id(); // permission id
+            $table->string('display_name')->nullable();
+            $table->string('permission_group')->nullable();
```
parallels999, line 29: Why? It looks like a customization you use.

drbyte:
Thanks for this suggestion; however, we've intentionally avoided adding those additional fields to the tables, in favor of letting userland code add what they need, instead of us maintaining them along with all the ripple-effect requirements that people will want to use them with.

Also, on the pregenerated-permissions approach, I think a bespoke migration/seeder could be written for those and included in your project, rather than using a generator command which still needs to reference a code change you'd have to add to your project anyway.

So thanks for the contribution; however, we're going to pass on this for now.

PR #2953, opened 2026-05-30
Extend BackedEnum support on Role/Permission 

YasserB94:
Iterating on  #2826, since the minimum php version is now 8.3 I don't think this introduces **breaking** changes anymore

This PR adds `BackedEnum` support to `create()`, `findByName()`, and `findOrCreate()` on the `Role` and `Permission` models.

## Changes
- `Permission::create()`, `findByName()`, `findOrCreate()` now accept `BackedEnum|string`
- `Role::create()`, `findByName()`, `findOrCreate()` now accept `BackedEnum|string`
- Contracts updated accordinglu
- Tests updated and/or added covering both string and enum usage
- Extended the existing documentation:
  - enum documentation now reflects these changes
  - extending documentation repeats the tip to use Laravel's `enum_value` when extending the model _(I was not sure about the best place to put this so I added a Tips section)_

### _Breaking?_
 - Both the `Permission` and `Role` contract have been updated to accept a `BackedEnum` as well. This might cause a minor change for custom implementations. _(Though substitution applies so I don't think this requires a major version update?)_

### DIFF
--- a/docs/advanced-usage/extending.md
+++ b/docs/advanced-usage/extending.md
@@ -95,3 +95,13 @@ In the rare case that you have need to REPLACE the existing `Role` or `Permissio
 - Your `Permission` model needs to implement the `Spatie\Permission\Contracts\Permission` contract
 - You need to update `config/permission.php` to specify your namespaced model
 
+### Tips
+
+If you override `findByName`, `findOrCreate`, or `create` on your model, you can use Laravel's `enum_value()` helper to resolve `BackedEnum` values:
+ ```php
+ public static function findByName(BackedEnum|string $name, ?string $guardName = null): self
+ {
+     $name = enum_value($name);
+     // ...
+ }
+ ```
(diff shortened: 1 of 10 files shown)


drbyte (approved):
Still have to decide whether this warrants a major version bump.
Other packages which depend on this one are more likely to have implemented the contracts than individual apps are, but that landscape affects a lot of users.

drbyte:
Given that the contract change is an _expansion_, I'm less worried about it being "painful": it's a very simple update.
But anyone updating without reading release notes or running a test suite, will run into contract warnings ... if they've implemented those contracts.

A github search shows that the impact vector is around 130 repos, and the majority of those are advanced setups which probably have solid dev teams who wouldn't just blindly upgrade.

Also, the several more popular packages which depend on this one all extend the Model, not the Contract, so they're not impacted by the change ... which is fantastic.

drbyte:
Thanks

YasserB94:
@drbyte  thanks for the thoughtful review. 

I wasn't confident in the impact of changing the contract _(hence the breaking? block)_, though the PR felt incomplete without it. I am glad someone with more context was able to make this decision. 

Thanks once again for merging this!

PR #2960, opened 2026-06-26
Fix quote stripping in convertPipeToArray for mismatched quotes

albertoarena:
## The bug

`HasRoles::convertPipeToArray()` is meant to strip a matching pair of surrounding quotes from a pipe-delimited string (e.g. `@hasanyrole("'writer|admin'")`) before splitting it on `|`. The quotes should only be stripped when the string is wrapped in the **same** quote character on both ends.

The check for "same character on both ends" was broken:

```php
$quoteCharacter = substr($pipeString, 0, 1);
$endCharacter   = substr($quoteCharacter, -1, 1); // reads $quoteCharacter, not $pipeString
```

`$quoteCharacter` is already a single character, so `substr($quoteCharacter, -1, 1)` always returns that same character. As a result `$endCharacter` was always equal to `$quoteCharacter`, and the guard:

```php
if ($quoteCharacter !== $endCharacter) {
    return explode('|', $pipeString);
}
```

was dead code that could never run. A string opening with a quote but **not** closing with a matching one would still have its leading quote stripped.

## Example

```php
// input: 'writer|admin   (leading single quote, no closing quote)

// before: ['writer', 'admin']    — leading quote wrongly stripped
// after:  ["'writer", 'admin']   — leading quote preserved
```

## The fix

Read the closing character from `$pipeString` instead of from `$quoteCharacter`, so surrounding quotes are stripped only when they genuinely match on both ends:

```php
$endCharacter = substr($pipeString, -1, 1);
```

Properly quoted (`'a|b'`, `"a|b"`) and unquoted (`a|b`) inputs behave exactly as before; only the previously-mishandled mismatched-quote case changes.

## Tests

Added a regression test (`can convert pipe strings to arrays, stripping only matched surrounding quotes`) covering unquoted, matched-quote, and mismatched-quote inputs. It fails on the previous code and passes with the fix. Full suite remains green (`650 passed`) and Pint is clean.

### DIFF
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -482,7 +482,7 @@ trait HasRoles
         }
 
         $quoteCharacter = substr($pipeString, 0, 1);
-        $endCharacter = substr($quoteCharacter, -1, 1);
+        $endCharacter = substr($pipeString, -1, 1);
 
         if ($quoteCharacter !== $endCharacter) {
             return explode('|', $pipeString);
--- a/tests/Traits/HasRolesTest.php
+++ b/tests/Traits/HasRolesTest.php
@@ -110,6 +110,26 @@ it('can scope a role using enums', function () {
     expect($scopedUsers3->count())->toEqual(3);
 });
 
+it('can convert pipe strings to arrays, stripping only matched surrounding quotes', function () {
+    $convert = function (string $pipeString) {
+        $method = new ReflectionMethod($this->testUser, 'convertPipeToArray');
+
+        return $method->invoke($this->testUser, $pipeString);
+    };
+
+    // Unquoted input splits on the pipe.
+    expect($convert('writer|admin'))->toBe(['writer', 'admin']);
+
+    // Matched surrounding quotes are stripped before splitting.
+    expect($convert("'writer|admin'"))->toBe(['writer', 'admin']);
+    expect($convert('"writer|admin"'))->toBe(['writer', 'admin']);
+
+    // Mismatched quotes (leading quote only) must NOT be stripped: the leading
+    // quote is part of the first value. The previous self-comparison bug made
+    // this guard dead and incorrectly stripped the quote.
+    expect($convert("'writer|admin"))->toBe(["'writer", 'admin']);
+});
+
 it('can assign and remove a role', function () {
     expect($this->testUser->hasRole('testRole'))->toBeFalse();
 


drbyte:
Thanks for this. 
Can you give a real-world use-case for when `'writer` would be a desirable role/permission name? ie: embedded quotes/quotations in the name string

albertoarena:
> Thanks for this. Can you give a real-world use-case for when `'writer` would be a desirable role/permission name? ie: embedded quotes/quotations in the name string

Good question, and honestly no, I don't think anyone deliberately wants a role named `'writer`. I wouldn't defend it on that basis.

The motivation is narrower: the guard just below those lines was clearly meant to only strip surrounding quotes when both ends match. Because `$endCharacter` reads from `$quoteCharacter` (already a single char) instead of `$pipeString`, that branch has been dead since the original 2017 implementation, so a leading quote gets stripped even with no matching closing one. The one-character fix just makes the code do what the guard already says.

The practical effect is that `hasAnyRole("'writer|admin")` currently checks for roles named `writer` and `admin`, quietly dropping the leading quote the caller passed. Since names are arbitrary strings, ones beginning with a quote char can come from imported or auto-generated sources rather than being typed on purpose. The matched-pair (`'a|b'`, `"a|b")` and unquoted cases are unchanged.

Happy to reframe the title as "honor the matched-pair guard" if that reads more clearly. And of course fine to close it if you feel it's too marginal to be worth the change.

drbyte:
Ya, I don't disagree.

Next challenge is whether this might introduce a Breaking Change where what "used to work, despite a typo" suddenly stops working. Sure it exposes a bug in userland code, but whether to force that mid-release or not is my question.

Likelihood of actual breakage is small, because likelihood of such typos is also probably very small. But still...

drbyte:
Thinking process:

The PR only changes malformed pipe strings that start with `'` or `"` and do not end with the same quote. Normal documented usage is unchanged:

```php
hasAnyRole('writer|admin')       // same
hasAnyRole("'writer|admin'")     // same
hasAnyRole('"writer|admin"')     // same
hasAnyRole("'writer|admin")      // changes
hasAnyRole("'writer|admin\"")    // changes
```

The only plausible break is as described: a typo like `hasAnyRole("'writer|admin")` which currently, accidentally, checks `writer` and `admin`. After the fix it checks `'writer` and `admin` instead, so access could change from allowed to denied. 

While technically behavior-changing, but it depends on invalid/malformed input and contradicts the original intent of the code ... which classifies more as a bug.

So I'm gonna call it a bugfix.

PR #2961, opened 2026-06-27
Fix team delete/sync when using custom pivot class

### LINKED ISSUE #2867: removeRole ignores team when using custom Pivot class (with teams enabled)
### Description

If you setup your `roles()` relationship to use a custom Pivot model and you have teams enabled, then the `removeRole()` method fails to include the team scope because `InteractsWithPivotTable::detach()` doesn't run a single query to detach the pivot records, rather it uses the function `InteractsWithPivotTable::detachUsingCustomClass()` which runs `InteractsWithPivotTable::getCurrentlyAttachedPivotsForIds()` and then iterates over the result and deletes each model.

For whatever reason the scoping applied when teams is enabled doesn't get applied when `InteractsWithPivotTable::detachUsingCustomClass()` is used. 

### Steps To Reproduce

works with teams:
```
    use HasRoles {
        roles as traitRoles;
    }

    public function roles(): BelongsToMany
    {
        // We're adding timestamps so we know when a role was added and/or updated.
        // And we always want the vessel_id even if we disable laravel-permissions teams support
        return $this->traitRoles()
            ->withPivot('vessel_id')
            ->withTimestamps();
            //->using(UserRole::class);
    }
```

does not work with teams
```
    use HasRoles {
        roles as traitRoles;
    }

    public function roles(): BelongsToMany
    {
        // We're adding timestamps so we know when a role was added and/or updated.
        // And we always want the vessel_id even if we disable laravel-permissions teams support
        return $this->traitRoles()
            ->withPivot('vessel_id')
            ->withTimestamps();
            ->using(UserRole::class);
    }
```

Test case:
```
<?php

use App\Models\Role;
use App\Models\User;
use App\Models\Vessel;
use Illuminate\Foundation\Testing\DatabaseTransactions;
use Tests\TestCase;

uses(TestCase::class);
uses(DatabaseTransactions::class);

test('Can sync or remove roles without detach on different teams', function () {

    app(\Spatie\Permission\PermissionRegistrar::class)->teams = true;

    $vessel1 = Vessel::factory()->create();
    $vessel2 = Vessel::factory()->create();

    Role::create(['name' => 'testRole', 'guard_name' => 'api']);
    Role::create(['name' => 'testRole2', 'guard_name' => 'api']);
    Role::create(['name' => 'testRole3', 'guard_name' => 'api']);

    $testUser = User::factory()->create();

    // For Vessel 1 attached testRole and testRole2
    setPermissionsTeamId($vessel1);
    $testUser->unsetRelation('roles');
    $testUser->syncRoles('testRole', 'testRole2');
    Log::debug('Vessel 1 Attached Roles', $testUser->roles->pluck('name')->toArray());

    // For Vessel 2 attached testRole and testRole3
    setPermissionsTeamId($vessel2);
    $testUser->unsetRelation('roles');
    $testUser->syncRoles('testRole', 'testRole3');
    Log::debug('Vessel 2 Attached Roles', $testUser->roles->pluck('name')->toArray());

    // Test Vessel 1 only has testRole and testRole2, and that testRole3 didn't get added or testRole2 removed in the syncRoles on Vessel 2
    setPermissionsTeamId($vessel1);
    $testUser->unsetRelation('roles');
    $this->assertEquals(
        collect(['testRole', 'testRole2']),
        $testUser->getRoleNames()->sort()->values()
    );

    // For Vessel 1 remove testRole and confirm the user only has testRole2 attached
    $testUser->removeRole('testRole');
    Log::debug('Vessel 1 Attached Roles', $testUser->roles->pluck('name')->toArray());
    $this->assertEquals(
        collect(['testRole2']),
        $testUser->getRoleNames()->sort()->values()
    );

    // For Vessel 2, confirm it still has testRole and testRole3, i.e. testRole wasn't incorrectly moved when it was removed from Vessel 1
    setPermissionsTeamId($vessel2);
    $testUser->unsetRelation('roles');

    Log::debug('Vessel 2 Attached Roles', $testUser->roles->pluck('name')->toArray());
    $this->assertEquals(
        collect(['testRole', 'testRole3']),
        $testUser->getRoleNames()->sort()->values()
    );
});
```

### Example Application

_No response_

6.21.0

12.21.0

### PHP version:

8.2

### Database engine and version:

_No response_

### OS: Windows/Mac/Linux version:

_No response_

drbyte:
## Summary

Fixes `removeRole()`/`syncRoles()` and `revokePermissionTo()`/`syncPermissions()` when the teams feature is enabled and a model overrides `roles()` or `permissions()` to use a custom pivot class via `->using(CustomPivot::class)`.

Previously, removing a role (or permission) from one team could also remove that same assignment from other teams for the same model.

Fixes #2867

## Bug

When teams are enabled, the `roles()`/`permissions()` relations correctly scope assignments by the current team id.

However, Laravel handles `detach()` differently when a custom pivot class is configured:

1. It selects the currently attached pivot rows using the scoped relation query.
2. It hydrates those rows as custom pivot models.
3. It calls `delete()` on each custom pivot model.

The selected row is team-scoped, but the custom pivot model's delete query does not include the team foreign key — it deletes by pivot identity columns (model id, role/permission id, morph type). As a result, if the same model has the same role or permission on multiple teams, removing it from one team can delete it from other teams too.

## Fix

Adds internal `detachRoles()` and `detachPermissions()` helpers.

For normal cases, both still delegate to Laravel's regular `detach()` behavior:

- teams disabled
- the "wrong-direction" model (`Permission` for roles, `Role` for permissions)
- default pivot class

For the affected case — `teams=true` plus a custom pivot class — deletion goes through the relation's scoped pivot query instead, which preserves the current team id in the delete predicate.

## Userland mitigation

This changes behavior for users who relied on custom pivot model `deleting`/`deleted` lifecycle hooks firing during `removeRole()`, `syncRoles()`, `revokePermissionTo()`, or `syncPermissions()` with teams enabled.

Those hooks were already unreliable in this scenario: Laravel would fire the hook for the selected team-scoped pivot row, but the resulting delete could remove additional matching rows from other teams with no corresponding hook events.

Users who need to react to package-level role/permission removal should listen for `Spatie\Permission\Events\RoleDetachedEvent` / `PermissionDetachedEvent` instead. Users who need row-level custom pivot lifecycle behavior for this specific team/custom-pivot case should perform their own explicitly team-scoped pivot operation and dispatch their own application event.

**Permissions only:** `revokePermissionTo()` now normalizes its input through the same path `givePermissionTo()` uses, which includes a guard-match check. It can now throw `GuardDoesNotMatch` in cases it previously wouldn't have (`removeRole()` already had this check, so roles are unaffected). In practice this should be a no-op — a permission being revoked is already attached to the model and therefore already guard-matched — but it's a real code-path change worth knowing about if you have unusual guard setups.

## Tests

Added coverage, for both roles and permissions, for:

- custom pivot with teams enabled: removing one role/permission from one team does not remove the same one from another team
- custom pivot with teams enabled: syncing to a non-empty set on one team does not touch another team's assignments
- custom pivot with teams enabled and events enabled: sync's events-branch (which revokes individually) does not cross team boundaries either
- custom pivot with teams enabled: syncing to an empty set on one team does not touch another team's assignments
- custom pivot with teams disabled: remove/sync continue to work normally (unaffected baseline)

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -5,6 +5,7 @@ namespace Spatie\Permission\Traits;
 use BackedEnum;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\Pivot;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Spatie\Permission\Contracts\Permission;
@@ -385,6 +386,32 @@ trait HasPermissions
             }, []);
     }
 
+    private function detachPermissions(?array $permissions = null): int
+    {
+        $relation = $this->permissions();
+
+        if (! Config::teamsEnabled() || $this instanceof Role || $relation->getPivotClass() === Pivot::class) {
+            return $relation->detach($permissions);
+        }
+
+        // Custom pivot deletes do not include the team key, so keep deletion on the scoped pivot query.
+        $query = $relation->newPivotQuery();
+
+        if (! is_null($permissions)) {
+            if (empty($permissions)) {
+                return 0;
+            }
+
+            $query->whereIn($relation->getQualifiedRelatedPivotKeyName(), $permissions);
+        }
+
+        $results = $query->delete();
+
+        $relation->touchIfTouching();
+
+        return $results;
+    }
+
     /**
      * Grant the given permission(s) to a role.
      *
@@ -458,7 +485,7 @@ trait HasPermissions
                     $this->revokePermissionTo($currentPermissions);
                 }
             } else {
-                $this->permissions()->detach();
+                $this->detachPermissions();
                 $this->setRelation('permissions', collect());
             }
         }
@@ -475,8 +502,9 @@ trait HasPermissions
     public function revokePermissionTo($permission): static
     {
         $storedPermission = $this->getStoredPermission($permission);
+        $permissions = $this->collectPermissions($storedPermission);
 
-        $this->permissions()->detach($storedPermission);
+        $this->detachPermissions($permissions);
 
         if ($this instanceof Role) {
             $this->forgetCachedPermissions();
--- a/src/Traits/HasRoles.php
+++ b/src/Traits/HasRoles.php
@@ -6,6 +6,7 @@ use BackedEnum;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+use Illuminate\Database\Eloquent\Relations\Pivot;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Spatie\Permission\Contracts\Permission;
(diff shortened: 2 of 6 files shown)


SlyDave:
Looks great, I'd ask if @maxvoly can pull this and run the tests against it confirming it's fixed, as I am no longer working in that team. (p.s. Thanks Max!)

drbyte:
@codex review

PR #2963, opened 2026-07-02
Fire PermissionDetachedEvent when syncing permissions (symmetry with syncRoles)

### LINKED ISSUE #2962: syncPermissions() does not fire PermissionDetachedEvent when events_enabled (asymmetric with syncRoles())
### Summary

First off, thank you for this package — it's a cornerstone of so many Laravel apps, ours included. 🙏

While adding audit logging on top of the permission events, I think I ran into a small inconsistency between `syncRoles()` and `syncPermissions()` when `permission.events_enabled` is `true`:

- `syncRoles()` fires `RoleDetachedEvent` for the roles it removes.
- `syncPermissions()` does **not** fire `PermissionDetachedEvent` for the permissions it removes.

So a permission removed via `syncPermissions()` is invisible to event listeners, while a role removed via `syncRoles()` is not. Apologies if this is intentional or already known — I couldn't find an existing issue.

### Root cause (as I understand it)

`syncRoles()` was given an events-aware branch (it calls `removeRole()` on the current roles, which fires the event), but `syncPermissions()` still does a bulk `detach()`, which bypasses the event:

```php
// HasRoles::syncRoles() — fires RoleDetachedEvent when events are enabled
if (Config::eventsEnabled()) {
    $currentRoles = $this->roles()->get();
    if ($currentRoles->isNotEmpty()) {
        $this->removeRole($currentRoles); // -> RoleDetachedEvent
    }
} else {
    $this->roles()->detach();
    $this->setRelation('roles', collect());
}

// HasPermissions::syncPermissions() — no events branch, no PermissionDetachedEvent
$this->permissions()->detach();          // bulk delete, no event
$this->setRelation('permissions', collect());
```

### Reproduction

```php
config(['permission.events_enabled' => true]);

Event::listen(PermissionAttachedEvent::class, fn ($e) => logger('attached', [$e->permissionsOrIds]));
Event::listen(PermissionDetachedEvent::class, fn ($e) => logger('detached', [$e->permissionsOrIds]));

$role->givePermissionTo(['a', 'b']);   // PermissionAttachedEvent [a, b]
$role->syncPermissions(['a']);         // PermissionAttachedEvent [a]  — no "detached" for b
```

Observed: only `PermissionAttachedEvent[a]` fires; `b` is removed from the DB with no `PermissionDetachedEvent`. The analogous `syncRoles()` call does fire `RoleDetachedEvent`.

### Environment

Confirmed on 6.25.0 and reproduced against current `main` (post-8.1.0). Related history: #1876 (original bulk-detach event gap) and #2742 (added these four events and fixed the `syncRoles` half).

### Fix

Happy to open a PR that mirrors the `syncRoles()` events branch in `syncPermissions()`, with a test alongside the existing `fires detach event when syncing roles`. Will link it here.

---

_Disclosure: the investigation and the write-up were done with AI assistance (Claude) under my supervision — I've reviewed and run everything, and I'm happy to iterate on anything that isn't quite right._

chengkangzai:
Thank you all for this package — it does an enormous amount of heavy lifting for the Laravel community, and it's been rock-solid for us. 🙏

### What this does

Fixes the small asymmetry reported in #2962: when `permission.events_enabled` is `true`, `syncRoles()` fires `RoleDetachedEvent` for removed roles, but `syncPermissions()` bulk-detaches with no `PermissionDetachedEvent`, so permission removals via `sync` are invisible to listeners.

This mirrors the existing `syncRoles()` events branch inside `syncPermissions()`: when events are enabled, it revokes the current permissions individually (via `revokePermissionTo()`, which already fires `PermissionDetachedEvent`) before granting the new set. When events are disabled the behaviour is unchanged — it keeps the fast bulk `detach()`.

```php
public function syncPermissions(...$permissions): static
{
    if ($this->getModel()->exists) {
        $this->collectPermissions($permissions);

        if (Config::eventsEnabled()) {
            $currentPermissions = $this->permissions()->get();

            if ($currentPermissions->isNotEmpty()) {
                $this->revokePermissionTo($currentPermissions);
            }
        } else {
            $this->permissions()->detach();
            $this->setRelation('permissions', collect());
        }
    }

    return $this->givePermissionTo($permissions);
}
```

### Notes / trade-offs

- This intentionally follows the exact same shape and trade-off as `syncRoles()`: like that method, it fires a "detached" event for the full current set (not only the net-removed items) when events are enabled. Kept it symmetrical rather than trying to be cleverer than the roles path, but happy to change direction if you'd prefer a net-diff approach for both.
- Only the events-enabled path changes; the default (events off) path is untouched, so there should be no performance impact for the common case.

### Tests

Added `fires detach event when syncing permissions` to `tests/Traits/HasPermissionsTest.php`, mirroring the existing `fires detach event when syncing roles`.

- Full suite: `651 passed`
- PHPStan: `No errors`
- Pint: clean

### Closes

Closes #2962

---

_Disclosure: this change was prepared with AI assistance (Claude) under my supervision. I've read the diff, run the full test suite / PHPStan / Pint locally, and I take responsibility for it — very happy to revise anything, address review feedback, or maintain it over time. Thanks again to the Spatie team for everything you build._

### DIFF
--- a/src/Traits/HasPermissions.php
+++ b/src/Traits/HasPermissions.php
@@ -443,8 +443,17 @@ trait HasPermissions
     {
         if ($this->getModel()->exists) {
             $this->collectPermissions($permissions);
-            $this->permissions()->detach();
-            $this->setRelation('permissions', collect());
+
+            if (Config::eventsEnabled()) {
+                $currentPermissions = $this->permissions()->get();
+
+                if ($currentPermissions->isNotEmpty()) {
+                    $this->revokePermissionTo($currentPermissions);
+                }
+            } else {
+                $this->permissions()->detach();
+                $this->setRelation('permissions', collect());
+            }
         }
 
         return $this->givePermissionTo($permissions);
--- a/tests/Traits/HasPermissionsTest.php
+++ b/tests/Traits/HasPermissionsTest.php
@@ -675,6 +675,35 @@ it('fires an event when a permission is removed', function () {
     });
 });
 
+it('fires detach event when syncing permissions', function () {
+    Event::fake([PermissionDetachedEvent::class, PermissionAttachedEvent::class]);
+    app('config')->set('permission.events_enabled', true);
+
+    $this->testUser->givePermissionTo('edit-articles', 'edit-news');
+
+    $this->testUser->syncPermissions('edit-articles');
+
+    expect($this->testUser->hasPermissionTo('edit-articles'))->toBeTrue();
+    expect($this->testUser->hasPermissionTo('edit-news'))->toBeFalse();
+
+    Event::assertDispatched(PermissionDetachedEvent::class, function ($event) {
+        $names = collect($event->permissionsOrIds)->map(
+            fn ($permission) => is_object($permission)
+                ? $permission->name
+                : app(Permission::class)::findById($permission)->name
+        );
+
+        return $event->model instanceof User
+            && ! $event->model->hasPermissionTo('edit-news')
+            && $names->contains('edit-news');
+    });
+
+    Event::assertDispatched(PermissionAttachedEvent::class, function ($event) {
+        return $event->model instanceof User
+            && $event->model->hasPermissionTo('edit-articles');
+    });
+});
+
 it('can be given a permission on role when lazy loading is restricted', function () {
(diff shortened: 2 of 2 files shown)


drbyte:
Thanks!

PR #2952, opened 2026-05-30
Remove stale phpstan ignore HasPermission instanceof.alwaysFalse

YasserB94:
Was running PHPStan as I tried to work on something and I noticed there  was a stale ignore:

```sh
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Traits/HasPermissions.php
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------
         Ignored error pattern instanceof.alwaysFalse in path /Volumes/DataT7/Developer/Web/laravel-permission/src/Traits/HasPermissions.php was not matched in reported errors.
         🪪  ignore.unmatched (non-ignorable)
 ------ -------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### Changes
- Remove the ignore rule from `phpstan.neon.dist`

PR #2966, opened 2026-07-03
Add wiring-level test coverage for the Octane reset listener

drbyte:
## Add wiring-level test coverage for the Octane reset listener

`PermissionServiceProvider::registerOctaneListener()` wires two listeners for `Laravel\Octane\Contracts\OperationTerminated`, resetting the permissions team ID on every request, and (opt-in via `permission.register_octane_reset_listener`) clearing the in-memory permissions collection, but neither had any test coverage. The registration code bails out early via `if ($this->app->runningInConsole() || ...)`, and since Pest/Testbench always runs under the CLI SAPI, `runningInConsole()` is always `true` in the test suite so this code path was silently never exercised.

This PR adds real coverage without requiring a running Swoole/RoadRunner server:

- Adds `laravel/octane` as a **dev-only** dependency: it's pure PHP with no extension requirement (swoole/roadrunner are only needed to actually *serve* traffic via `octane:start`, not to use Octane's `Contracts`/`Events` classes).
- `tests/Integration/OctaneListenerTest.php` forces `Application::$isRunningInConsole` to `false` via reflection (the real, memoized gate Laravel checks — see `Illuminate\Foundation\Application::runningInConsole()`), invokes the real (protected) `registerOctaneListener()` method, and dispatches an actual `Laravel\Octane\Events\RequestTerminated` event to exercise the genuine listener wiring end-to-end.
- Each test dispatches the event with a **cloned sandbox app distinct from the base app** (`clone app()` + `forgetInstance(PermissionRegistrar::class)`), so the assertions only pass if the listener acts on `$event->sandbox()` specifically, not `$event->app()`. The tests were confirmed to fail when the production code was temporarily changed to use `$event->app()` instead.

**Also fixes a related correctness issue:** `registerOctaneListener()` accessed `$event->sandbox` as a property, which isn't part of the `OperationTerminated` interface (only `sandbox(): Application` is). Adding `laravel/octane` as a dependency let PHPStan finally resolve the class and catch this once the stale `@phpstan-ignore-next-line` suppressions were removed. Fixed by calling `$event->sandbox()` instead — no behavior change, but now type-checked.

**Coverage added:**
- Listener is not registered when `octane.listeners` is disabled.
- `PermissionRegistrar::getPermissionsTeamId()` resets to `null` on the sandbox instance on operation termination, while the base app's instance is untouched.
- The permissions collection is cleared on the sandbox instance only when `permission.register_octane_reset_listener` is enabled, and left untouched (on both instances) when it isn't.

**Maintenance trade-off:** the tests reach into a protected framework property (`Application::$isRunningInConsole`) and a protected provider method via reflection. If Laravel renames that property, or Octane changes `RequestTerminated`'s constructor signature, these tests could break for reasons unrelated to an actual regression here ... but we keep in sync with Laravel releases already.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,6 +31,7 @@
     },
     "require-dev": {
         "larastan/larastan": "^3.9",
+        "laravel/octane": "^2.0",
         "laravel/passport": "^13.0",
         "laravel/pint": "^1.0",
         "orchestra/testbench": "^10.0|^11.0",
--- a/src/PermissionServiceProvider.php
+++ b/src/PermissionServiceProvider.php
@@ -121,19 +121,15 @@ class PermissionServiceProvider extends PackageServiceProvider
         }
 
         $dispatcher = $this->app[Dispatcher::class];
-        // @phpstan-ignore-next-line
         $dispatcher->listen(function (OperationTerminated $event) {
-            // @phpstan-ignore-next-line
-            $event->sandbox->make(PermissionRegistrar::class)->setPermissionsTeamId(null);
+            $event->sandbox()->make(PermissionRegistrar::class)->setPermissionsTeamId(null);
         });
 
         if (! $this->app['config']->get('permission.register_octane_reset_listener')) {
             return;
         }
-        // @phpstan-ignore-next-line
         $dispatcher->listen(function (OperationTerminated $event) {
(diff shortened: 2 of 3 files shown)


drbyte:
Related: #2952

